PR #3, opened 2020-09-23
Change bucket_id function

dokshina:
Change bucket_id_mpcrc32 -> router_bucket_id_strcrc32

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ across the cluster.
 **Notes:**
 
 * A space should have a format.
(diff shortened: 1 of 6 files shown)


Totktonada:
But why?

(I'm here from 2021 via #185.)

olegrok:
Because it was a plan that users could start to use crud on existing installations. They work with bucket_id/bucket_id_strcrc32 functions. 

Usage of mpcrc32 created restriction that only new customers/projects could use crud.
And of course initially nobody think about a case when it could be configurable.

PR #4, opened 2020-09-23
test: add "vinyl" engine tests to "select_test"

olegrok:


dokshina (commented):
Thanks for your patch!

But what about other operations?

olegrok:
> But what about other operations?

Should be done in separate patches

dokshina (approved):

PR #9, opened 2020-09-23
CI: support of github actions

olegrok:
This patch adds initial support of GitHub Actions CI


.github/workflows/test_on_push.yaml
```diff
@@ -0,0 +1,23 @@
...
+    branches:
+      - '*'
+jobs:
+  run_tests:
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@master
+      - name: Install requirements
+        run: |
+          sudo apt install -y git gcc cmake make unzip curl
```
tsafin, line 16: No need to install majority of those tools because they are part of preinstalled software on GitHub actions runners https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

tsafin (changes requested):

.github/workflows/test_on_push.yaml
```diff
@@ -0,0 +1,23 @@
+name: Run tests on any Push event
+
+on:
```
tsafin, line 3: Also I'd suggest:
- to extend the list of events with pull-requests
- and to simplify conditions a bit
https://github.com/tsafin/tarantool-merge/blob/206281d3b52035d534d0e5b3fbff185ceaefbaf4/.github/workflows/luarocks-build.yml#L3

.github/workflows/test_on_push.yaml
olegrok: Thanks for your comments.

I don't find gcc and make in this list. I left only them.

.github/workflows/test_on_push.yaml
olegrok: Thanks! Changed.

.github/workflows/test_on_push.yaml
tsafin: They are part of `build-essentials`

.github/workflows/test_on_push.yaml
```diff
@@ -0,0 +1,21 @@
+name: Run tests
+
+on: [push, pull_request]
+
+jobs:
+  run_tests:
+    strategy:
+      matrix:
+        tarantool-version: ["1.10", "2.3", "2.4", "2.5"]
```
tsafin, line 9: I'd recommend to always use `fail-fast: false` strategy, so all invoked job runners would be given chance to complete, regardless of failure of otehr runners

tsafin (changes requested):

tsafin (approved):

dokshina:
Looks great! 

I think, that `.gitlab-ci.yml` can be removed now.

I'm waiting for Tarantool Enterprise tests.

dokshina (approved):

Issue #7, opened 2020-09-23
Support replace and upsert operations


akudiyar:
"Replace" operation is more effective than "insert" for batch operations and is used in high-level connectors like Spring Data connector.

Upsert operation is a more effective counterpart for the "update" operation for the same purposes.

Both operations are supported in the box API, so they should exist for closer compatibility with this API.

PR #11, opened 2020-09-23
Support replace and upsert operations

savolgin:
Issue #7

### DIFF
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -48,10 +48,10 @@ function utils.flatten(object, space_format, bucket_id)
             if not field_format.is_nullable and value == nil then
                 return nil, FlattenError:new("Field %q isn't nullable", field_format.name)
             end
-        else
-            if bucket_id ~= nil then
-                value = bucket_id
-            end
+        end
+
+        if bucket_id ~= nil and field_format.name == 'bucket_id' then
+            value = bucket_id
         end
 
         tuple[fieldno] = value
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -36,7 +36,7 @@ end
 
 local system_fields = { bucket_id = true }
 
-function utils.flatten(object, space_format)
+function utils.flatten(object, space_format, bucket_id)
     if object == nil then return nil end
 
     local tuple = {}
@@ -47,6 +47,10 @@ function utils.flatten(object, space_format)
         if not system_fields[field_format.name] then
             if not field_format.is_nullable and value == nil then
                 return nil, FlattenError:new("Field %q isn't nullable", field_format.name)
+            end
+        else
+            if bucket_id ~= nil then
+                value = bucket_id
             end
         end
 
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -3,6 +3,7 @@ local errors = require('errors')
 
 local FlattenError = errors.new_class("FlattenError", {capture_stack = false})
 local UnflattenError = errors.new_class("UnflattenError", {capture_stack = false})
+local ParseOperationsError = errors.new_class('ParseOperationsError',  {capture_stack = false})
 
 local utils = {}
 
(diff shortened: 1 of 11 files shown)


olegrok:
Thanks for your patch, please drop the first commit (Support tuples as result type) from current pull request, it's not related to "Support replace and upsert operations"

dokshina:
Thanks a lot!

Please, alseo describe new operations in README.

savolgin:
> 
> 
> Please, alseo describe new operations in README.

redme updated

> 
> 
> Thanks for your patch, please drop the first commit (Support tuples as result type) from current pull request, it's not related to "Support replace and upsert operations"

dropped

crud/replace.lua
```diff
@@ -0,0 +1,103 @@
...
+
+    opts = opts or {}
+
+    local space = utils.get_space(space_name, vshard.router.routeall())
+    if space == nil then
+        return nil, ReplaceError:new("Space %q doesn't exists", space_name)
+    end
+
+    -- compute default buckect_id
+    local tuple, err = utils.flatten(obj, space:format())
```
olegrok, line 64: You use `space:format()` call several times. Seems you could save first result to variable and reuse then

crud/replace.lua
```diff
@@ -0,0 +1,103 @@
...
+
+    local key = utils.extract_key(tuple, space.index[0].parts)
+
+    local bucket_id = vshard.router.bucket_id_strcrc32(key)
+    local replicaset, err = vshard.router.route(bucket_id)
+    if replicaset == nil then
+        return nil, ReplaceError:new("Failed to get replicaset for bucket_id %s: %s", bucket_id, err.err)
+    end
+
+    obj = table.copy(obj)
```
olegrok, line 77: It could decrease performance
But you've already "utils.flatten" function that copies all values from tuple.

You need something as:
```
obj.bucket_id = bucket_id
local tuple = flatten(...)
obj.bucket_id = nil
```

Or even you could pass "bucket_id" to flatten function

crud/upsert.lua
```diff
@@ -0,0 +1,112 @@
...
+    local results, err = call.rw(UPSERT_FUNC_NAME, {space_name, tuple, operations}, {
+        replicasets = {replicaset},
+        timeout = opts.timeout,
+    })
+
+    if err ~= nil then
+        return nil, UpsertError:new("Failed to upsert: %s", err)
+    end
+
+    local tuple = results[replicaset.uuid]
```
olegrok, line 103: Vinyl returns nothing on update/upsert/delete - it should be considered

olegrok:
I suggest you to add tests for both engines - memtx and vinyl.

See example - https://github.com/tarantool/crud/pull/4

crud/replace.lua
savolgin: Variable added

crud/replace.lua
savolgin: added param to flatten function

crud/upsert.lua
savolgin: Tests updated for vinyl engine

crud/replace.lua
```diff
@@ -0,0 +1,101 @@
...
+        replicasets = {replicaset},
+        timeout = opts.timeout,
+    })
+
+    if err ~= nil then
+        return nil, ReplaceError:new("Failed to replace: %s", err)
+    end
+
+    local tuple = results[replicaset.uuid]
+    local object, err = utils.unflatten(tuple, space:format())
```
olegrok, line 93: Here you can use "space_format" variable

crud/upsert.lua
```diff
@@ -0,0 +1,109 @@
...
+    })
+
+    opts = opts or {}
+
+    local space = utils.get_space(space_name, vshard.router.routeall())
+    if space == nil then
+        return nil, UpsertError:new("Space %q doesn't exists", space_name)
+    end
+
+    local operations, err = utils.convert_operations(user_operations, space:format())
```
olegrok, line 67: Here you still use space:format() several times

savolgin: fixed

crud/replace.lua
savolgin: fixed

crud/upsert.lua
olegrok: Oh, sorry. I was confused. Seems upsert returns nothing even for "memtx". Let's drop "unflatten" function here. Actually it doesn't do anything useful.

crud/common/utils.lua
```diff
@@ -47,6 +48,10 @@ function utils.flatten(object, space_format)
             if not field_format.is_nullable and value == nil then
                 return nil, FlattenError:new("Field %q isn't nullable", field_format.name)
             end
+        else
+            if bucket_id ~= nil then
```
olegrok, line 52: I think general approach a bit better.
```
if field_format.name == 'bucket_id' then
    value = bucket_id
end
```

crud/upsert.lua
olegrok: Thanks!

olegrok:
See my last comment. After fixing it - LGTM

crud/common/utils.lua
savolgin: cnaged

crud/upsert.lua
savolgin: dropped

olegrok (approved):

olegrok:
LGTM. Seems you need to update changelog as well

savolgin:
Change log updated

akudiyar (approved):

crud/common/utils.lua
```diff
@@ -49,6 +50,10 @@ function utils.flatten(object, space_format)
             end
         end
 
+        if bucket_id ~= nil and field_format.name == 'bucket_id' then
```
dokshina, line 53: a nit: since we already have `system_fields` conception, maybe it can be

```lua
function utils.flatten(object, space_format, system_values)
    system_values = system_values or {}
    ...
    if system_values.bucket_id ~= nil and field_format.name == 'bucket_id' then ...
    -- if sometimes we have more system fields, it can be done in a cycle
    ...
end
```

dokshina (approved):
Thanks a lot for your patch!

Only one: please wait for new CI based on GitHub Actions is merged and rebase on it.
Then, it will be merged.

Issue #14, opened 2020-09-24
Fix flaky "select_big.test_le_condition_with_index"


olegrok:
Failed jobs:
  - https://github.com/tarantool/crud/runs/1159713619?check_suite_focus=true
  - https://github.com/tarantool/crud/runs/1155584650

dokshina:
I think that it would be fixed within #16.

PR #17, opened 2020-09-28
More tests less bugs

dokshina:
Closes https://github.com/tarantool/crud/issues/14


crud/common/collations.lua
```diff
@@ -0,0 +1,27 @@
+local json = require('json')
+
+local COLLATION_NAME_FN = 2
+
+local collations = {}
+
+collations.NONE = 'none'
```
olegrok, line 7: https://www.tarantool.io/en/doc/2.5/book/box/data_model/#collations

```
By default, when Tarantool compares strings, it uses what we call a “binary” collation.
```

Is there any difference between 'none' and 'binary' collation?
Should we add some tests for it?

olegrok: Seems it's the same
```
tarantool> box.space._collation:select()
---
- - [0, 'none', 1, 'BINARY', '', {}]
  - [1, 'unicode', 1, 'ICU', '', {'strength': 'tertiary'}]
  - [2, 'unicode_ci', 1, 'ICU', '', {'strength': 'primary'}]
  - [3, 'binary', 1, 'BINARY', '', {}]
```

crud/common/collations.lua
```diff
@@ -0,0 +1,27 @@
...
+    if index_part.collation ~= nil then
+        return index_part.collation
+    end
+
+    if index_part.collation_id == nil then
+        return collations.NONE
+    end
+
+    local collation_tuple = box.space._collation:get(index_part.collation_id)
+    assert(collation_tuple ~= nil, "Unknown collation_id: " .. json.encode(index_part.collation_id))
```
olegrok, line 21: You should check is collation supported or not.

Currently it's possible to support only binary, unicode and unicode_ci.

But not all 277:
```
tarantool> #box.space._collation:select()
---
- 277
...
```

crud/select/comparators.lua
```diff
@@ -158,23 +169,29 @@ local array_cmp_funcs_by_operators = {
     [operators.GE] = array_ge,
 }
 
-function comparators.gen_func(operator, key_parts)
-    local cmp_func = array_cmp_funcs_by_operators[operator]
+function comparators.get_cmp_operator(tarantool_iter)
+    local cmp_operator = cmp_operators_by_tarantool_iter[tarantool_iter]
+    assert(cmp_operator ~= nil, 'Unknown Tarantool iterator %q', tarantool_iter)
```
olegrok, line 174: Unsupported or unknown

```
tarantool> box.index
---
- EQ: 0
  GT: 6
  BITS_ALL_SET: 7
  GE: 5
  NEIGHBOR: 11
  OVERLAPS: 10
  REQ: 1
  BITS_ALL_NOT_SET: 9
  BITS_ANY_SET: 8
  LT: 3
  LE: 4
  ALL: 2
...
```

There are some iterator types for bitset/rtree indexes

crud/common/collations.lua
dokshina: It is checked in the code that uses collation, e.g. `select.filters`. 

I sure that it's a good approach. The `collations` module responsibility is only conversion from collcation ID to collation.

crud/common/collations.lua
dokshina: I just wanted to decrease number of `collation ~= nil and collation ~= collations.NONE` comparisons.

crud/select/comparators.lua
dokshina: Unknown, you are right.

dokshina:
Was separated to #18, #19, #20, #21.

PR #18, opened 2020-09-29
Bug: Equal Condition

### LINKED ISSUE #23: [2pt] Bug: Select with equal conditions
* Tuple comparator should use scan key merged w/ primary key
* Comparator operator should be one of > or < for tuples sorting
* Seems that every select by equal condition is interpreted as a select with a fully specified sharding key

dokshina:
Closes #23

### DIFF
--- a/crud/select/comparators.lua
+++ b/crud/select/comparators.lua
@@ -167,6 +167,12 @@ local array_cmp_funcs_by_operators = {
     [operators.GE] = array_ge,
 }
 
+--[=[
+    Each tarantool iterator returns tuples in a strictly defined order
+    (scan key merged with primary key is used to guarantee that)
+    GE, GT and EQ interators return tuples in ascending order
+    LE, LT and REQ - in descending
+--]=]
 function comparators.get_cmp_operator(tarantool_iter)
     local cmp_operator = cmp_operators_by_tarantool_iter[tarantool_iter]
     assert(cmp_operator ~= nil, 'Unknown Tarantool iterator %q', tarantool_iter)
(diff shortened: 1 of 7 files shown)


crud/select/comparators.lua
```diff
@@ -150,6 +150,15 @@ local function gen_array_cmp_func(target, key_parts)
     end
 end
 
+local cmp_operators_by_tarantool_iter = {
```
olegrok, line 153: Please add a comment with description. At first glance I don't understand why it's needed

crud/select/comparators.lua
```diff
@@ -158,23 +167,29 @@ local array_cmp_funcs_by_operators = {
...
+    assert(cmp_operator ~= nil, 'Unknown Tarantool iterator %q', tarantool_iter)
+
+    return cmp_operator
+end
 
+function comparators.gen_func(cmp_operator, key_parts)
+    local cmp_func = array_cmp_funcs_by_operators[cmp_operator]
     if cmp_func == nil then
-        return nil, ComparatorsError:new('Unsupported operator %q', operator)
+        return nil, ComparatorsError:new('Unsupported operator %q', cmp_operator)
```
olegrok, line 180: Still I'm sure "Unsupported" and "Unknown" iterators should be splitted

crud/select/comparators.lua
dokshina: Added a comment to `comparators.get_cmp_operator` function.

crud/select/comparators.lua
dokshina: It's not a check that iterator isn't exists, it's an only check that it can be used for tuples comparator, so it's `Unsupported`.

olegrok (approved):

PR #32, opened 2020-10-01
Refactor select plan

### LINKED ISSUE #30: [4pt] Select Plan Rework
### RFC

Select plan is a table that contains all information we need to know to perform select.
It's created based on user-specified conditions and such options as `first` and `after`.
Select plan is created on both storage and router. 

Now select plan looks like this:

```lua
plan = {
    scanner = {
        space_name = scan_space_name,
        index_id = scan_index.id,
        index_name = scan_index.name,
        iter = scan_iter,
        value = scan_value,
        condition_num = scan_condition_num,
        operator = scan_operator, -- redundant
        limit = scan_limit,
        after_tuple = scan_after_tuple,
    },
    filter_conditions = { ... }, -- used by storage,
    is_scan_by_full_sharding_key_eq, -- used by router
}
```

I don't like current plan structure for some reasons:

* we compute `filter_conditions` on router, but it's used only by storage;
* storage knows everything about the whole select query that is bad (I would like storage to be very simple - it just selects specified number of tuples (`batch_size`, not `limit` of `first`)  from specified space with specified iterator and filters it by filter conditions;
* storage- and router-side logic has many intersections that causes changing both router and storage code to implement new features (which are often affects router logic).

I suggest next approach:

* select plan is separated into three parts: common part for router and storage;
* "common" part is computed by router and is passed to storage on the call;
* storage uses it to compute `filter_conditions`
* computing `after_tuple` for storage is a router responsibility (in fact, it's a cursor for all calls except the first one, but now storage selects one value between `scan_value` and `after_tuple` on every call);
* storage accepts `batch_size` option that specifies the count of tuples to be returned;
* router uses "common" part to check `is_scan_by_full_sharding_key_eq` and `only_one_value_is_needed` conditions;
* `first` option is processed by router (described below)

In details:

* user calls `crud.select(space_name, conditions, { first = ..., after = ... })` 

* router computes `plan` based on conditions:

  ```lua
  plan = {
       conditions = ...,
       space_name = ...,
       index_id = ...,
       scan_value = ...,
       scan_condition_num = ...,
       iter = ..., -- based on scan condition
  }
  ```

* router applies specified `first`:
  * if `first` is negative, `plan.iter` is inversed;
  * `limit` (the whole number of tuples to select) is set to `first` absolute value;
  * if `first` is negative, `reversed`  flag is set to `true`
* router checks if only one value is needed and sets `limit` to 1 and `plan.iter` to `REQ`;
* router checks if select by fully specified sharding key is performed and decides if only one storage should be called;

So, router has the final version of the `plan` and some values that are used only by router.

* router computes `batch_size` on each call (`min(batch_size, limit - count_tuples)`)
* router calls select on storages with `(plan, { batch_size = batch_size })` and aggregates received tuples.

dokshina:
I didn't forget about

- [x] Tests

Closes #30

### DIFF
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -104,6 +104,23 @@ function utils.merge_primary_key_parts(key_parts, pk_parts)
     return merged_parts
 end
 
+function utils.extract_subkey(key_value, key_parts, subkey_parts)
+    local value_parts_by_fieldnos = {}
+    for i, part in ipairs(key_parts) do
+        value_parts_by_fieldnos[part.fieldno] = key_value[i]
+    end
+
+    -- check that sharding key is included in the scan index fields
+    local subkey_value = {}
+    for _, subkey_part in ipairs(subkey_parts) do
+        local fieldno = subkey_part.fieldno
+        local field_value = value_parts_by_fieldnos[fieldno]
+        table.insert(subkey_value, field_value)
+    end
+
+    return subkey_value
+end
+
 local __tarantool_supports_fieldpaths
 local function tarantool_supports_fieldpaths()
     if __tarantool_supports_fieldpaths ~= nil then
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -104,27 +104,6 @@ function utils.merge_primary_key_parts(key_parts, pk_parts)
     return merged_parts
 end
 
-function utils.extract_sharding_key_from_scan_key(scan_key, scan_index, sharding_index)
-    if scan_index.id == sharding_index.id then
-        return scan_key
-    end
-
-    local scan_key_parts_by_fieldnos = {}
-    for i, scan_key_part in ipairs(scan_index.parts) do
-        scan_key_parts_by_fieldnos[scan_key_part.fieldno] = scan_key[i]
-    end
-
-    -- check that sharding key is included in the scan index fields
-    local sharding_key = {}
-    for _, sharding_key_part in ipairs(sharding_index.parts) do
-        local fieldno = sharding_key_part.fieldno
-        local field_value = scan_key_parts_by_fieldnos[fieldno]
-        table.insert(sharding_key, field_value)
-    end
-
-    return sharding_key
-end
-
 local __tarantool_supports_fieldpaths
 local function tarantool_supports_fieldpaths()
     if __tarantool_supports_fieldpaths ~= nil then
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -104,21 +104,25 @@ function utils.merge_primary_key_parts(key_parts, pk_parts)
     return merged_parts
 end
 
-function utils.extract_subkey(key_value, key_parts, subkey_parts)
-    local value_parts_by_fieldnos = {}
-    for i, part in ipairs(key_parts) do
-        value_parts_by_fieldnos[part.fieldno] = key_value[i]
+function utils.extract_sharding_key_from_scan_key(scan_key, scan_index, sharding_index)
+    if scan_index.id == sharding_index.id then
+        return scan_key
+    end
+
+    local scan_key_parts_by_fieldnos = {}
+    for i, scan_key_part in ipairs(scan_index.parts) do
+        scan_key_parts_by_fieldnos[scan_key_part.fieldno] = scan_key[i]
     end
 
     -- check that sharding key is included in the scan index fields
-    local subkey_value = {}
-    for _, subkey_part in ipairs(subkey_parts) do
-        local fieldno = subkey_part.fieldno
-        local field_value = value_parts_by_fieldnos[fieldno]
-        table.insert(subkey_value, field_value)
+    local sharding_key = {}
+    for _, sharding_key_part in ipairs(sharding_index.parts) do
+        local fieldno = sharding_key_part.fieldno
+        local field_value = scan_key_parts_by_fieldnos[fieldno]
+        table.insert(sharding_key, field_value)
     end
 
-    return subkey_value
+    return sharding_key
 end
 
 local __tarantool_supports_fieldpaths
--- a/crud/select.lua
+++ b/crud/select.lua
@@ -25,11 +25,13 @@ local SELECT_FUNC_NAME = '__select'
 
 local DEFAULT_BATCH_SIZE = 100
 
-local function call_select_on_storage(space_name, index_id, scan_value, iter, conditions, opts)
-    checks('string', 'number', '?table', 'number', 'table', {
-        scan_condition_num = '?number',
+local function call_select_on_storage(space_name, index_id, conditions, opts)
+    checks('string', 'number', '?table', {
+        scan_value = 'table',
         after_tuple = '?table',
+        iter = 'number',
         limit = 'number',
+        scan_condition_num = '?number',
     })
 
     local space = box.space[space_name]
@@ -43,7 +45,7 @@ local function call_select_on_storage(space_name, index_id, scan_value, iter, co
     end
 
     local filter_func, err = select_filters.gen_func(space, conditions, {
-        iter = iter,
+        iter = opts.iter,
         scan_condition_num = opts.scan_condition_num,
     })
     if err ~= nil then
@@ -51,8 +53,10 @@ local function call_select_on_storage(space_name, index_id, scan_value, iter, co
     end
 
     -- execute select
-    local tuples, err = select_executor.execute(space, index, scan_value, iter, filter_func, {
+    local tuples, err = select_executor.execute(space, index, filter_func, {
+        scan_value = opts.scan_value,
         after_tuple = opts.after_tuple,
+        iter = opts.iter,
         limit = opts.limit,
     })
     if err ~= nil then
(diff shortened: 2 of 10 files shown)


crud/select.lua
```diff
@@ -22,29 +23,36 @@ local select_module = {}
 
 local SELECT_FUNC_NAME = '__select'
 
-local function call_select_on_storage(space_name, conditions, opts)
-    checks('string', '?table', {
-        limit = '?number',
+local function call_select_on_storage(space_name, index_id, scan_value, iter, conditions, opts)
```
olegrok, line 26: As for me "iter" should be part of opts

`box.space[*space_name*].index[*index_id*]:pairs(*scan_value*, *{<options included iter>}*)`. Or it could be "scan_opts" if you want separate it from batch_size/after_tuple.

crud/select.lua
```diff
@@ -22,29 +23,36 @@ local select_module = {}
...
+    checks('string', 'number', '?table', 'number', 'table', {
+        scan_condition_num = '?number',
         after_tuple = '?table',
+        batch_size = 'number',
```
olegrok, line 30: I believe it should be limit.
Also `batch_size = 1ULL` will raise because `1ULL/LL` is cdata

crud/select.lua
```diff
@@ -86,14 +97,8 @@ local function select_iteration(space_name, conditions, opts)
...
 
-local function get_replicasets_to_select_from(plan, all_replicasets)
-    if not plan.is_scan_by_full_sharding_key_eq then
-        return all_replicasets
-    end
-
-    plan.scanner.limit = 1
-
-    local bucket_id = vshard.router.bucket_id_strcrc32(plan.scanner.value)
+local function get_replicaset_by_scan_value(scan_value)
```
olegrok, line 100: It confuses me. It's get_replicasets_by_sharding_key.

To be `by_scan_value` this function should include logic that scan_value is full specified sharding key and returns routeall if it's not so. But it doesn't.

crud/select.lua
```diff
@@ -137,32 +142,33 @@ local function build_select_iterator(space_name, user_conditions, opts)
...
+    local replicasets_to_select = replicasets
 
-    local scan_index = space.index[plan.scanner.index_id]
+    local scan_index = space.index[plan.index_id]
     local primary_index = space.index[0]
+    if select_plan.is_scan_by_full_sharding_key_eq(plan, scan_index, primary_index) then
+        limit = 1
+        plan.iter = box.index.REQ
+
+        replicasets_to_select = get_replicaset_by_scan_value(plan.scan_value)
```
olegrok, line 164: yes, it's `full_sharding_key`, not simple scan_value already

crud/select.lua
```diff
@@ -137,32 +142,33 @@ local function build_select_iterator(space_name, user_conditions, opts)
...
+    end
+
+    -- set after tuple
+    local after_tuple = utils.flatten(opts.after, space_format)
```
olegrok, line 168: Not sure it's a good place for unflatten...

crud/select/executor.lua
```diff
@@ -11,13 +11,12 @@ local ExecuteSelectError = errors.new_class('ExecuteSelectError')
 
 local executor = {}
 
-local function scroll_to_after_tuple(gen, param, state, space, scanner)
-    local scan_index = space.index[scanner.index_id]
+local function scroll_to_after_tuple(gen, param, state, space, scan_index, iter, after_tuple)
```
olegrok, line 14: You don't need all of this options.

`gen, param, state` could be reduced to `gen`

```lua
tarantool> s = box.space.log:pairs()
---
...

tarantool> s(s.param, s.state)
---
- <iterator state>
- tuple
...
```

Seven arguments... It's too big.

crud/select/executor.lua
```diff
@@ -31,52 +30,51 @@ local function scroll_to_after_tuple(gen, param, state, space, scanner)
...
             return tuple
         end
     end
 end
 
-function executor.execute(plan)
-    local scanner = plan.scanner
+function executor.execute(space, index, scan_value, iter, filter_func, opts)
+    checks('table', 'table', '?table', 'number', 'function', {
+        after_tuple = '?cdata|table',
```
olegrok, line 41: Is "cdata" needed here?

tuple is converted to lua table after transferring via netbox.

crud/select/executor.lua
```diff
@@ -31,52 +30,51 @@ local function scroll_to_after_tuple(gen, param, state, space, scanner)
...
+        batch_size = '?number',
```
olegrok, line 42: number|cdata

olegrok: Or even `number|uint64|int64 `

olegrok: So, not sure checks really needed here it's internal module

crud/select/executor.lua
```diff
@@ -31,52 +30,51 @@ local function scroll_to_after_tuple(gen, param, state, space, scanner)
...
+    local gen,param,state = index:pairs(value, {iterator = iter})
 
-    if scanner.after_tuple ~= nil then
+    if opts.after_tuple ~= nil then
         local err
-        tuple, err = scroll_to_after_tuple(gen, param, state, space, scanner)
+        tuple, err = scroll_to_after_tuple(gen, param, state, space, index, iter, opts.after_tuple)
         if err ~= nil then
-            return nil, ExecuteSelectError:new("Failed to scroll to the last tuple: %s", err)
+            return nil, ExecuteSelectError:new("Failed to scroll to the after tuple: %s", err)
```
olegrok, line 77: `the after tuple`... If you mean after_tuple you need underscore here `after tuple` -> `after_tuple`

crud/select.lua
dokshina: Please, se the commit below. I've fixed full scan key case processing.

crud/select/executor.lua
dokshina: Yes, I just use `space:frommap` in executor tests.

dokshina: BTW, why not? Executor knows nothing about router and other netbox staff.

crud/select/executor.lua
dokshina: Yes, but it's very helpful for local development.
Moreover, all functions have a lot of arguments and different opts.

crud/select.lua
dokshina: Fixed in the commit below.

crud/select/executor.lua
olegrok: > very helpful for local development.

Then do something like:
```lua
local checks = require('checks')
if os.getenv('DEV') ~= '1' then
    checks = function() end
end
```

And run all your tests with "DEV=1" option.

crud/common/utils.lua
```diff
@@ -104,6 +104,23 @@ function utils.merge_primary_key_parts(key_parts, pk_parts)
     return merged_parts
 end
 
+function utils.extract_subkey(key_value, key_parts, subkey_parts)
```
olegrok, line 107: `subkey`

Please don't introduce new terms when it doesn't really needed.

crud/select.lua
```diff
@@ -86,14 +99,8 @@ local function select_iteration(space_name, conditions, opts)
...
 
-local function get_replicasets_to_select_from(plan, all_replicasets)
-    if not plan.is_scan_by_full_sharding_key_eq then
-        return all_replicasets
-    end
-
-    plan.scanner.limit = 1
-
-    local bucket_id = vshard.router.bucket_id_strcrc32(plan.scanner.value)
+local function get_replicasets_by_sharding_key(sharding_key_value)
```
olegrok, line 102: `sharding_key_value`

`key_value`

Why not just `sharding_key`?

crud/select.lua
olegrok: I believe we should pass raw tuple to this function as option...

But ok, feel free to ignore this comment

crud/select/executor.lua
olegrok: We should not customize the code for tests. We should just write tests that don't require changes in code (if it's not a bug of course ;) )

BTW, frommap():totable() solves your problem.

> BTW, why not? Executor knows nothing about router and other netbox staff.

I've just noticed that tuple as cdata is impossible input for this function.

crud/select.lua
dokshina: Since crud returns map, user will use it as an `after` value for selecting next batch of data.

crud/select/executor.lua
olegrok: Well, just for you `checks('tuple')` was implemented - https://github.com/tarantool/checks/commit/e4bd1989db6e4e831f7611ddf2d5d20da074b44e

dokshina: It's so cute, I'm waiting for release!

crud/select.lua
```diff
@@ -137,32 +146,36 @@ local function build_select_iterator(space_name, user_conditions, opts)
...
+    local replicasets_to_select = replicasets
 
-    local scan_index = space.index[plan.scanner.index_id]
+    local scan_index = space.index[plan.index_id]
     local primary_index = space.index[0]
+    if select_plan.is_scan_by_full_sharding_key_eq(plan, scan_index, primary_index) then
+        total_tuples_count = 1
+        plan.iter = box.index.REQ
+
+        local sharding_key_value = utils.extract_subkey(
```
olegrok, line 168: Simply `sharding_key`

You don't use `parts_value`, `scan_value_value` or `space_value` in any places

crud/select/executor.lua
dokshina: See #35.

crud/select/executor.lua
dokshina: See #35.

crud/common/utils.lua
```diff
@@ -104,6 +104,27 @@ function utils.merge_primary_key_parts(key_parts, pk_parts)
     return merged_parts
 end
 
+function utils.extract_sharding_key_from_scan_key(scan_key, scan_index, sharding_index)
```
olegrok, line 107: It does the same as `is_scan_by_full_sharding_key_eq`

olegrok (approved):

PR #38, opened 2020-10-02
Use "merger" for fetch&sort tuples on router

### LINKED ISSUE #33: [4pt] Investigate merger/key_def embedding into current implementation
Write test branch for 2.4+ Tarantool versions until https://github.com/tarantool/tarantool/issues/5273 and https://github.com/tarantool/tarantool/issues/5384 are not implemented

olegrok:
Before this patch we used hand-crafted heap to sort tuples on
router. This patch removes it and changes to "merger" module
written in C.

"Merger" is available since Tarantool 2.2. Later "tuple-merger"
was introduced it's a version that is compatible with some
Tarantool 1.10 versions. In fact support several Tarantool
versions is quite complex task. Some msgpack helpers is not
avaliable in Tarantool 1.10 to solve this problem small
compatibility lawer is implemented.
Also "tuple-keydef" doesn't support collation_id options and it's
not so for built-in "key_def" module. We need to normalize index
parts and change "collation_id" to "collation" when create
"key_def" instance.

For versions where merger is not supported/available yet
(before 1.10.8 and before 2.3.3) we use previous approach with
lua-implemented heap.

Closes #33


olegrok:
https://github.com/tarantool/crud/tree/perf-test

Before patch:
```
SELECT
Fibers count - 200
Connection count - 10
Timeout  - 30.000000
Requests - 102824
Errors - 0
RPS - 3427.466667
```

After patch:
```
SELECT
Fibers count - 200
Connection count - 10
Timeout  - 30.000000
Requests - 168745
Errors - 0
RPS - 5624.833333
Ok
```

crud/select.lua
```diff
@@ -222,31 +198,32 @@ function select_module.pairs(space_name, user_conditions, opts)
...
 
-    local gen = function(_, iter)
-        if not iter:has_next() then
+    local tuple, _
+
+    if opts.use_tomap ~= true then
+        return iter.iter:pairs()
+    end
+
+    local merger_gen = iter.iter:pairs()
```
dokshina, line 207: I would like to rename `iter` field to `tarantool_iter`?
`plan.tarantool_iter`, `iter.tarantool_iter:pairs()`

dokshina: It looks like `Iterator` became `Merger`, do you think so?

crud/select.lua
```diff
@@ -222,31 +198,32 @@ function select_module.pairs(space_name, user_conditions, opts)
...
 
         return iter, result
     end
 
-    return gen, nil, iter
+    return gen
 end
 
-function select_module.call(space_name, user_conditions, opts)
+local function select_module_call_xc(space_name, user_conditions, opts)
```
dokshina, line 226: What does `_xc` mean?

crud/select/iterator.lua
```diff
@@ -1,196 +1,223 @@
...
+local collations = require('crud.common.collations')
+
+local key_def_lib
+local merger_lib
+
+if pcall(require, 'tuple.merger') then
+    merger_lib = require('tuple.merger')
+    key_def_lib = require('tuple.keydef')
+elseif pcall(require, 'merger') then
+    log.info('Impossible to load "tuple-merger" module. Use built-in "merger"')
```
dokshina, line 14: Maybe `built-in merger is used`?

crud/select/iterator.lua
```diff
@@ -1,196 +1,223 @@
...
+    merger_lib = require('merger')
+    key_def_lib = require('key_def')
+else
+    error('Seems your Tarantool version (' .. _TARANTOOL ..
+            ') does not support "tuple-merger" or "merger" modules')
+end
 
-        heap = Heap.new({ comparator = opts.comparator }),
-        tuples_count = 0,
+local SELECT_FUNC_NAME = '_crud.select_on_storage'
```
dokshina, line 22: I seems that this function name should be defined in `select` module (where function is defined) and passed to constructor.

crud/select.lua
olegrok: https://github.com/tarantool/tarantool/search?q=xc

crud/select/iterator.lua
```diff
@@ -1,196 +1,223 @@
...
+        local state = {future = future}
+        local source = merger_lib.new_buffer_source(fetch_chunk, context, state)
+        table.insert(merger_sources, source)
     end
 
-    return tuple
+    local merger = merger_lib.new(key_def, merger_sources, {
+        reverse = reverse_iterators[opts.iter],
+    })
+    return merger
```
dokshina, line 218: `Iterator.new` returns `merger`.

crud/select.lua
dokshina: Ok, but it's not hte only one function that can raise in crud, but the only one with `_xc` suffix.

olegrok: - This is not a public function
- Other functions returns support `nil, err` contract.

.github/workflows/test_on_push.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
       github.event_name == 'pull_request' && github.event.pull_request.head.repo.owner.login != 'tarantool'
     strategy:
       matrix:
-        tarantool-version: ["1.10", "2.3", "2.4", "2.5"]
+        tarantool-version: ["1.10", "2.1", "2.2", "2.6"]
```
dokshina, line 14: Please, add a comment why exactly these versions was chosen. It's very important.

crud/select/compat/select.lua
```diff
@@ -0,0 +1,199 @@
+local checks = require('checks')
```
dokshina, line 1: Why not `crud/compat/select.lua`?

crud/select/compat/select.lua
```diff
@@ -0,0 +1,199 @@
...
+        local result
+        result, err = utils.unflatten(tuple, iter.space_format)
+        if err ~= nil then
+            error(string.format("Failed to unflatten next object: %s", err))
+        end
+        return result
+    end)
+end
+
+local function select_module_call_xc(space_name, user_conditions, opts)
```
dokshina, line 143: Do other crud modules use `_xc` naming convention?

crud/select/compat/select.lua
dokshina: Please, describe what is the difference between this file and `select_old`.

crud/select/merger.lua
```diff
@@ -0,0 +1,224 @@
...
+
+if pcall(require, 'tuple.merger') then
+    merger_lib = require('tuple.merger')
+    key_def_lib = require('tuple.keydef')
+elseif pcall(require, 'merger') then
+    log.info('Impossible to load "tuple-merger" module. Built-in "merger" is used')
+    merger_lib = require('merger')
+    key_def_lib = require('key_def')
+else
+    error('Seems your Tarantool version (' .. _TARANTOOL ..
```
dokshina, line 18: Please, use `string.format` instead of `..`.

crud/select/merger.lua
```diff
@@ -0,0 +1,224 @@
...
+end
+
+local function bswap_u16(num)
+    return bit.rshift(bit.bswap(tonumber(num)), 16)
+end
+
+-- See
+-- https://github.com/tarantool/tarantool/blob/0ab21ac9eeaaae2aa0aef5e598d374669f96df9e/src/lua/msgpackffi.lua
+-- to understand following hell
+-- This code will work for ALL Tarantool versions
```
dokshina, line 72: will work -> works?

crud/select/compat/select.lua
olegrok: Other functions simply don't raise

crud/select.lua
```diff
@@ -1,35 +1,40 @@
...
 
 local SelectError = errors.new_class('SelectError')
-local GetReplicasetsError = errors.new_class('GetReplicasetsError')
 
-local select_module = {}
+local select_module
+
+-- "merger" segfaults here
+-- See https://github.com/tarantool/tarantool/issues/4954
+if '2' <= _TARANTOOL and _TARANTOOL <= '2.3.3' then
```
dokshina, line 13: What about using smth like `tarantool_supports_fieldpaths`?

crud/select.lua
```diff
@@ -1,35 +1,40 @@
...
+    select_module = require('crud.select.compat.select_old')
+else
+    -- Try to require built-in or external merger
+    local ok, module = pcall(require, 'crud.select.compat.select')
+    if ok then
+        select_module = module
+    else
+        select_module = require('crud.select.compat.select_old')
```
dokshina, line 21: Are you sure that we shouldn't fail here?

crud/select.lua
olegrok: tarantool_merger_doesnt_segfault ?

crud/common/call.lua
```diff
@@ -32,22 +31,19 @@ local function call_impl(vshard_call, func_name, func_args, opts)
...
 
-    local futures_by_replicasets = {}
-    local call_opts = {is_async = true}
+    local futures = {}
     for _, replicaset in pairs(replicasets) do
-        local future = replicaset[vshard_call](replicaset, func_name, func_args, call_opts)
-        futures_by_replicasets[replicaset.uuid] = future
+        local future = replicaset[vshard_call](replicaset, func_name, func_args, {
+            timeout = timeout,
+            is_async = true,
```
olegrok, line 38: `timeout` will be ignored if `is_async=true` is specified. Correct way to handle such case were described below - https://github.com/tarantool/crud/pull/38/files#diff-d0dbf8e86bcf0f98f008ed20030bfdc2c3f106a245d3343f7cf3acac36b6f921L43-L48

See also https://github.com/tarantool/doc/issues/1710

olegrok:
Hi! Thanks for your patch. Could you please rebase on master branch and fix conflicts?

.github/workflows/test_on_push.yaml
olegrok: Comment should be follwing:

```
1.10 - separate tuple-merger module (https://github.com/tarantool/tuple-merger)
2.1 - there is no merger
2.2 - there is a merger but it doesn't work (https://github.com/tarantool/tarantool/issues/4954)
2.3+ - OK
```

crud/common/utils.lua
```diff
@@ -212,8 +212,8 @@ local function add_nullable_fields_recursive(operations, operations_map, space_f
     return operations
 end
 
--- Tarantool < 2.3 has no fields `box.error.NO_SUCH_FIELD_NO` and `box.error.NO_SUCH_FIELD_NAME`.
-if _TARANTOOL >= "2.3" then
+-- Tarantool < 2.1 has no fields `box.error.NO_SUCH_FIELD_NO` and `box.error.NO_SUCH_FIELD_NAME`.
+if _TARANTOOL >= "2.1" then
```
olegrok, line 216: Could you please explain this change?

JSON updates by path was introduced in 2.3 (https://github.com/tarantool/tarantool/commit/9ee5cf82ae579d62543a4bb335dd1e99b00dacf6). I don't understand why do you change it to 2.1.

crud/select.lua
olegrok: Hope Lua-implemented merger won't fail

olegrok: NB:

- crud.select.compat.select - default module with merged
- crud.select.compat.select_old - compatibility module without merger support written in Lua

@AnaNek maybe it would be great to reflect it in commit message

crud/select/compat/select.lua
```diff
@@ -0,0 +1,227 @@
...
+
+    return {
+        merger = merger,
+        space_format = filtered_space_format,
+    }
+end
+
+function select_module.pairs(space_name, user_conditions, opts)
+    checks('string', '?table', {
+        after = '?table|cdata',
```
olegrok, line 119: it's better to use `?table|tuple` instead of `*|cdata`

olegrok: The same for all places where tuple expected.

crud/select/merger.lua
```diff
@@ -0,0 +1,235 @@
...
+    end
+
+    return result
+end
+
+local function get_key_def(replicasets, space_name, field_names, index_name)
+    -- Get requested and primary index metainfo.
+    local conn = select(2, next(replicasets)).master.conn
+    local space = conn.space[space_name]
+    local space_format = space:format()
```
olegrok, line 49: Let's put this call after
```lua
if key_def_cache[index_by_field_names] ~= nil then
    return key_def_cache[index_by_field_names]
end
```

crud/select/merger.lua
```diff
@@ -0,0 +1,235 @@
...
+    local index = space.index[index_name]
+    local index_by_field_names = {index, field_names}
+
+    if key_def_cache[index_by_field_names] ~= nil then
+        return key_def_cache[index_by_field_names]
+    end
```
olegrok, line 55: It looks a bit tricky. You need to calculate some hash (string). Otherwise this table will be recreated at each call - finally, it will work as without any cache.

Alternative - use `msgpack.encode(index_by_field_names)` as key.

crud/select/merger.lua
```diff
@@ -0,0 +1,235 @@
...
+    local replicaset = context.replicaset
+    local future = state.future
+
+    -- The source was entirely drained.
+    if future == nil then
+        return nil
+    end
+
+    -- Wait for requested data.
+    local res, err = future:wait_result()
```
olegrok, line 167: Let's add some timeout here.

crud/common/utils.lua
AnaNek: In Tarantool version >= 2.1 there is no `box.error.NO_SUCH_FIELD` error code, there are `box.error.NO_SUCH_FIELD_NO` and `box.error.NO_SUCH_FIELD_NAME` codes (https://github.com/tarantool/tarantool/blob/2.1/src/box/errcode.h#L92 , https://github.com/tarantool/tarantool/blob/2.1/src/box/errcode.h#L208). That's why `test_intermediate_nullable_fields_update` failed for Tarantool version < 2.3.

olegrok: Well, seems all this time we had crud broken at 2.1.

I suggest to move this change into separate patch and corresponding update of targets in CI as well.

dokshina (approved):
Great patch!

Issue #6, opened 2020-09-23
Support tuples as result type


akudiyar:
Now the results of the operations are returned as maps (Lua tables):

```
crud.insert('customers', {
    id = 1, name = 'Elizabeth', age = 23,
})
---
- bucket_id: 7614
  age: 23
  name: Elizabeth
  id: 1
...
```

If the result is further used in some connectors, the tuple-to-map conversion may be unnecessary, especially for the purpose of partial compliance with the box API, which has an option for returning maps and returns tuples by default.

Proposal: implement an option "tuples_as_map" with default value "true", which may be passed along with the options of each operation. In case if this option is set to "false", the tuple-to-map conversion is not performed and the result tuples are returned as-is.

Totktonada:
NB: A tuple has `:tomap()` method. Maybe it would be good to borrow the option name from it.

dokshina:
I like the idea of using `tomap` in the option name. (`use_tomap`, `tuples_tomap` etc.)
 
Moreover, (in the future) I would like to have two functions for insert - `insert_raw` (which accepts tuples) and `insert` (which accepts objects). Rather than other operations, insert accepts tuples or objects, not only returns.

I like the idea to use tuples by default (as @knazarov suggested).

I don't see any reason to discuss on #10 PR until we have a fully discussed and accepted by @mtrempoltsev and @knazarov  RFC here. 

My suggestion (within #10 ) is:

1. `CRUD` operations return objects by default (until it's not possible to return tuple with `:tomap()` method). 
2. `CRUD` operations return tuples (firstly, a simple `table` objects) if `use_tomap/tuples_tomap` option is specified.

Later, all operations should return `box.tuple` objects by default. 
In this case, maybe we need to get rid of `*_tomap` option and simply introduce pair operations: `insert`  -`insert_map`/`insert_obj`.

knazarov:
We discussed this with @mtrempoltsev and came to the conclusion that it would be impossible to have tuple format appear on the client auto-magically. There are ways to modify Tarantool core to attach format to arbitrary tuples, but it will (a) be a long discussion and (b) will not happen in Tarantool 1.10 (LTS).

For those reasons, we decided to mimic the return style of Tarantool SQL. Namely:

```yaml
---
- metadata:
  - name: id
    type: unsigned
  - name: first_name
    type: string
  - name: last_name
    type: string
  rows:
  - [1, 'Ivan', 'Ivanov']
  - [2, 'John', 'Doe']
...
```

[...]

Issue #15, opened 2020-09-24
Hide "bucket_id" field from user


olegrok:
I don't thing it's OK to show our internals for users. I was confused to see that insert returns "bucket_id".

https://github.com/tarantool/crud#insert

Actually I can do nothing with such information. For performance reasons we could return this field when work with raw tuple, but I see no reasons to preserve it in result object.

knazarov:
There is a feature in Tarantool that's called "system fields" (https://github.com/tarantool/tarantool/issues/3985) , which will help solve this. Until then I propose to forget about it.

no1seman:
Suggested solution:
1. Remove bucket_id from data and metadata by default.
- for simple operations - rigth before return. Remove will be done on router for both data and metadata.
- for select and pairs - in low level functions on storage.
2. Add new option (show_bucket_id) to all top level functions (propagate this option to low level ones if needed) to give an ability to NOT remove bucket_id form data and metadata.

PR for simple operations:
https://github.com/tarantool/crud/pull/48

no1seman:
> Hi! Thanks for your patch.
> 
> Unfortunately, currently I have more questions than answers. 
> 
> * Call table.remove for each tuple just decrease general performance of each query (especially of select). Therefore it's quite strange do enable this feature it by default. 
> 
> * In future it will be really box.tuples (after merger and key_def will be ported to Tarantool 1.10). So we will need to call tuple:truncate() for each tuple. So it will be even worse for performance.
> 
> * We won't be able to create formatted tuples because format can't be created we could only use "_format_cdata" provided by schema from connection. 
> 
> Finally, my idea was relevant when we worked with unflattened objects - the cost of actions to hide some field wasn't so high. Currently it's not so.
> 
> _Originally posted by @olegrok in https://github.com/tarantool/crud/pull/48#issuecomment-704676187_

# Task description

By default every CRUD methods:
    - Simple methods:
        - crud.delete()
        - crud.delete_object()
        - crud.get()
        - crud.get_object()
        - crud.insert()
        - crud.insert_object()
        - crud.update()
        - crud.update_object()
        - crud.upsert()
        - crud.upsert_object()
        - crud.replace()
        - crud.replace_object()
    Complex methods:
        - crud.select()
        - crud.pairs()

have to return tuple or set of tuples and metadata without bucket_id (and possibly without any other "system specific" fields). But in some cases it will be usefull to have "system specific" in data and metadata sets.
[...]

PR #39, opened 2020-10-05
Enrich CRUD API output with metadata

olegrok:
This patch changes output format of CRUD API from objects to rows
\+ metadata.

Before:
```lua
crud.insert('customers', {
    id = 1, name = 'Name', age = 1,
})
---
- bucket_id: 1
  age: 1
  name: Name
  id: 1
...
```

After:
```lua
crud.insert('customers', {
    id = 1, name = 'Name', age = 1,
})
---
- - rows:
     - [1, 1, 1, 'Name']
    metadata:
     - name: id
       type: unsigned
     - name: bucket_id
       type: unsigned
     - name: age
       type: unsigned
     - name: name
       type: string

```

It's useful when user call crud functions via connectors and (s)he
doesn't really need to deserialize tuples to map.

Need for #6

### DIFF
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -157,4 +157,20 @@ function utils.convert_operations(user_operations, space_format)
     return converted_operations
 end
 
+function utils.unflatten_rows(rows, metadata)
+    if metadata == nil then
+        return nil, UnflattenError:new('Metadata is not provided')
+    end
+
+    local result = table.new(#rows, 0)
+    local err
+    for i, row in ipairs(rows) do
+        result[i], err = utils.unflatten(row, metadata)
+        if err ~= nil then
+            return nil, err
+        end
+    end
+    return result
+end
+
 return utils
--- a/crud/select.lua
+++ b/crud/select.lua
@@ -228,9 +228,14 @@ function select_module.pairs(space_name, user_conditions, opts)
             return nil
         end
 
-        local obj, err = iter:get()
+        local tuple, err = iter:get()
         if err ~= nil then
             error(string.format("Failed to get next object: %s", err))
+        end
+
+        local obj, err = utils.unflatten(tuple, iter.space_format)
+        if err ~= nil then
+            error(string.format("Failed to unflatten next object: %s", err))
         end
 
         return iter, obj
@@ -260,7 +265,7 @@ function select_module.call(space_name, user_conditions, opts)
         return nil, err
     end
 
-    local objects = {}
+    local tuples = {}
 
     while iter:has_next() do
         local obj, err = iter:get()
@@ -272,10 +277,13 @@ function select_module.call(space_name, user_conditions, opts)
             break
         end
 
-        table.insert(objects, obj)
+        table.insert(tuples, obj)
     end
 
-    return objects
+    return {
+        metadata = table.copy(iter.space_format),
+        rows = tuples,
+    }
 end
 
 return select_module
--- a/test/integration/pairs_test.lua
+++ b/test/integration/pairs_test.lua
@@ -1,13 +1,15 @@
 local fio = require('fio')
 
 local t = require('luatest')
-local g = t.group('pairs')
+local g_memtx = t.group('pairs_memtx')
+local g_vinyl = t.group('pairs_vinyl')
+local crud = require('crud')
 
 local helpers = require('test.helper')
 
 math.randomseed(os.time())
 
-g.before_all = function()
+local function before_all(g, engine)
     g.cluster = helpers.Cluster:new({
         datadir = fio.tempdir(),
         server_command = helpers.entrypoint('srv_select'),
@@ -40,16 +42,20 @@ g.before_all = function()
                 },
             }
         },
+        env = {
+            ['ENGINE'] = engine,
+        },
     })
+    g.engine = engine
     g.cluster:start()
 end
 
-g.after_all = function()
+local function after_all(g)
     g.cluster:stop()
     fio.rmtree(g.cluster.datadir)
 end
 
-g.before_each(function()
+local function before_each(g)
     for _, server in ipairs(g.cluster.servers) do
         server.net_box:eval([[
             local space = box.space.customers
@@ -58,20 +64,37 @@ g.before_each(function()
             end
         ]])
     end
-end)
+end
 
-local function insert_customers(customers)
+g_memtx.before_all = function() before_all(g_memtx, 'memtx') end
+g_vinyl.before_all = function() before_all(g_vinyl, 'vinyl') end
+
+g_memtx.after_all = function() after_all(g_memtx) end
+g_vinyl.after_all = function() after_all(g_vinyl) end
+
+g_memtx.before_each(function() before_each(g_memtx) end)
+g_vinyl.before_each(function() before_each(g_vinyl) end)
+
+local function add(name, fn)
+    g_memtx[name] = fn
+    g_vinyl[name] = fn
+end
+
+local function insert_customers(g, customers)
     local inserted_objects = {}
 
     for _, customer in ipairs(customers) do
-        local obj, err = g.cluster.main_server.net_box:eval([[
+        local result, err = g.cluster.main_server.net_box:eval([[
             local crud = require('crud')
             return crud.insert('customers', ...)
         ]],{customer})
 
         t.assert_equals(err, nil)
 
-        table.insert(inserted_objects, obj)
+        local objects, err = crud.unflatten_rows(result.rows, result.metadata)
+        t.assert_equals(err, nil)
+        t.assert_equals(#objects, 1)
+        table.insert(inserted_objects, objects[1])
     end
 
     return inserted_objects
@@ -85,8 +108,8 @@ local function get_by_ids(customers, ids)
     return results
 end
 
-g.test_pairs_no_conditions = function()
-    local customers = insert_customers({
+add('test_pairs_no_conditions', function(g)
+    local customers = insert_customers(g, {
         {
             id = 1, name = "Elizabeth", last_name = "Jackson",
             age = 12, city = "New York",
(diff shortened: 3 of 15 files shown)


crud/select.lua
```diff
@@ -229,11 +229,16 @@ function select_module.pairs(space_name, user_conditions, opts)
             return nil
         end
 
-        local obj, err = iter:get()
+        local tuple, err = iter:get()
         if err ~= nil then
             error(string.format("Failed to get next object: %s", err))
         end
 
+        local obj, err = utils.unflatten(tuple, iter.space_format)
```
akudiyar, line 237: I suggest adding a flag to opts for forcing unflatten and return here tuples by default as in other methods. Or it will cause additional code for flattening in custom proxy methods

akudiyar (changes requested):

akudiyar:
@olegrok thank you for the patch, my praises have been heard!

akudiyar:
@olegrok @dokshina will the support for tuples as input type be added in the nearest patches?

olegrok:
> @olegrok @dokshina will the support for tuples as input type be added in the nearest patches?

I don't know. I hope we will introduce `*_tuple` methods for all DML operations.
But currently it's my own thoughts. And it should be additionally discussed.

crud/common/utils.lua
```diff
@@ -156,4 +156,20 @@ function utils.convert_operations(user_operations, space_format)
     return converted_operations
 end
 
+function utils.unflatten_rows(rows, metadata)
```
dokshina, line 159: Why not `utils.unflatten_rows({ rows = ..., metadata = ...})`? 

CRUD operations return exactly `{ rows = ..., metadata = ...}` table.

knazarov:
@olegrok I think you can trivially differentiate between a tuple and a map. Maps have at least one non-integer key.

crud/select.lua
```diff
@@ -276,7 +281,10 @@ function select_module.call(space_name, user_conditions, opts)
         table.insert(objects, obj)
     end
 
-    return objects
+    return {
+        metadata = table.copy(iter.space_format),
+        rows = objects,
```
dokshina, line 286: Objects or tuples?

test/integration/pairs_test.lua
```diff
@@ -58,19 +63,41 @@ g.before_each(function()
...
     for _, customer in ipairs(customers) do
-        local obj, err = g.cluster.main_server.net_box:eval([[
+        local result, err = g.cluster.main_server.net_box:eval([[
             local crud = require('crud')
             return crud.insert('customers', ...)
         ]],{customer})
 
         t.assert_equals(err, nil)
-
+        local obj = unflatten(result.rows[1], result.metadata)
```
dokshina, line 100: Why not `crud.unflatten_rows`?

test/integration/select_test.lua
```diff
@@ -73,17 +74,26 @@ end
...
     for _, customer in ipairs(customers) do
-        local obj, err = g.cluster.main_server.net_box:eval([[
+        local result, err = g.cluster.main_server.net_box:eval([[
             local crud = require('crud')
             return crud.insert('customers', ...)
         ]],{customer})
 
         t.assert_equals(err, nil)
 
+        local obj = unflatten(result.rows[1], result.metadata)
```
dokshina, line 96: Why not `crud.unflatten_rows`?

dokshina:
> This means you can accept both in the insert() method.

I would like to use `insert` and `insert_object` functions rather than detection it from the passed object structure. It causes less errors and allows to check user input correctly. Moreover, it's quite simple to to implement, and I don't see any reason to accept both tuple and object by one `insert` function.

knazarov:
> > This means you can accept both in the insert() method.
> 
> I would like to use `insert` and `insert_object` functions rather than detection it from the passed object structure. It causes less errors and allows to check user input correctly. Moreover, it's quite simple to to implement, and I don't see any reason to accept both tuple and object by one `insert` function.

OK, it makes sense.

crud/common/utils.lua
olegrok: I plan it will be a general purpose function.

And your suggestion is a restriction of it for specific task. If you have tuples and space:format() locally you can simply call this function without any formatting `unflatten(tuples, space:format())`.

dokshina: In this case second argument is `space_format`, not metadata?

olegrok: Well, I'll rename

olegrok: However, no... No. In future we allow to specify set of fields that should be fetched from storage.

space:format() - it's a case of metadata and currently (not in future) it's the same as metadata.

crud/select.lua
olegrok: tuples. Fixed, thanks

test/integration/pairs_test.lua
olegrok: Thanks, changed to `crud.unflatten_rows`

test/integration/select_test.lua
olegrok: Fixed

crud/select.lua
olegrok: `use_tomap` option is introduced (https://github.com/tarantool/crud/pull/39/commits/4a1b3644d5f4c39f5a5133a002860b1cda631dca)

Not sure we should implement it just now.

test/integration/select_test.lua
```diff
@@ -133,12 +138,35 @@ add('test_select_all', function(g)
...
+        {name = 'last_name', type = 'string'},
+        {name = 'age', type = 'number'},
+        {name = 'city', type = 'string'},
+    })
+
     -- no after
     local objects, err = g.cluster.main_server.net_box:eval([[
         local crud = require('crud')
 
         local objects, err = crud.select('customers', nil)
```
dokshina, line 168: Is it objects? Maybe is should be called `res` or `result`?

crud/common/utils.lua
dokshina: Yes, I agree,

crud/select.lua
dokshina: Maybe it would be better to implement this option in separated patch.

test/integration/select_test.lua
olegrok: Well, fixed

crud/select.lua
akudiyar: It costs nothing since it is already implemented, and it blocks our connector development. Since that code has been already accepted, can we port it here?

olegrok: Moved to separate PR - https://github.com/tarantool/crud/pull/43

test/integration/pairs_test.lua
```diff
@@ -58,20 +64,37 @@ g.before_each(function()
...
+g_memtx.after_all = function() after_all(g_memtx) end
+g_vinyl.after_all = function() after_all(g_vinyl) end
+
+g_memtx.before_each(function() before_each(g_memtx) end)
+g_vinyl.before_each(function() before_each(g_vinyl) end)
+
+local function add(name, fn)
+    g_memtx[name] = fn
+    g_vinyl[name] = fn
+end
```
dokshina, line 81: Can we separate this code to some helper sometimes?

olegrok: Should be moved into separate patch

test/integration/simple_operations_test.lua
```diff
@@ -241,136 +259,162 @@ add('test_update', function(g)
...
+    local result, err = g.cluster.main_server.net_box:eval([[
         local crud = require('crud')
         return crud.get('customers', 66)
     ]])
     t.assert_equals(err, nil)
-    t.assert_covers(obj, {id = 66, name = 'Leo Tolstoy', age = 50})
-    t.assert(type(obj.bucket_id) == 'number')
+    local objects = crud.unflatten_rows(result.rows, result.metadata)
+    t.assert_equals(objects, {{id = 66, name = 'Leo Tolstoy', age = 50, bucket_id = 486}})
+    t.assert(type(objects[1].bucket_id) == 'number')
```
dokshina, line 419: You've just checked that `bucket_id` us 487.

crud/common/utils.lua
```diff
@@ -157,4 +157,20 @@ function utils.convert_operations(user_operations, space_format)
...
+    local err
+    for i, row in ipairs(rows) do
+        result[i], err = utils.unflatten(row, metadata)
+        if err ~= nil then
+            return nil, err
+        end
+    end
+    return result
+end
+
```
dokshina, line 175: What do you think about hiding `bucket_id` field by default?

olegrok: https://github.com/tarantool/crud/issues/15

test/integration/simple_operations_test.lua
olegrok: Removed

test/integration/select_test.lua
```diff
@@ -798,8 +826,8 @@ add('test_select_with_collations', function(g)
 
         local conditions = ...
 
-        local objects, err = crud.select('customers', conditions)
-        return objects, err
+        local result, err = crud.select('customers', conditions)
+        return crud.unflatten_rows(result.rows, result.metadata), err
```
dokshina, line 830: nit: in `simple_operations` tests rows are unflattened outside of netbox call.

dokshina: (and I like such approach more)

dokshina:
Also, please don't forget to update CHANGELOG.

test/integration/select_test.lua
olegrok: Fixed

olegrok:
> Also, please don't forget to update CHANGELOG.

Added

test/integration/pairs_test.lua
dokshina: Agree.

dokshina (approved):
Bless you.

akudiyar (commented):
Magnificent!

PR #41, opened 2020-10-05
Reverse pagination

### LINKED ISSUE #25: [1pt] Refactor `limit` option
* Rename `limit` -> `first`
* Support negative values of `first`:
    * For `{1, 2, 3, 4, 5, 6}` select w/ options `{after = 5, first = -2}` returns `{3, 4}` (XXX: how should it work for `pairs`?)
    * Deny to use negative `first` w/o `after`

dokshina:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #25

### DIFF
--- a/crud/select/iterator.lua
+++ b/crud/select/iterator.lua
@@ -33,7 +33,6 @@ function Iterator.new(opts)
         iteration_func = opts.iteration_func,
 
         plan = opts.plan,
-        reversed = opts.plan.reversed,
 
         timeout = opts.timeout,
 
--- a/crud/select/iterator.lua
+++ b/crud/select/iterator.lua
@@ -20,7 +20,6 @@ function Iterator.new(opts)
         iteration_func = 'function',
 
         plan = 'table',
-        after_tuple = '?table',
 
         batch_size = 'number',
         replicasets = 'table',
@@ -34,7 +33,8 @@ function Iterator.new(opts)
         iteration_func = opts.iteration_func,
 
         plan = opts.plan,
-        after_tuple = opts.after_tuple,
+        reversed = opts.plan.reversed,
+
         timeout = opts.timeout,
 
         replicasets = table.copy(opts.replicasets),
@@ -56,7 +56,7 @@ function Iterator.new(opts)
 
     setmetatable(iter, Iterator)
 
-    iter:_update_replicasets_tuples(iter.after_tuple)
+    iter:_update_replicasets_tuples(iter.plan.after_tuple)
 
     return iter
 end
--- a/crud/select/plan.lua
+++ b/crud/select/plan.lua
@@ -1,6 +1,7 @@
 local errors = require('errors')
 
 local select_conditions = require('crud.select.conditions')
+local utils = require('crud.common.utils')
 local dev_checks = require('crud.common.dev_checks')
 
 local select_plan = {}
@@ -87,7 +88,9 @@ end
 function select_plan.new(space, conditions, opts)
     dev_checks('table', '?table', {
         first = '?number',
+        after_tuple = '?table',
     })
+
     conditions = conditions ~= nil and conditions or {}
     opts = opts or {}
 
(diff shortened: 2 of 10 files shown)


crud/select/iterator.lua
```diff
@@ -34,7 +33,8 @@ function Iterator.new(opts)
         iteration_func = opts.iteration_func,
 
         plan = opts.plan,
-        after_tuple = opts.after_tuple,
+        reversed = opts.plan.reversed,
```
olegrok, line 36: I could consider this change only as attempt to make me resolve merge conflicts at [merger integration](https://github.com/tarantool/crud/pull/38) branch and finally put "reversed" this logic back to select.

I suggest you remove this change and just check `opts.first < 0` just at select. Because it's actually unused in all places except select.

crud/select/plan.lua
```diff
@@ -135,8 +138,29 @@ function select_plan.new(space, conditions, opts)
...
+
+            -- scan condition becomes border consition
+            scan_condition_num = nil
+
+            if scan_after_tuple ~= nil then
+                scan_value = utils.extract_key(scan_after_tuple, scan_index.parts)
+            else
+                scan_value = nil
+            end
+
```
olegrok, line 161: redundant line

test/integration/select_test.lua
```diff
@@ -219,6 +219,126 @@ add('test_select_all_with_limit', function(g)
...
+            batch_size = batch_size,
+        })
+        return objects, err
+    ]], {first, after, batch_size})
+
+    t.assert_equals(err, nil)
+    t.assert_equals(objects, get_by_ids(customers, {2, 3, 4}))
+
+    -- age <= 33 (eq for id: 3)
+    -- (in sorted by age order: 4, 2, 8, 5, 3, 6, 7, 1)
```
olegrok, line 291: I hope you perfectly understand that this test shouldn't test that you query planner works fine.
And such comments and conditions just overcomplicate test logic.

I suggest you to use "id > 100" with reverse pagination - it allows to preserve clear logic (all your elements will be placed in order that specified at start of this test, by id). Some person more easily understand this logic and you don't make resort the data in mind or verify that specified result here is correct.

Also you need to check several basic cases:
  - you get inverted result set
  - you invert them at router not storage

To test first case you just need to query all data. To test the second case you need to query all records one by one.

crud/select/iterator.lua
dokshina: OK, I was thinking about checking `opts.first < 0` in select, geniuses' thoughts converge.

test/integration/select_test.lua
dokshina: I've added this check for `age` condition to be sure that scan value is set properly for this case.

test/integration/pairs_test.lua
```diff
@@ -296,3 +296,37 @@ add('test_le_condition_with_index', function(g)
     t.assert_equals(err, nil)
     t.assert_equals(objects, get_by_ids(customers, {1})) -- in age order
 end)
+
+add('test_negative_forst', function(g)
```
olegrok, line 300: typo: `forst` -> `first`

test/integration/pairs_test.lua
```diff
@@ -296,3 +296,37 @@ add('test_le_condition_with_index', function(g)
...
+    -- negative first
+    local first = -10
+    t.assert_error_msg_contains("Negative first isn't allowed for pairs", function()
+        g.cluster.main_server.net_box:eval([[
+            local crud = require('crud')
+
+            local first = ...
+
+            local objects = {}
+            for _, object in crud.pairs('customers', nil, {first = first}) do
```
olegrok, line 325: could be replaced with simple one-liner. You just need to call `crud.pairs('customers', nil, {first = ...})` - without loop body

test/integration/select_test.lua
olegrok: Well, test looks good now. Thanks.

olegrok (approved):

PR #40, opened 2020-10-05
introduce internal API for raw tuples

olegrok:
This patch introduces internal API for insert, replace and upsert
raw tuples. To reduce amonut of changes this methods will be
exposed in next patches.
    
Also some kind of optimization is introduced. Before this patch we
call unflatten for object twice. This patch removes one of them.
This way reduce amount of Lua garbage.
    
Part of #6

PR #42, opened 2020-10-06
Expand DML API with tuple-based functions

### LINKED ISSUE #6: Support tuples as result type

olegrok:
This patch completes series of patches that allow to use raw
    tuples as input and output arguments.
    It was decided to use insert/replace/delete functions only for raw
    tuples. In contrast {insert/replace/delete}_object functions
    allow to put unflattened objects as it was before.
    
Closes #6


crud.lua
```diff
@@ -25,25 +25,37 @@ crud.register = registry.add
...
 
--- @refer insert.call
+-- @refer insert.call_raw
 -- @function insert
-crud.insert = insert.call
+crud.insert = insert.call_raw
+
+-- @refer insert.call
+-- @function insert_object
+crud.insert_object = insert.call
```
dokshina, line 34: It looks a bit strange:

```lua
crud.insert = insert.call_raw
crud.insert_object = insert.call
```

It seems that default flow for `insert` module is "objects" flow, but `crud` uses "raw" flow by default.

```lua
crud.insert = insert.raw
crud.insert_object = insert.obj
```

So, `crud` module decides which flow is default, while `insert` module just introduces two different functions.

crud/upsert.lua
```diff
@@ -18,7 +18,7 @@ local function call_upsert_on_storage(space_name, tuple, operations)
 
     local space = box.space[space_name]
     if space == nil then
-        return nil, UpsertError:new("Space %q doesn't exists", space_name)
+        return nil, UpsertError:new("Space %q doesn't exist", space_name)
```
dokshina, line 21: I would like this changes to be performed in a separated patch.

test/integration/simple_operations_test.lua
```diff
@@ -279,7 +342,7 @@ add('test_delete', function(g)
     -- insert tuple
     local result, err = g.cluster.main_server.net_box:eval([[
         local crud = require('crud')
-        return crud.insert('customers', {id = 33, name = 'Mayakovsky', age = 36})
+        return crud.insert_object('customers', {id = 33, name = 'Mayakovsky', age = 36})
```
dokshina, line 345: I see comment `--insert tuple`, but then you call `insert_object`...

Maybe `tuple/object` functions should be tested in a separated tests? `test_delete` and `test_delete_object`?

crud.lua
olegrok: I've change a bit naming convention here - https://github.com/tarantool/crud/pull/40

insert.tuple - for tuples
insert.object - for objects

test/integration/simple_operations_test.lua
olegrok: Sounds reasonable

olegrok: Done

crud/upsert.lua
olegrok: Sure, let's push https://github.com/tarantool/crud/pull/45 firstly

crud.lua
olegrok: Naming convention was changed in previous patch. This patch only changes API and test them.

crud/replace.lua
```diff
@@ -120,7 +120,7 @@ function replace.object(space_name, obj, opts)
 
     local space = utils.get_space(space_name, vshard.router.routeall())
     if space == nil then
-        return nil, ReplaceError:new("Space %q doesn't exists", space_name)
+        return nil, ReplaceError:new("Space %q doesn't exist", space_name)
```
dokshina, line 123: Again?

crud/upsert.lua
```diff
@@ -134,7 +134,7 @@ function upsert.object(space_name, obj, user_operations, opts)
 
     local space = utils.get_space(space_name, vshard.router.routeall())
     if space == nil then
-        return nil, UpsertError:new("Space %q doesn't exists", space_name)
+        return nil, UpsertError:new("Space %q doesn't exist", space_name)
```
dokshina, line 137: Again again?

test/integration/simple_operations_test.lua
```diff
@@ -82,66 +82,94 @@ end
...
     -- delete
-    local obj, err = g.cluster.main_server.net_box:eval([[
+    local result, err = g.cluster.main_server.net_box:eval([[
         local crud = require('crud')
         return crud.delete('non_existent_space', 0)
     ]])
 
-    t.assert_equals(obj, nil)
+    t.assert_equals(result, nil)
     t.assert_str_contains(err.err, 'Space "non_existent_space" doesn\'t exist')
```
dokshina, line 130: get, update, delete object?

dokshina (approved):

crud/replace.lua
olegrok: Bad rebase, sorry :(

PR #43, opened 2020-10-06
introduce "use_tomap" option for pairs

olegrok:
This patch introduces "use_tomap" (`false` by default) option to
    convert raw selected tuples to objects.
    
 Usage:
```lua
-- returns raw tuples
crud.pairs('customers', {{'=', 'id', 1}}, {use_tomap = false})
    
-- returns objects
crud.pairs('customers', {{'=', 'id', 1}}, {use_tomap = true})
```

Part of #6

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,12 +304,17 @@ crud.select('customers', {{'<=', 'age', 35}})
 You can iterate across a distributed space using the `crud.pairs` function.
 Its arguments are the same as [`crud.select`](#select) arguments,
 but negative `first` values aren't allowed.
+User could pass use_tomap flag (false by default) to iterate over flat tuples or objects.
 
 **Example:**
 
 ```lua
-for _, obj in crud.pairs('customers', {{'<=', 'age', 35}}) do
-    -- do smth with the object
+for _, tuple in crud.pairs('customers', {{'<=', 'age', 35}}, {use_tomap = false}) do
+    print(json.encode(tuple)) -- {5, 1172, 'Jack', 35}
+end
+
+for _, object in crud.pairs('customers', {{'<=', 'age', 35}}, {use_tomap = true}) do
+    print(json.encode(object)) -- {id = 5, name = 'Jack', bucket_id = 1172, age = 35}
 end
 ```
 
--- a/crud/select.lua
+++ b/crud/select.lua
@@ -204,6 +204,7 @@ function select_module.pairs(space_name, user_conditions, opts)
         first = '?number',
         timeout = '?number',
         batch_size = '?number',
+        use_tomap = '?boolean',
     })
 
     opts = opts or {}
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -304,11 +304,14 @@ crud.select('customers', {{'<=', 'age', 35}})
 You can iterate across a distributed space using the `crud.pairs` function.
 Its arguments are the same as [`crud.select`](#select) arguments,
 but negative `first` values aren't allowed.
+Also since it's possible to return only rows without metadata
```
dokshina, line 307: > only rows

I think that rows is a tuples, maybe it would be better to say just "User could pass `use_tomap` flag (`false` by default) to iterate over flat tuples or objects."

README.md
```diff
@@ -304,11 +304,14 @@ crud.select('customers', {{'<=', 'age', 35}})
...
+user could pass `use_tomap` flag (`false` by default) to
+iterate over flat tuples or objects.
 
 **Example:**
 
 ```lua
-for _, obj in crud.pairs('customers', {{'<=', 'age', 35}}) do
+for _, obj in crud.pairs('customers', {{'<=', 'age', 35}}, {use_tomap = true}) do
     -- do smth with the object
 end
```
dokshina, line 316: What about adding an example w/ `use_tomap = false`.

BTW, this example doesn't show what exactly is returned.

crud/select.lua
```diff
@@ -233,9 +234,12 @@ function select_module.pairs(space_name, user_conditions, opts)
             error(string.format("Failed to get next object: %s", err))
         end
 
-        local obj, err = utils.unflatten(tuple, iter.space_format)
-        if err ~= nil then
-            error(string.format("Failed to unflatten next object: %s", err))
+        local obj = tuple
```
dokshina, line 237: Maybe it's `result`, not object? Or valiue.

`obj = tuple` confuses me, this feature should separate objects and tuples cases.

README.md
olegrok: Ok, just `User could pass use_tomap flag (false by default) to iterate over flat tuples or objects.`

README.md
olegrok: Rewritten to

```lua
for _, obj in crud.pairs('customers', {{'<=', 'age', 35}}, {use_tomap = false}) do
    print(json.encode(obj)) -- {5, 1172, 'Jack', 35}
end

for _, obj in crud.pairs('customers', {{'<=', 'age', 35}}, {use_tomap = true}) do
    print(json.encode(obj)) -- {id = 5, name = 'Jack', bucket_id = 1172, age = 35}
end
```

crud/select.lua
olegrok: Ok, fixed

README.md
dokshina: Is it `obj` in both cases?

olegrok: Well, let's use `tuple` in the first case and `object` in the second.

dokshina (approved):
Blessed!

PR #51, opened 2020-10-07
Select: after option is tuple

dokshina:
Since all operations return rows, it seems logical to use row (a.k.a. tuple) as `after` value for `select` and `pairs`.

- [x] Tests
- [x] Changelog
- [x] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ where:
 * `opts`:
   * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
 
-Returns metadata and array contains one deleted row, error.
+Returns metadata and array contains one deleted row (empty for vinyl), error.
 
 **Example:**
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ where:
 * `opts`:
   * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
 
-Returns inserted rows and metadata or nil with error.
+Returns metadata and array contains one inserted row, error.
 
 **Example:**
 
@@ -76,7 +76,7 @@ where:
 * `opts`:
   * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
 
-Returns object, error.
+Returns metadata and array contains one row, error.
 
 **Example:**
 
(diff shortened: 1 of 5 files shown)


test/integration/pairs_test.lua
```diff
@@ -161,7 +165,7 @@ add('test_pairs_no_conditions', function(g)
     t.assert_equals(objects, get_by_ids(customers, {3, 4}))
 
     -- after obj 4 (last)
-    local after = customers[4]
+    local after = crud_utils.flatten(customers[4], g.space_format)
```
olegrok, line 168: Seems we need to merge https://github.com/tarantool/crud/pull/43 before

test/integration/select_test.lua
```diff
@@ -174,7 +178,7 @@ add('test_select_all', function(g)
     t.assert_equals(objects, customers)
 
     -- after obj 2
-    local after = customers[2]
+    local after = crud_utils.flatten(customers[2], g.space_format)
```
olegrok, line 181: `local after = result.rows[2]`

dokshina: I don't like it - in this case we depends on previous result.

test/integration/pairs_test.lua
dokshina: Agree.

dokshina: Merged ;)

test/integration/select_test.lua
olegrok: Well, ok...

But I see nothing bad

README.md
```diff
@@ -137,7 +137,7 @@ where:
 * `opts`:
   * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
 
-Returns deleted object, error.
+Returns metadata and array contains one deleted row, error.
```
olegrok, line 140: It's not so in case of vinyl

olegrok (approved):

README.md
dokshina: Fixed

Issue #26, opened 2020-09-30
Add router role with automatic proxying of the crud API via public functions


akudiyar:
Suppose we have a sharded space with customer profiles and want to access them by IPROTO (for example, from a connector). In this case, we have to write a lot of boilerplate code like that:

```lua
...
local function profile_storage_select(conditions)
    return crud.select('profiles', conditions)
end

local function profile_storage_get(id)
    return crud.get('profiles', id)
end

local function profile_storage_insert(tuple)
    return crud.insert('profiles', tuple)
end
...
local function init(opts)
    if opts.is_master then
        box.schema.func.create('profile_storage_insert', {if_not_exists = true})
        box.schema.func.create('profile_storage_get', {if_not_exists = true})
        box.schema.func.create('profile_storage_select', {if_not_exists = true})
        ....
    end

    rawset(_G, 'profile_storage_insert', profile_storage_insert)
    rawset(_G, 'profile_storage_get', profile_storage_get)
    rawset(_G, 'profile_storage_select', profile_storage_select)
    ...

    return true
end

return {
    role_name = 'app.roles.api_storage',
    init = init,
    utils = {
        profile_storage_get = profile_storage_get,
        profile_storage_insert = profile_storage_insert,
        profile_storage_select = profile_storage_select,
        ...
    },
    dependencies = {
        ....
    }
}
```

And we need to do this for each space we want to access using CRUD.

I suggest adding a small module, which encapsulates all this boilerplate logic so that all we need to write is something like

```lua
local proxy = require('proxy').new()

local function init(opts)
    if opts.is_master then
        proxy.register('profiles', 'customers')
    ...
    end
    ...
    return true
end

return {
    role_name = 'app.roles.api_storage',
    init = init,
    utils = table.merge(proxy.api(), {
        ...
    }),
    dependencies = {
        ....
    }
}
```

This module can be automatically enabled in a special "proxy" role when ddl is installed.

knazarov:
Having to do all the boilerplate you list is a problem. But I see a better solution from what you propose.

I think client libraries should call "crud.*" directly, without any wrappers. For this to work, the crud module should register itself in a global variable with the same name as the module name when first loaded. Then it will be possible to call "crud.*" directly.

akudiyar:
@knazarov I will accept either of these solutions, but exposing the whole module seems to be less safe than a list of proxy functions.

knazarov:
[...]

PR #58, opened 2020-10-13
Expose the crud module in global scope

knazarov:
This is required for clients to be able to call the module via
iproto. In scope of #26.


crud.lua
```diff
@@ -91,4 +85,13 @@ function crud.init()
     select.init()
 end
 
+function crud.init_router()
+   rawset(_G, 'crud', crud)
+end
+
+function crud.init()
```
dokshina, line 92: I don't like this function. Nobody takes care about `init_router` and `init_storage` when `init` function exists.

crud/common/call.lua
```diff
@@ -165,4 +135,33 @@ function call.ro(func_name, func_args, opts)
...
+
+    -- This is a workaround, until vshard supports telling us where the error happened
+    if err ~= nil then
+        if type(err) == 'table' and err.type == 'ClientError' and type(err.message) == 'string' then
+            if err.message == string.format("Procedure '%s' is not defined", function_name) then
+               err = NotInitializedError:new("crud isn't initialized on replicaset")
+            end
+        end
+
+
```
dokshina, line 149: Redundant empty line.

crud/common/call.lua
```diff
@@ -165,4 +135,33 @@ function call.ro(func_name, func_args, opts)
     return call_impl('callro', func_name, func_args, opts)
 end
 
+function call.rw_single(bucket_id, function_name, argument_list, options)
```
dokshina, line 138: This module uses `func_name` and `func_args`, please make all names consistent.

crud/common/call.lua
```diff
@@ -165,4 +135,33 @@ function call.ro(func_name, func_args, opts)
...
+    -- This is a workaround, until vshard supports telling us where the error happened
+    if err ~= nil then
+        if type(err) == 'table' and err.type == 'ClientError' and type(err.message) == 'string' then
+            if err.message == string.format("Procedure '%s' is not defined", function_name) then
+               err = NotInitializedError:new("crud isn't initialized on replicaset")
+            end
+        end
+
+
+        local replicaset, err2 = vshard.router.route(bucket_id)
```
dokshina, line 150: What is `err2`? Maybe it's `route_err`?

crud/common/call.lua
```diff
@@ -165,4 +135,33 @@ function call.ro(func_name, func_args, opts)
...
+        if replicaset == nil then
+            return nil, CallError:new("Failed to get replicaset for bucket_id %s: %s", bucket_id, err2)
```
dokshina, line 152: Maybe in this case we should return `err` anyway and additionally say that `route` failed and it's not possible to get replicaset UUID?

crud/delete.lua
```diff
@@ -66,24 +63,17 @@ function delete.call(space_name, key, opts)
...
 
     if err ~= nil then
         return nil, DeleteError:new("Failed to delete: %s", err)
     end
 
-    local tuple = results[replicaset.uuid]
     return {
         metadata = table.copy(space:format()),
-        rows = {tuple},
+        rows = {results},
```
dokshina, line 76: `result`? `{results}` looks like an array of arrays.

crud/get.lua
```diff
@@ -65,26 +62,25 @@ function get.call(space_name, key, opts)
...
-    }
+    if results == nil then
+        return {
+            metadata = table.copy(space:format()),
+            rows = {},
+        }
+    else
+        return {
+           metadata = table.copy(space:format()),
+           rows = {results},
```
dokshina, line 81: `{result}`?

crud/get.lua
```diff
@@ -65,26 +62,25 @@ function get.call(space_name, key, opts)
...
-    local replicaset, err = vshard.router.route(bucket_id)
-    if replicaset == nil then
-        return nil, GetError:new("Failed to get replicaset for bucket_id %s: %s", bucket_id, err.err)
-    end
-
-    local results, err = call.rw(GET_FUNC_NAME, {space_name, key}, {
-        replicasets = {replicaset},
-        timeout = opts.timeout,
-    })
+    local results, err = call.rw_single(
```
dokshina, line 65: Looks like `ro_single` (and `call.ro)`.

crud/insert.lua
```diff
@@ -78,19 +70,17 @@ function insert.tuple(space_name, tuple, opts)
...
 
     if err ~= nil then
         return nil, InsertError:new("Failed to insert: %s", err)
     end
 
-    local tuple = results[replicaset.uuid]
     return {
         metadata = table.copy(space_format),
-        rows = {tuple},
+        rows = {results},
```
dokshina, line 83: `{result}`

crud/replace.lua
```diff
@@ -78,19 +70,18 @@ function replace.tuple(space_name, tuple, opts)
...
     if err ~= nil then
         return nil, ReplaceError:new("Failed to replace: %s", err)
     end
 
-    local tuple = results[replicaset.uuid]
-    return {
+     return {
         metadata = table.copy(space:format()),
-        rows = {tuple},
+        rows = {results},
```
dokshina, line 84: `{result}`

crud/update.lua
```diff
@@ -75,24 +72,17 @@ function update.call(space_name, key, user_operations, opts)
...
 
     if err ~= nil then
         return nil, UpdateError:new("Failed to update: %s", err)
     end
 
-    local tuple = results[replicaset.uuid]
     return {
         metadata = table.copy(space_format),
-        rows = {tuple},
+        rows = {results},
```
dokshina, line 85: `{result}`

crud/delete.lua
```diff
@@ -66,24 +63,17 @@ function delete.call(space_name, key, opts)
...
-        return nil, DeleteError:new("Failed to get replicaset for bucket_id %s: %s", bucket_id, err.err)
-    end
-
-    local results, err = call.rw(DELETE_FUNC_NAME, {space_name, key}, {
-        replicasets = {replicaset},
-        timeout = opts.timeout,
-    })
+    local results, err = call.rw_single(
+        bucket_id, DELETE_FUNC_NAME,
+        {space_name, key}, {timeout=opts.timeout})
```
dokshina, line 68: `{timeout=opts.timeout}`

Please, use `{key = value}`, see [our Lua style guide](https://www.tarantool.io/en/doc/latest/dev_guide/lua_style_guide/), section `Using spaces`.

crud/upsert.lua
```diff
@@ -88,10 +80,9 @@ function upsert.tuple(space_name, tuple, user_operations, opts)
     end
 
     tuple[bucket_id_fieldno] = bucket_id
-    local _, err = call.rw(UPSERT_FUNC_NAME, {space_name, tuple, operations}, {
-        replicasets = {replicaset},
-        timeout = opts.timeout,
-    })
+    local _, err = call.rw_single(
+        bucket_id, UPSERT_FUNC_NAME,
+        {space_name, tuple, operations}, {timeout=opts.timeout})
```
dokshina, line 85: JFYI, I like the old style more

```lua
local _, err = call.rw_single( bucket_id, UPSERT_FUNC_NAME, {space_name, tuple, operations}, {
    timeout = opts.timeout,
})
```

test/entrypoint/srv_select.lua
```diff
@@ -51,13 +51,22 @@ package.preload['customers-storage'] = function()
     }
 end
 
+package.preload['customers-router'] = function()
+    return {
+        role_name = 'customers-router',
+        init = function()
+        end,
```
dokshina, line 58: nit: `init = function() end`

crud/get.lua
```diff
@@ -65,26 +62,25 @@ function get.call(space_name, key, opts)
...
+        return {
+            metadata = table.copy(space:format()),
+            rows = {},
+        }
+    else
+        return {
+           metadata = table.copy(space:format()),
+           rows = {results},
+        }
+    end
```
dokshina, line 83: What about

```lua
if results == nil then 
    results = nil
end

return {
           metadata = table.copy(space:format()),
           rows = {results},
}
```

?

dokshina:
Thanks a lot for your patch!

I'm incredibly happy looking on new tests using `call` instead of `eval`.

But I can be happy a little more. I would like this patch to be separated into three patches - `get rid of the registry` can be merged separately, as I can see. The same for introducing `call_{rw,ro}_single` - it's not about exposing crud module.

Or you can simply re-organize commits and merge this PR without squashing.

cartridge/roles/crud-storage.lua
```diff
@@ -2,7 +2,7 @@ local crud = require('crud')
 
 -- removes routes that changed in config and adds new routes
 local function init()
-    crud.init()
```
dokshina, line 5: The comment above looks like a (my) copy-paste, can you remove it?

crud/get.lua
knazarov: It was `call.rw()` originally, so I didn't change it. If you think about it, there is some danger while executing this on the replica. The changes may not have propagated after you change them on the master.

crud/get.lua
knazarov: This is because the `result` can be `box.NULL`, in this case, it will make it into the rows table. I will make a comment about that.

knazarov:
At the moment, the patch does more than just expose the functions. But since I had to remove the registry to make the expose behavior consistent between the router and the storage, I had to also rewrite portions of the `call.lua`, which in turn led to simplifying it.

The only unrelated change is `call.rw_single`, which is rather small.

crud/get.lua
dokshina: OK, but please add a comment with this problem description.

.gitignore
```diff
@@ -3,3 +3,6 @@ build.luarocks/
 myapp/
 **.snap
 **.xlog
+.tarantool.cookie
+CMakeCache.txt
+CMakeFiles/
```
dokshina, line 8: No newline

dokshina:
> At the moment, the patch does more than just expose the functions.

The PR name and description tells me other.

crud/get.lua
dokshina: Your `if` branch returns, why do you use `else`?

This code handles `box.NULL` case and makes code much more clear:

```
if results == nil then 
    results = nil
end
```

crud/common/call.lua
```diff
@@ -165,4 +135,40 @@ function call.ro(func_name, func_args, opts)
...
+
+        local replicaset, _ = vshard.router.route(bucket_id)
+        if replicaset == nil then
+            return nil, CallError:new("Function returned an error, but we couldn't figure out the replicaset: %s",
+                                      err)
+        end
+
+        return nil, CallError:new(utils.format_replicaset_error(
+                                     replicaset.uuid, "Function returned an error: %s", err
+                                 ))
```
dokshina, line 164: OK, here use use `)` on a new line, but 

```lua
return nil, CallError:new("Function returned an error, but we couldn't figure out the replicaset: %s",
                                      err)
```

and

```lua
local result, err = call.rw_single(
        bucket_id, DELETE_FUNC_NAME,
        {space_name, key}, {timeout = opts.timeout})
```

Please, use approach that is used in `crud` everywhere:

```lua
return nil, CallError:new(
    "Function returned an error, but we couldn't figure out the replicaset: %s", err
)
```

dokshina:
Are you planning to fix PR title and description? We really want to merge it into master.

dokshina (approved):

dokshina (approved):

PR #63, opened 2020-10-15
Fix getting primary space index

### LINKED ISSUE #62: Select by space primary index fail
Call select with conditions by primary index returns error:

`SelectError: Failed to plan select: SelectPlanError: Passed bad conditions: ValidateConditionsError: No field or index "primary" found`

savolgin:
Closes #62

### DIFF
--- a/test/entrypoint/srv_select.lua
+++ b/test/entrypoint/srv_select.lua
@@ -24,7 +24,7 @@ package.preload['customers-storage'] = function()
                 if_not_exists = true,
                 engine = engine,
             })
-            customers_space:create_index('id', {
+            customers_space:create_index('id_index', {
                 parts = { {field = 'id'} },
                 if_not_exists = true,
             })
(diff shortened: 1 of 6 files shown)


test/entrypoint/srv_select.lua
```diff
@@ -24,7 +24,7 @@ package.preload['customers-storage'] = function()
                 if_not_exists = true,
                 engine = engine,
             })
-            customers_space:create_index('id', {
+            customers_space:create_index('id_index', {
```
olegrok, line 27: But seems we still need to test a case when field and index have the same name.

savolgin: Added indexes with same names as fields.

dokshina: Maybe it would be better to add unit tests for `get_index_by_name`? To make select tests more e2e?

savolgin: Test added

olegrok (approved):

dokshina:
Thanks for your patch!

Please, update CHANGELOG before merging.

dokshina (approved):

dokshina:
Thanks for your patch!

Please, update CHANGELOG before merging.

savolgin:
Change log updated

PR #65, opened 2020-10-16
Allow to specify custom bucket ID

### LINKED ISSUE #46: [4pt] Support sharding by custom field/index

dokshina:
* Bucket ID can be specified using `opts.bucket_id` for all operations
* For operations that accepts tuple/object bucket ID can be specified as
   tuple/object field as well as `opts.bucket_id` value

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #46

### DIFF
--- a/crud/common/sharding.lua
+++ b/crud/common/sharding.lua
@@ -1,0 +1,54 @@
+local vshard = require('vshard')
+local errors = require('errors')
+
+local BucketIDError = errors.new_class("BucketIDError", {capture_stack = false})
+
+local utils = require('crud.common.utils')
+
+local sharding = {}
+
+function sharding.key_get_bucket_id(key, specified_bucket_id)
+    if specified_bucket_id ~= nil then
+        return specified_bucket_id
+    end
+
+    return vshard.router.bucket_id_strcrc32(key)
+end
+
+function sharding.tuple_get_bucket_id(tuple, space, specified_bucket_id)
+    if specified_bucket_id ~= nil then
+        return specified_bucket_id
+    end
+
+    local key = utils.extract_key(tuple, space.index[0].parts)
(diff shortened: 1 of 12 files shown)


crud/get.lua
```diff
@@ -51,6 +52,7 @@ end
 function get.call(space_name, key, opts)
     checks('string', '?', {
         timeout = '?number',
+        bucket_id = '?number',
```
olegrok, line 55: `cdata` numbers are possible as well

test/integration/custom_bucket_id_test.lua
```diff
@@ -0,0 +1,541 @@
...
+
+local t = require('luatest')
+local g_memtx = t.group('custom_bucket_id_memtx')
+local g_vinyl = t.group('custom_bucket_id_vinyl')
+
+local crud = require('crud')
+
+local helpers = require('test.helper')
+
+math.randomseed(os.time())
```
olegrok, line 11: How do you use it?

dokshina: Removed

crud/common/sharding.lua
```diff
@@ -0,0 +1,54 @@
...
+                )
+            end
+        end
+    end
+
+    if tuple[bucket_id_fieldno] == nil then
+        tuple[bucket_id_fieldno] = sharding.get_bucket_id_by_tuple(tuple, space)
+    end
+
+    local bucket_id = tuple[bucket_id_fieldno]
```
olegrok, line 50: nit: May be `set` should return just `nil` or `true` but not bucket_id?

dokshina: Renamed to `tuple_set_and_return_bucket_id`.

olegrok (approved):

PR #66, opened 2020-10-20
ci: fix release push

olegrok:
Previousely we use wrong syntax to detect the case when tag was
pushed. This patch fixes it in following way:
  - Use startsWith(github.ref, 'refs/tags') for tag detection
  - Use ${GITHUB_REF#refs/*/} to extract tag from ref

### DIFF
--- a/.github/workflows/test_on_push.yaml
+++ b/.github/workflows/test_on_push.yaml
@@ -52,6 +52,9 @@ jobs:
     runs-on: [ ubuntu-latest ]
     steps:
       - uses: actions/checkout@master
+      # https://stackoverflow.com/questions/58177786/get-the-current-pushed-tag-in-github-actions
+      - name: Set env
+        run: echo "GIT_TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
       - name: Push scm rockspec
         if: github.event_name == 'push' && github.ref == 'refs/heads/master'
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test_on_push.yaml
```diff
@@ -58,13 +58,13 @@ jobs:
...
           curl --fail -X PUT -F rockspec=@$ROCK_NAME-scm-1.rockspec \
             https://${{ secrets.ROCKS_USERNAME }}:${{ secrets.ROCKS_PASSWORD }}@${{ secrets.ROCKS_SERVER }} \
           cat $ROCK_NAME-scm-1.rockspec |
               sed -E \
-                -e "s/branch = '.+'/tag = '$GITHUB_REF'/g" \
-                -e "s/version = '.+'/version = '$GITHUB_REF-1'/g" |
-              curl --fail -X PUT -F "rockspec=@-;filename=$ROCK_NAME-$GITHUB_REF-1.rockspec" \
+                -e "s/branch = '.+'/tag = '${GITHUB_REF#refs/*/}'/g" \
+                -e "s/version = '.+'/version = '${GITHUB_REF#refs/*/}-1'/g" |
+              curl --fail -X PUT -F "rockspec=@-;filename=$ROCK_NAME-${GITHUB_REF#refs/*/}-1.rockspec" \
```
dokshina, line 69: Is it possible to move `${GITHUB_REF#refs/*/}` into variable?

.github/workflows/test_on_push.yaml
```diff
@@ -58,13 +58,13 @@ jobs:
           curl --fail -X PUT -F rockspec=@$ROCK_NAME-scm-1.rockspec \
             https://${{ secrets.ROCKS_USERNAME }}:${{ secrets.ROCKS_PASSWORD }}@${{ secrets.ROCKS_SERVER }}
       - name: Push release rockspec
-        if: github.event_name == 'push' && github.ref == 'refs/tags/'
```
dokshina, line 61: Why did you removed `event_name` check?

olegrok: `startsWith(github.ref, 'refs/tags')` is enough. Also I don't know alternative ways to create a tag

dokshina (approved):

PR #67, opened 2020-10-22
Refactor GitHub Actions workflows

dokshina:
* Separate push-rockspec flow
* Fix push-tagged-rockspec job
* Run tests for PR or push, not both


olegrok (approved):
Thanks for your patch. And bless you after `github.event.pull_request.head.repo.owner.login != 'tarantool'`.

PR #81, opened 2020-10-30
Fixed typo in error format

AndreevSemen:
Fixed typo in error format in `tuple_set_and_return_bucket_id()` (`%s` was missed)


dokshina (approved):
Thanks for your patch!

PR #86, opened 2020-11-11
Add support for UUID keys comparison

### LINKED ISSUE #84: Support UUIDs as keys
Given this space:

```lua
local customers = box.schema.space.create(
        'customers',
        {
            format = {
                {name = 'uuid', type = 'uuid'},
                {name = 'bucket_id', type = 'unsigned'},
                {name = 'name', type = 'string'},
                {name = 'tags', type = 'array'},
            },
            if_not_exists = true,
        }
    )

    customers:create_index('uuid', {
        parts = {'uuid'},
        if_not_exists = true,
    })

    customers:create_index('bucket_id', {
        parts = {'bucket_id'},
        unique = false,
        if_not_exists = true,
    })
```

It is impossible to perform `crud.select('customers')`, since it fails with an error like 

```
E> LuajitError: /app/.rocks/share/tarantool/crud/select/comparators.lua:53: attempt to compare 'struct tt_uuid' with 'struct tt_uuid'
```

akudiyar:
Add support for indexes with UUID key parts. UUID comparison is done via the uuid:str() method and may be further replaced with a more efficient comparison using `tt_uuid_compare` (see https://github.com/tarantool/tarantool/issues/5511)

I didn't forget about

- [x] Tests
- [x] Changelog
- [ ] Documentation

Closes #84


crud/select/comparators.lua
```diff
@@ -127,24 +128,27 @@ local function array_ge(lhs, rhs, len, lt_funcs, eq_funcs)
...
+    if collations.is_default(collation) then
+        return types.lt(key_part) or lt, types.eq(key_part) or eq
+    elseif collation == collations.UNICODE then
+        return lt_unicode, eq_unicode
+    elseif collation == collations.UNICODE_CI then
+        return lt_unicode_ci, eq_unicode_ci
+    else
+        return nil, GenFuncError:new('Unsupported Tarantool collation %q', collation)
+    end
+end
```
dokshina, line 142: I suggest adding `eq_uuid` and `lt_uuid` functions and check that `key_part.type` is `uuid` before checking collation.
It's much more simple and clear solution.

Moreover, I don't like `cmp_funcs_by_key_type` name (in fact, it takes not `key_type` arg but `key_part`). I suggest `get_cmp_func_by_key_part`.

test/unit/select_comparators_test.lua
```diff
@@ -139,3 +141,24 @@ g.test_unicode_collations = function()
...
 end
+
+g.test_uuid_type_parts = function()
+    local key_parts = {
+        {type = 'uuid'}, {}
+    }
+
+    local func_eq, err = select_comparators.gen_func(operators.EQ, key_parts)
+    t.assert(err == nil)
+    local u1, u2 = uuid.fromstr('8a1ffbc0-241e-11eb-8d27-0800200c9a66'), uuid.fromstr('8f1ffbc0-241e-11eb-8d27-0800200c9a66')
```
dokshina, line 152: Why don't you split it into 2 lines?

test/unit/select_comparators_test.lua
```diff
@@ -139,3 +141,24 @@ g.test_unicode_collations = function()
...
+    local u1_copy = uuid.fromstr(u1:str())
+    t.assert(func_eq({u1, 'str'}, {u1, 'str'}))
+    t.assert(func_eq({u1, 'str'}, {u1_copy, 'str'}))
```
dokshina, line 155: Why not `t.assert(func_eq({u1, 'str'}, {u1, 'str'}))`?

test/integration/select_test.lua
```diff
@@ -96,6 +97,23 @@ local function insert_customers(g, customers)
     return inserted_objects
 end
 
+local function insert_goods(g, skus)
+    local inserted_objects = {}
+
+    for _, sku in ipairs(skus) do
```
dokshina, line 103: What does `skus` mean?

test/integration/select_test.lua
```diff
@@ -96,6 +97,23 @@ local function insert_customers(g, customers)
     return inserted_objects
 end
 
+local function insert_goods(g, skus)
```
dokshina, line 100: This function looks like a copy-paste of `insert_customers`.

What about `insert_objects(g, space_name, objects)`?

test/integration/select_test.lua
akudiyar: Since the test space is about goods, sku is an SKU (storage keeping unit), a known abbreaviation for goods units.

crud/common/types.lua
```diff
@@ -0,0 +1,23 @@
+local types = {}
+
+local lt_by_key_type = {
+    uuid = function (lhs, rhs)
+        return lhs:str() < rhs:str()
```
olegrok, line 5: Seems it will throw for nullable uuids

crud/common/types.lua
```diff
@@ -0,0 +1,23 @@
+local types = {}
+
+local lt_by_key_type = {
```
olegrok, line 3: I think such approach could break decimal support

crud/common/types.lua
```diff
@@ -0,0 +1,23 @@
...
+
+function types.lt(key_part)
+    return key_part and key_part.type and lt_by_key_type[key_part.type] or nil
+end
+
+function types.eq(key_part)
+    return key_part and key_part.type and eq_by_key_type[key_part.type] or nil
+end
+
+return types
```
olegrok, line 23: Empty line is missed

test/unit/select_comparators_test.lua
olegrok: Personally I don't see any difference between this values without detail reading. It's better to use something like "aaaaaaaa-aaaa-4000-b000-000000000001" and "bbbbbbbb-bbbb-4000-b000-000000000001"

test/unit/select_comparators_test.lua
akudiyar: I wanted to make the test safe from wrong equals implementation which treats different UUID instances as different objects (by ref, for example)

test/unit/select_comparators_test.lua
```diff
@@ -139,3 +141,24 @@ g.test_unicode_collations = function()
...
+    t.assert(not func_eq({u1, 'str'}, {u2, 'str'}))
+
+    local func_ge, err = select_comparators.gen_func(operators.GE, key_parts)
+    t.assert(err == nil)
+    t.assert(not func_ge({u1, 'str'}, {u2, 'str'}))
```
olegrok, line 160: This test doesn't looks correct because you assume the lexicographic sorting of strings is the same as for fields of uuid type in the Tarantool core.
Correct way:

```
key_def_lib = require('key_def')
key_def = key_def_lib.new({{field = 1, type = 'uuid'}})
-- It's slow way but we should compare this results with it (at least in tests)
-- because it's a rule how Tarantool sort UUIDs internally
key_def:compare({uuid1}, {uuid2})
```

test/integration/select_test.lua
```diff
@@ -805,3 +823,34 @@ add('test_select_with_collations', function(g)
...
+    })
+
+    table.sort(goods, function(obj1, obj2) return obj1.uuid:str() < obj2.uuid:str() end)
+
+    local result, err = g.cluster.main_server.net_box:call('crud.select', {'goods', nil})
+    t.assert_equals(err, nil)
+    local objects = crud.unflatten_rows(result.rows, result.metadata)
+    t.assert_equals(#objects, 4)
+
+    local conditions = {{'>', 'uuid', goods[3].uuid}}
```
olegrok, line 850: There is no test where field with uuid type is additional check

crud/common/types.lua
akudiyar: If it works, it will continue working. If a comparator is missing, it falls back to a common one.

test/unit/select_comparators_test.lua
akudiyar: I selected manually two predefined UUID which precede each other by type and by string representation.

test/integration/select_test.lua
akudiyar: +

crud/common/types.lua
akudiyar: fixed

crud/common/types.lua
akudiyar: fixed

test/unit/select_comparators_test.lua
akudiyar: Splitted into two lines.

crud/common/types.lua
akudiyar: fixed

crud/common/types.lua
akudiyar: fixed

test/integration/select_test.lua
akudiyar: fixed

crud/select/comparators.lua
akudiyar: I think that collations should be moved to the new types module also.

akudiyar: Moved collations to the "types" module

test/integration/select_test.lua
akudiyar: Added additoinal checks.

test/integration/select_test.lua
```diff
@@ -805,3 +806,55 @@ add('test_select_with_collations', function(g)
     local objects = crud.unflatten_rows(result.rows, result.metadata)
     t.assert_equals(objects, get_by_ids(customers, {2, 4}))
 end)
+
+if crud_utils.tarantool_supports_uuids() then
```
dokshina, line 810: Look at `skip_if` function of `luatest`.

test/unit/select_comparators_test.lua
```diff
@@ -100,10 +102,14 @@ end
 
 g.test_unicode_collations = function()
     local unicode_parts = {
-        {collation='unicode'}, {collation='unicode'}, {collation='unicode'},
+        {collation='unicode', type = 'string'},
```
dokshina, line 105: Please, fix formatting.

`{collation = 'unicode', type = 'string'},`

crud/common/utils.lua
```diff
@@ -120,9 +117,26 @@ local function tarantool_supports_fieldpaths()
...
+    return features.tarantool_supports_fieldpaths
+end
+
+function utils.tarantool_supports_uuids()
+    if features.tarantool_supports_uuids == nil then
+        determine_enabled_features()
+    end
 
-    return __tarantool_supports_fieldpaths
+    return features.tarantool_supports_uuids
```
dokshina, line 139: I like this approach, but what about naming?

For example,
`enabled_tarantool_features.fieldpaths`
`enabled_tarantool_features.uuids`

test/unit/select_comparators_test.lua
```diff
@@ -139,3 +145,25 @@ g.test_unicode_collations = function()
...
+    t.assert(func_eq({u1, 'str'}, {u1_copy, 'str'}))
+    t.assert(not func_eq({u1, 'str'}, {u2, 'str'}))
+
+    local func_ge, err = select_comparators.gen_func(operators.GE, key_parts)
+    t.assert(err == nil)
+    t.assert(not func_ge({u1, 'str'}, {u2, 'str'}))
+    t.assert(func_ge({u1, 'str'}, {u1, 'str'}))
+    t.assert(func_ge({u1, 'ttr'}, {u1, 'str'}))
+    t.assert(func_ge({u2, 'str'}, {u1, 'str'}))
+end
```
dokshina, line 169: What about `unit/select_filters_test`?

crud/select/filters.lua
```diff
@@ -164,7 +170,7 @@ local function format_value(value)
     elseif type(value) == 'number' then
         return tostring(value)
     elseif type(value) == 'cdata' then
-        return tostring(value)
+        return cdata_formats[value_type] and cdata_formats[value_type](value) or tostring(value)
```
dokshina, line 173: Function takes `value_type`, but is uses both `type(value)` and `value_type`. BTW, `box.NULL` is also `cdata`.

Maybe it would be better to [check](https://github.com/tarantool/checks/blob/master/checks.lua#L237) UUID type directly. 

I don't see any reason to separate UUID string formatting - this function is responsible for this:

```lua
elseif utils.is_uuid(value) then
    return ("%q"):format(value)
...
```

crud/select/filters.lua
```diff
@@ -491,6 +510,22 @@ local function lt_boolean_strict(lhs, rhs)
     return (not lhs) and rhs
 end
 
+local function lt_uuid_nullable(lhs, rhs)
+    if lhs == nil and rhs ~= nil then
+        return true
+    elseif rhs == nil then
+        return false
+    end
+    return lhs:str() < rhs
```
dokshina, line 519: How should I understand why `lhs:str()` us compared with `rhs`? It should be obvious by arguments names. 
For example, `lt_uuid_nullable(operand, value)`.

crud/select/filters.lua
```diff
@@ -251,29 +257,36 @@ local function add_collation_postfix(func_name, value_opts)
...
+    elseif value_type == 'uuid' then
+        func_name = 'eq_uuid'
     end
 
-    return cond_strings
+    return func_name
 end
 
-local function format_lt(cond)
+local function format_func(func_name_getter, cond)
```
dokshina, line 289: I sure that this refactoring can be performed in a separated patch.

crud/common/types.lua
```diff
@@ -0,0 +1,109 @@
...
+
+function types.comparators(key_part)
+    if key_part and key_part.type and functions_by_key_type[key_part.type] then
+        return functions_by_key_type[key_part.type](key_part)
+    else
+        return lt_default, eq
+    end
+end
+
+return types
```
dokshina, line 109: I have questions to this module.

1. Why is it called `types`? It exports only one function that takes `key_part` (not type) and returns functions.

2. Why this module is placed in common scope? It's used only by `comparators` module, it's too specific (it can't be used for `filters` module, for example). I still don't see any reasons to separate this logic.

crud/common/types.lua
```diff
@@ -0,0 +1,109 @@
...
+    uuid = function ()
+        return comparing_nil(lt_uuid), eq
+    end
+}
+
+function types.comparators(key_part)
+    if key_part and key_part.type and functions_by_key_type[key_part.type] then
+        return functions_by_key_type[key_part.type](key_part)
+    else
+        return lt_default, eq
```
dokshina, line 105: What is the difference between `lt` and `lt_default`? Where is `eq_default`?

crud/common/types.lua
```diff
@@ -0,0 +1,109 @@
+local errors = require('errors')
+
+local collations = require('crud.common.collations')
+local TypeMismatchError = errors.new_class('TypeMismatchError')
+local UnsupportedCollationError = errors.new_class('UnsupportedCollationError')
+
+local types = {}
+
+local function comparing_nil(cmp)
```
dokshina, line 9: Is it work for all cmp functions or only for `lt`?

crud/common/types.lua
```diff
@@ -0,0 +1,109 @@
...
+        else
+            UnsupportedCollationError:assert(false, 'Unsupported Tarantool collation %q', collation)
+        end
+    end,
+    uuid = function ()
+        return comparing_nil(lt_uuid), eq
+    end
+}
+
+function types.comparators(key_part)
```
dokshina, line 101: `key_parts` can't be a `nil` (as well as `key_part.type`).

crud/common/types.lua
```diff
@@ -0,0 +1,109 @@
...
+    if key_part and key_part.type and functions_by_key_type[key_part.type] then
+        return functions_by_key_type[key_part.type](key_part)
```
dokshina, line 103: Why function by key type accepts `key_part`? (It's simply ignored for all types except `string`)

crud/common/types.lua
akudiyar: it works for any, for example, `lt_boolean` and `lt_uuid` are decorated with it

crud/common/types.lua
akudiyar: `functions_by_key_type ` return type-aware comparators, which accept type metadata (includes type, collation and whatever else)

crud/common/types.lua
akudiyar: There is no `eq_default` because there is no difference between them. Seems that `lt` itself is used once, so `lt_default` may be renamed to `lt`.

crud/select/filters.lua
akudiyar: THis refactoring is related to the two new functions: `lt_uuid` and `eq_uuid`

crud/common/utils.lua
olegrok: We can't call it just "uuids" since exact feature name "indexes by uuid" (https://github.com/tarantool/tarantool/commit/b238def8065d20070dcdc50b54c2536f1de4c7c7)

crud/select/filters.lua
olegrok: My suggestion is `tostring(lhs) < tostring(rhs)`

crud/common/types.lua
olegrok: It's not "comparing_nil" it's "comparing_nullable".

So, your usage of this function is a bit awful. For each call you create new closure (at least it makes suffer Lua GC).
Seems you need to predefine comparators and then use them.

crud/common/types.lua
olegrok: Also I have a question why is it called "functions_by_key_type". First of all, "functions"... why not e.g. "comparators"? Function is too abstract thing. Also "key_type"... Why is it not just "type"?

test/unit/select_comparators_test.lua
akudiyar: Added new test `unit/select_filters_uuid_test`

test/unit/select_comparators_test.lua
akudiyar: Fixed

test/integration/select_test.lua
akudiyar: Added skip_if

crud/select/filters.lua
akudiyar: Renamed rhs values

crud/select/filters.lua
akudiyar: Added `utils.is_uuid` method, removed `cdata_formats` and `value_type` checking

crud/common/utils.lua
akudiyar: Renamed

crud/common/types.lua
akudiyar: Renamed the module and moved it to the `select` folder

crud/common/types.lua
akudiyar: Renamed `lt_default` to `lt`

crud/common/types.lua
akudiyar: Removed nullability checks

crud/select/type_comparators.lua
```diff
@@ -6,7 +6,7 @@ local UnsupportedCollationError = errors.new_class('UnsupportedCollationError')
 
 local types = {}
 
-local function comparing_nil(cmp)
+local function comparing_nullable(cmp)
```
dokshina, line 9: It looks like it's `lt_nullable`, not comparing.

crud/select/filters.lua
```diff
@@ -163,10 +164,13 @@ local function format_value(value)
         return ("%q"):format(value)
     elseif type(value) == 'number' then
         return tostring(value)
-    elseif type(value) == 'cdata' then
-        return tostring(value)
     elseif type(value) == 'boolean' then
         return tostring(value)
+    elseif type(value) == 'cdata' then
```
dokshina, line 169: Seems that `cdata` check is redundant.

crud/select/type_comparators.lua
akudiyar: But it is not. There are three `lt` implementations using the same `nil` values comparison.

crud/select/filters.lua
akudiyar: `cdata` was here before, removing it will change the type checking logic.

olegrok: No, cdata is needed e.g. for "decimals" and cdata numbers

```
tarantool> type(1ULL)
---
- cdata
...
```

crud/select/type_comparators.lua
dokshina: Does it work for `eq`, `gt` and so on? If yes - it's really `comparing`. If it works only for `lt` - it's `lt`-specific comparison, and function name should reflect it.

crud/select/filters.lua
dokshina: It seems that my comment wasn't correct.

I mean that UUID check can be performed without checking cdata before.

```lua
elseif utils.is_uuid(value) then
...
elseif type(value) == 'cdata' then
...
```

crud/select/filters.lua
mtrempoltsev: Just add these functions to the existing file. Please keep patch simpler. And we will refactor in the next patch.

crud/select/filters.lua
akudiyar: I don't agree. It looks clear and uses the desired field type, not the runtime type of the value.
Btw, checking for uuid type requires checking for cdata too:

```
tarantool> ffi = require'ffi'
---
...

tarantool> ffi.istype(nil, 'uuid')
---
- error: 'bad argument #1 to ''?'' (C type expected, got nil)'
...

tarantool> type(nil)
---
- nil
...

tarantool> ffi.istype(1, 'uuid')
---
- error: 'bad argument #1 to ''?'' (C type expected, got number)'
...

tarantool> type(1)
---
- number
...
```

Feel free to refactor this place in further patches.

crud/select/type_comparators.lua
akudiyar: Renamed

crud/select/filters.lua
akudiyar: Moved common code back to the existing functions

crud/select/filters.lua
akudiyar: The checks in the comment above are wrong, (wrong order of arguments):

```
tarantool> ffi = require'ffi'
---
...

tarantool> local uuid_t = ffi.typeof('struct tt_uuid')
---
...

tarantool> uuid_t = ffi.typeof('struct tt_uuid')
---
...

tarantool> ffi.istype(uuid_t, nil)
---
- false
...

tarantool> ffi.istype(uuid_t, 1)
---
- false
...

tarantool> ffi.istype(uuid_t, "value")
---
- false
...

tarantool> ffi.istype(uuid_t, box.NULL)
---
- false
...
```

Moved the check as suggested.

dokshina (approved):

dokshina:
Please, rebase your branch and I will merge it immediately.

PR #87, opened 2020-11-13
Add tests for #73 and #82

### LINKED ISSUE #73: Return LuaFun-compatible iterator as a result of pairs()
There are many scenarios for filtering, mapping and reducing the returned data and the most comfortable way of doing it is using the LuaFun library. Simple Lua iterator is not enough.

### LINKED ISSUE #74: Implement count() method
Use cases: 
a) customer wants to see counts of templates in message template catalogs in application UI. The templates and catalogs are stored in vshard, and operated via CRUD through a connector.
b) pagination -- we need to know the total amount of results for displaying it in the UI.

The count() method must accept conditions.

### LINKED ISSUE #82: crud.pairs() returns one nil object when iterating over an empty space
Expected:

crud.pairs() returns empty iterator (next() returns nil)

Actual:

crud.pairs() return non-empty iterator with one `nil` value

Steps to reproduce:

1) Create an empty sharded space
2) Call crud.pairs() on the router in a for loop and check for the elements returned

```
    for _, record in crud.pairs('test_space', {use_tomap = true}) do log.info(record) end
```

olegrok:
This patch doesn't introduce anything new, it simply adds two tests for two features.

Follow-up #33
Closes #73
Closes #82

### DIFF
--- a/test/integration/pairs_test.lua
+++ b/test/integration/pairs_test.lua
@@ -353,3 +353,15 @@ add('test_negative_first', function(g)
         ]])
     end)
 end)
+
+add('test_empty_space', function(g)
+    local count = g.cluster.main_server.net_box:eval([[
+        local crud = require('crud')
+        local count = 0
+        for _, object in crud.pairs('customers') do
+            count = count + 1
+        end
+        return count
+    ]])
+    t.assert_equals(count, 0)
+end)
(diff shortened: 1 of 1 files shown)


olegrok:
Closed in favour of #93

PR #93, opened 2020-11-25
Iterator improvements

### LINKED ISSUE #73: Return LuaFun-compatible iterator as a result of pairs()

### LINKED ISSUE #82: crud.pairs() returns one nil object when iterating over an empty space

olegrok:
This patchset just ports of https://github.com/tarantool/crud/pull/87 to crud without merger 

Closes #73 
Closes #82

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Fixed select by part of sharding key equal. Before this patch
   selecting by equality of partially specified multipart primary index
   value was misinterpreted as a selecting by fully specified key value.
+* Fixed iteration with `pairs` through empty space returned `nil`.
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


dokshina:
You forget about CHANGELOG (and maybe we should note in README that `pairs` returns lua-fun compatible iterator).

dokshina (approved):
Great!

olegrok:
> CHANGELOG

Added

PR #97, opened 2020-12-04
cleanup query filters code

olegrok:
This patch removes some strange and complex logic when initially
we collect all field numbers of index but then match this numbers
to index fields to get type of every field.

Also redundant loop by indexes is removed since indexes and can be
obtained by name directly from indexes table.

### DIFF
--- a/crud/select/filters.lua
+++ b/crud/select/filters.lua
@@ -52,57 +52,27 @@ local function get_index_fieldnos(index)
     return index_fieldnos
 end
 
-local function get_values_types(space_format, fieldnos)
-    local values_types = {}
+local function get_index_fields_types(index)
+    local index_fields_types = {}
 
-    for _, fieldno in ipairs(fieldnos) do
-        local field_format = space_format[fieldno]
-        assert(field_format ~= nil)
-
-        table.insert(values_types, field_format.type)
+    for _, part in ipairs(index.parts) do
+        table.insert(index_fields_types, part.type)
     end
 
-    return values_types
+    return index_fields_types
 end
 
-local function get_values_opts(index, fieldnos)
-    local values_opts = {}
-    for _, fieldno in ipairs(fieldnos) do
-        local is_nullable = true
-        local collation
+local function get_values_opts(index)
+    local index_field_opts = {}
 
-        if index ~= nil then
-            local index_part
-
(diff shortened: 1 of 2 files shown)


crud/select/filters.lua
```diff
@@ -127,15 +97,22 @@ local function parse(space, conditions, opts)
...
             elseif fieldnos_by_names[condition.operand] ~= nil then
-                fieldnos = {
-                    fieldnos_by_names[condition.operand],
-                }
+                local fiendno = fieldnos_by_names[condition.operand]
+                fieldnos = {fiendno}
+                local field_format = space_format[fiendno]
+                field_types = {field_format.type}
+                local is_nullable = field_format.is_nullable == true
+                values_opts = {{is_nullable = is_nullable, collation = nil}}
```
dokshina, line 115: Maybe 

```lua
values_opts = {
    {is_nullable = is_nullable, collation = nil},
}
```

?

crud/select/filters.lua
```diff
@@ -127,15 +97,22 @@ local function parse(space, conditions, opts)
         if i ~= opts.scan_condition_num then
             -- Index check (including one and multicolumn)
             local fieldnos
+            local field_types
+            local values_opts
```
dokshina, line 101: Field types, but valueS opts. Probably, we need to choose one approach.

dokshina (commented):
Thanks a lot for your refactoring, please check my minor comments.

crud/select/filters.lua
olegrok: Done

crud/select/filters.lua
olegrok: It seems to be ok. Because value can has several options but only one type.

dokshina: No, it's not about opt and opts, it's about valueS and field. I mean we need to use `values_opts`+ `fields_types` or `value_opts` + `field_types`.

olegrok: Well. Renamed `field_types` -> `fields_types` (https://github.com/tarantool/crud/compare/078e1962a1bf4e27ccd950cd36c0fbca1a31249c..7caeffad3cc937fa7380356b553004545d715112)

dokshina (approved):

PR #108, opened 2021-01-20
Fixed select crash when dropping indexes.

### LINKED ISSUE #103: [1pt] Select could fail in case of sparsed indexes array
https://github.com/tarantool/crud/commit/e5eb7ca3076bd61451b5ceccd7338c07cc96351e#diff-8afead6b3cb6ef4117be97c1bf3ac723a454b7ddff094d300e42c20431b37bf0R42

A problem is following:
```lua
box.cfg{}
box.schema.space.create('space')
box.space.space:create_index('i1')
box.space.space:create_index('i2')
box.space.space:create_index('i3')
box.space.space:create_index('i4')
box.space.space.index.i2:drop()
box.space.space.index.i3:drop()

tarantool> #box.space.space.index  -- Ooops
---
- 0
...
tarantool> box.space.space.index[0]
---
- unique: true
  parts:
  - type: unsigned
    is_nullable: false
    fieldno: 1
  type: TREE
  id: 0
  space_id: 512
  name: i1
...

tarantool> box.space.space.index[1]
---
- null
...

tarantool> box.space.space.index[2]
---
- null
...

tarantool> box.space.space.index[3]
---
- unique: true
  parts:
  - type: unsigned
    is_nullable: false
    fieldno: 1
  id: 3
  type: TREE
  space_id: 512
  name: i4
...
```

Seems crud should consider it. See `table.maxn`

mRrvz:
Fixed selection error when dropping indexes by replacing the ``#`` operator with a call to ``table.maxn`` when creating a select plan.

Closes #103.

### DIFF
--- a/crud/select/plan.lua
+++ b/crud/select/plan.lua
@@ -15,19 +15,23 @@ local function index_is_allowed(index)
 end
 
 local function get_index_for_condition(space_indexes, space_format, condition)
-    for i= 0, #space_indexes do
+    for i= 0, table.maxn(space_indexes) do
         local index = space_indexes[i]
-        if index.name == condition.operand and index_is_allowed(index) then
-            return index
+        if index ~= nil then
+            if index.name == condition.operand and index_is_allowed(index) then
+                return index
+            end
         end
     end
 
-    for i = 0, #space_indexes do
+    for i = 0, table.maxn(space_indexes) do
         local index = space_indexes[i]
-        local first_part_fieldno = index.parts[1].fieldno
-        local first_part_name = space_format[first_part_fieldno].name
-        if first_part_name == condition.operand and index_is_allowed(index) then
-            return index
+        if index ~= nil then
+            local first_part_fieldno = index.parts[1].fieldno
+            local first_part_name = space_format[first_part_fieldno].name
+            if first_part_name == condition.operand and index_is_allowed(index) then
+                return index
+            end
         end
     end
 end
@@ -39,18 +43,40 @@ local function validate_conditions(conditions, space_indexes, space_format)
     end
 
     local index_names = {}
-    for i = 0, #space_indexes do
+
+    --local file = io.open("test100500.txt", "w+")
+    --io.output(file)
+    --io.close(file)
+    for i = 0, table.maxn(space_indexes) do
         local index = space_indexes[i]
         if index ~= nil then
             index_names[index.name] = true
+            --[=====[
+            --if index.name == "name_tree" then
+                io.write(index.name)
+                io.write("\n")
+                for _, condition in ipairs(conditions) do
+                    io.write(string.format("%q", index_names[condition.operand]))
+                    io.write("\n")
+                    io.write(string.format("%q", field_names[condition.operand]))
+                    io.write("\n")
+                    io.write(condition.operand)
+                    io.write("\n")
+                end
+                io.write("cycle end\n")
+            --end
+            --]=====]
         end
     end
+ --   io.close(file)
 
+    --io.output(file)
     for _, condition in ipairs(conditions) do
         if index_names[condition.operand] == nil and field_names[condition.operand] == nil then
             return false, ValidateConditionsError:new("No field or index %q found", condition.operand)
         end
     end
+   -- io.close(file)
 
     return true
 end
--- a/crud/select/plan.lua
+++ b/crud/select/plan.lua
@@ -15,7 +15,7 @@ local function index_is_allowed(index)
 end
 
 local function get_index_for_condition(space_indexes, space_format, condition)
-    -- If we use # (not table.maxn), we may lose indexes, 
+    -- If we use # (not table.maxn), we may lose indexes,
     -- when user drop some indexes.
     local max_index = table.maxn(space_indexes)
     for i = 0, max_index do
--- a/crud/select/plan.lua
+++ b/crud/select/plan.lua
@@ -44,39 +44,18 @@ local function validate_conditions(conditions, space_indexes, space_format)
 
     local index_names = {}
 
-    --local file = io.open("test100500.txt", "w+")
-    --io.output(file)
-    --io.close(file)
     for i = 0, table.maxn(space_indexes) do
         local index = space_indexes[i]
         if index ~= nil then
             index_names[index.name] = true
-            --[=====[
-            --if index.name == "name_tree" then
-                io.write(index.name)
-                io.write("\n")
-                for _, condition in ipairs(conditions) do
-                    io.write(string.format("%q", index_names[condition.operand]))
-                    io.write("\n")
-                    io.write(string.format("%q", field_names[condition.operand]))
-                    io.write("\n")
-                    io.write(condition.operand)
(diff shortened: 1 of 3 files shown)


crud/select/plan.lua
```diff
@@ -15,19 +15,23 @@ local function index_is_allowed(index)
 end
 
 local function get_index_for_condition(space_indexes, space_format, condition)
-    for i= 0, #space_indexes do
+    for i= 0, table.maxn(space_indexes) do
```
olegrok, line 18: Hi! Thanks for your patch!

Could you move length calculation into separate variable that could be used here and below?
I'll place some benchmarks below. It's too expensive to call `table.maxn` twice.

Also could you add a space between "i" and "=" `i= 0` -> `i = 0` since you change this place?

```lua
local clock = require('clock')

local t = {}
for i = 1, 1e5 do
    t[i] = i
end

local s = clock.time()
local _
for _ = 1, 1e5 do
    _ = table.maxn(t)
end
print('maxn', clock.time() - s)

s = clock.time()

local _
for _ = 1, 1e5 do
    _ = #t
end
print('#', clock.time() - s)

-- tarantool test.lua 
-- maxn	6.1955099105835
-- #	0.012683868408203
```

olegrok: However, my benchmark could be not relevant since amount of indexes is restricted with 65k:

For arrays with length ~1k table.maxn even faster than `#`. But anyway if we could call it only once we should do it.

test/unit/select_dropped_indexes_test.lua
```diff
@@ -0,0 +1,136 @@
...
+
+    customers:create_index('name_tree', {
+        type = 'TREE',
+        parts = {'name'},
+        unique = false,
+        if_not_exists = true,
+    })
+
+    customers.index.name_hash:drop()
+    customers.index.age_hash:drop()
```
dokshina, line 65: You've copied this code from test that checks smth related to index type, thats why indexes are named `*_<hash,tree>`. But these tests aren't about it. I suggest naming indexes just `index1, index2, ...` because we want to test that removing index doesn't affect iterating over indexes, so these names are good IMHO.

dokshina: You can also name some of them `index_dropped` or smth like this.

test/unit/select_dropped_indexes_test.lua
```diff
@@ -0,0 +1,136 @@
...
+
+    t.assert_equals(plan.conditions, conditions)
+    t.assert_equals(plan.space_name, 'customers')
+    t.assert_equals(plan.index_id, 2)
+    t.assert_equals(plan.scan_value, {20})
+    t.assert_equals(plan.after_tuple, nil)
+    t.assert_equals(plan.scan_condition_num, 1)
+    t.assert_equals(plan.iter, box.index.EQ)
+    t.assert_equals(plan.total_tuples_count, nil)
+    t.assert_equals(plan.sharding_key, nil)
```
dokshina, line 99: Maybe it's enough to check only these conditions:

```
 t.assert_equals(plan.conditions, conditions)
    t.assert_equals(plan.space_name, 'customers')
    t.assert_equals(plan.index_id, 2)
```

test/unit/select_dropped_indexes_test.lua
```diff
@@ -0,0 +1,118 @@
...
+
+    customers:create_index('index6', {
+        type = 'TREE',
+        parts = {'name'},
+        unique = false,
+        if_not_exists = true,
+    })
+
+    customers.index.index_dropped1:drop()
+    customers.index.index_dropped2:drop()
```
olegrok, line 65: It would be great to add some check to show that `#customers.index == 3` but actually there are 4 indexes.

crud/select/plan.lua
```diff
@@ -15,19 +15,24 @@ local function index_is_allowed(index)
 end
 
 local function get_index_for_condition(space_indexes, space_format, condition)
-    for i= 0, #space_indexes do
+    local maxn = table.maxn(space_indexes)
```
dokshina, line 18: Please, add a comment why do we need `maxn` here.

dokshina: And I don't sure that `maxn` is a good name for this variable.

test/unit/select_dropped_indexes_test.lua
```diff
@@ -0,0 +1,118 @@
...
+    })
+
+    customers:create_index('index3', {
+        type = 'TREE',
+        parts = {'age'},
+        unique = false,
+        if_not_exists = true,
+    })
+
+    customers:create_index('index_dropped1', {
```
dokshina, line 45: A nit: maybe smth like `index4_dropped`?

test/unit/select_dropped_indexes_test.lua
```diff
@@ -0,0 +1,120 @@
...
+    customers.index.index4_dropped:drop()
+    customers.index.index5_dropped:drop()
+end
+
+g.after_all = function()
+    box.space.customers:drop()
+end
+
+g.test_conditions = function()
+    t.assert(#box.space.customers.index ~= table.maxn(box.space.customers.index))
```
dokshina, line 73: I don't understand such test, @olegrok, was it your suggestion?

test/unit/select_dropped_indexes_test.lua
```diff
@@ -0,0 +1,120 @@
...
+end
+
+g.test_dropped_index_call = function()
```
dokshina, line 76: This test checks what? That removing indexes works?)

test/unit/select_dropped_indexes_test.lua
olegrok: yes, it was my suggestion

dokshina: I need a comment to understand why do we need such test.

Moreover, I sure that it shouldn't be a separated test. I think that such assertions should be performed for all test group (I mean `before_all` section), and should lead to fail of all group tests.

olegrok: My suggestion was just add a check. It's needed to check initial conditions correctness that test actually covers a problem. Yes, I agree that assert in before_all should be enough.

test/unit/select_dropped_indexes_test.lua
```diff
@@ -63,24 +63,13 @@ g.before_all = function()
...
     customers.index.index4_dropped:drop()
     customers.index.index5_dropped:drop()
-end
-
-g.after_all = function()
-    box.space.customers:drop()
-end
 
-g.test_conditions = function()
+    -- We need this check to make sure that tests actually covers a problem.
```
olegrok, line 67: typo: `tests actually cover` (without "s").

dokshina:
Thanks a lot for your patch! Please, update Changelog.

I would like you to make your comment about why do we use `table.maxn` more clear (probably add an example). And since we have 2 places where it's used, we need two comments. Think about people who will read this code sometimes - they should understand it without suffering.

dokshina (approved):

PR #109, opened 2021-01-21
Added new examples of select and pairs queries.

### LINKED ISSUE #49: [1pt] Add more select/pairs examples to README.
I'd love to see more complex examples regarding select.
How do I do paginated select easily? Snippet would be lovely
How do I select using composite index?
How do I perform select using partial key (for composite indexes)?
Can I use luafun (filter/take/_while/etc) with pairs? (An example would be lovely as well)

mRrvz:
Closes #49

### DIFF
--- a/doc/pairs.md
+++ b/doc/pairs.md
@@ -1,6 +1,8 @@
 # Pairs examples
 
-With ``crud.pairs``, you can iterate across a distributed space. The arguments are the same as [``crud.select``](https://github.com/tarantool/crud/docs/select.md), except or the ``use_tomap`` parameter. Below are examples that may help you.
+With ``crud.pairs``, you can iterate across a distributed space.  
+The arguments are the same as [``crud.select``](https://github.com/tarantool/crud/docs/select.md), except of the ``use_tomap`` parameter.  
+Below are examples that may help you.
 
 ## Getting space
 
@@ -16,11 +18,11 @@ end
 
 tuples
 ---
-- - - 1
-    - 7331
-    - Alexey
-    - Adams
-    - 20
+- - - 1 -- id
+    - 7331 -- bucket_id
+    - Alexey -- name
+    - Adams -- surname
+    - 20 -- age
   - - 2
     - 899
     - Sergey
@@ -73,7 +75,8 @@ objects
 
 ## Pagination
 
-``crud.pairs``, like [``crud.select``](https://github.com/tarantool/crud/doc/select#pagination), supports pagination. To use it, combine the ``first`` and ``after`` parameters. 
+``crud.pairs``, like [``crud.select``](https://github.com/tarantool/crud/doc/select#pagination), supports pagination.  
+To use it, combine the ``first`` and ``after`` parameters. 
 
 **Example:**
 
--- a/doc/pairs.md
+++ b/doc/pairs.md
@@ -21,6 +21,8 @@ local age_sum = fun.reduce(function(acc, x) return acc + x.age end, 0, crud.pair
 ```lua
 local objects = {}
 for _, obj in fun.map(function(x) return {obj.id, obj.name, obj.age * 2} end, crud.pairs('customers', nil, {use_tomap = true}))
+    table.insert(objects, obj)
+end
 ```
 
 **Take example**:
--- a/doc/pairs.md
+++ b/doc/pairs.md
@@ -1,35 +1,141 @@
 # Pairs examples
 
+With ``crud.pairs``, you can iterate across a distributed space. The arguments are the same as [``crud.select``](https://github.com/tarantool/crud/docs/select.md), except or the ``use_tomap`` parameter. Below are examples that may help you.
+
+## ``Use_tomap`` parameter
+
+With ``use_tomap`` flag, you can choose to iterate over objects or over tuples. If ``use_tomap = false``, you will iterate over tuples. This parameter is false by default.
+
+**Example:**
+
+```lua
+tuples = {}
+for _, tuple in crud.pairs('customers', nil, {use_tomap = false}) do
+    table.insert(tuples, tuple)
+end
+
+tuples
+---
+- - - 1
+    - 2313
+    - Alexey
+    - 20
+  - - 2
+    - 241
+    - Vladimir
+    - 18
+  - - 3
+    - 571
+    - Alexander
+    - 24
+...
+```
+
+If ``use_tomap = true``, you will iterate over objects.
+
+**Example:**
+
+```lua
+objects = {}
+for _, obj in crud.pairs('customers', nil, {use_tomap = true}) do
+    table.insert(tuples, tuple)
+end
+
+objects
+---
+- - id: 1
+    bucket_id: 2313
+    name: Alexey
+    age: 20
+  - id: 2
+    bucket_id: 241
+    name: Vladimir
+    age: 14
+  - id: 3
+    bucket_id: 571
+    name: Alexander
+    age: 24
+...
+```
+
+## Pagination
+
+``crud.pairs``, like [``crud.select``](https://github.com/tarantool/crud/doc/select#pagination), supports pagination. To use it, combine the ``first`` and ``after`` parameters. 
+
+**Example:**
+
+```lua
+TODO
+```
+
+Note that ``crud.pairs``, unlike ``crud.select``, **don't support reverse pagination.**
+
 ## Lua Fun
 
-Queries using pairs support the [Lua Fun](https://github.com/luafun/luafun) library. Some examples of working with basic functional functions below.
+[``Pairs``](https://github.com/tarantool/crud#pairs) is [Lua Fun](https://github.com/luafun/luafun) compatible. Some examples of working with basic functional functions below.
 
 **Filter example:**
+
 ```lua
-local objects = {}
-for _, obj in fun.filter(function(x) return x.age % 5 == 0 end, crud.pairs('customers', {{'==', 'name', 'Alexey'}}, {use_tomap = true})) do
+objects = {}
+for _, obj in crud.pairs('customers', {{'>=', 'age', 20}}, {use_tomap = true}):filter(function(x) return x.age % 5 == 0 end) do
     table.insert(objects, obj)
 end
+
+objects
+---
+- - id: 1
+    bucket_id: 2313
+    name: Alexey
+    age: 20
+...
 ```
 
 **Reduce (foldl) example:**
+
 ```lua
-local age_sum = fun.reduce(function(acc, x) return acc + x.age end, 0, crud.pairs('customers', nil, {use_tomap = true}))
+age_sum = crud.pairs('customers', nil, {use_tomap = true}):reduce(function(acc, x) return acc + x.age end, 0)
+age_sum
+---
+- 30
+....
 ```
 
 **Map example:**
+
 ```lua
-local objects = {}
-for _, obj in fun.map(function(x) return {obj.id, obj.name, obj.age * 2} end, crud.pairs('customers', nil, {use_tomap = true}))
+objects = {}
+for _, obj in crud.pairs('customers', nil, {use_tomap = true}):map(function(x) return {id = obj.id, name = obj.name, age = obj.age * 2}) do
     table.insert(objects, obj)
 end
+
+objects
+---
+- - id: 1
+    name: Alexey
+    age: 40
+  - id: 2
+    name: Vladimir
+    age: 28
+  - id: 3
+    name: Alexander
+    age: 48
+...
 ```
 
 **Take example**:
 
 ```lua
-local tuples = {}
-for _, tuple in fun.take(3, crud.pairs('customers', {{'>=', 'age', 18}})) do
+tuples = {}
+for _, tuple in crud.pairs('customers', {{'>=', 'age', 18}}):take(1) do
     table.insert(tuples, tuple)
 end
+
+tuples
+---
+- - - 1
+    - 2313
+    - Alexey
+    - 20
+...
 ```
--- a/doc/pairs.md
+++ b/doc/pairs.md
@@ -2,16 +2,16 @@
 
 With ``crud.pairs``, you can iterate across a distributed space. The arguments are the same as [``crud.select``](https://github.com/tarantool/crud/docs/select.md), except or the ``use_tomap`` parameter. Below are examples that may help you.
 
-## ``Use_tomap`` parameter
+## Getting space
 
-With ``use_tomap`` flag, you can choose to iterate over objects or over tuples. If ``use_tomap = false``, you will iterate over tuples. This parameter is false by default.
+Let's check ``developers`` space contents to make other examples more clear. Just select first 4 values without conditions.
 
 **Example:**
 
 ```lua
 tuples = {}
-for _, tuple in crud.pairs('developers', nil, { use_tomap = false, first = 3 }) do
-    table.insert(tuples, tuple)
+for _, tuple in crud.pairs('developers', nil, { first = 4 }) do
+  table.insert(tuples, tuple)
 end
 
 tuples
@@ -31,10 +31,17 @@ tuples
     - Pavel
     - Adams
     - 27
+  - - 4
+    - 501
+    - Mikhail
+    - Liston
+    - 51
 ...
 ```
 
-If ``use_tomap = true``, you will iterate over objects.
+## ``Use_tomap`` parameter
+
+With ``use_tomap`` flag, you can choose to iterate over objects or over tuples. If ``use_tomap = true``, you will iterate over objects. This parameter is false by default.
 
 **Example:**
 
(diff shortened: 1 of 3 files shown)


doc/pairs.md
```diff
@@ -0,0 +1,35 @@
+# Pairs examples
+
+## Lua Fun
+
+Queries using pairs support the [Lua Fun](https://github.com/luafun/luafun) library. Some examples of working with basic functional functions below.
```
olegrok, line 5: I'm not sure that `Queries support LuaFun` is correct. "pairs" is "luafun"-compatible or something like that.
Iterator can't support library - it's not clear.

doc/pairs.md
```diff
@@ -0,0 +1,35 @@
...
+
+**Filter example:**
+```lua
+local objects = {}
+for _, obj in fun.filter(function(x) return x.age % 5 == 0 end, crud.pairs('customers', {{'==', 'name', 'Alexey'}}, {use_tomap = true})) do
```
olegrok, line 10: I think it's better to rewrite to
```lua
pairs('customers', {{'==', 'name', 'Alexey'}}, {use_tomap = true})):filter(function(x) return x.age % 5 == 0 end)
```

See https://github.com/tarantool/crud/blob/2bb0c29dd8453dba27053143f967ac273e8cc65e/test/integration/pairs_test.lua#L359 for example

doc/select.md
```diff
@@ -0,0 +1,84 @@
+# Select examples
+
+## Pagination
+
+Select supports pagination. To use it, use the [``after``](https://github.com/tarantool/crud#select) and [``first``](https://github.com/tarantool/crud#select) parameter.
```
olegrok, line 5: There is no such section for pairs. It would be great to list all possible options and then show how they help for users.

doc/select.md
```diff
@@ -0,0 +1,84 @@
+# Select examples
```
olegrok, line 1: It would be grate to say somewhere if we specify sharding key (in our case it's primary key always) or bucket_id (https://github.com/tarantool/crud/commit/d7d45af3d962a8c347f63c4253742c211f851bcf) select will be performed on single node otherwise map-reduce over all nodes will be occurred.

olegrok (commented):
Hi! Thanks for your patch, consider several comments below.

doc/select.md
dokshina, line 1: The main idea of these examples is to show `crud.select` capabilities, but also to teach user how to use it.

I think that select examples should be reordered:

* Filters - show using conditions by index, by composite index, by first part of index, by non-indexed field (add a note that in this case full-scan is performed). We also need to show that condition by indexed field should be placed first to avoid full-scan.

* Pagination - first, after, first and after together, reverse pagination.

Maybe we can show the simple loop that scans space and shows results by pages, or add a comments to your examples, smth like "Ok, we've got first 3 tuples, let's get next 3 tuples - specify row[3] as an `after`".

doc/pairs.md
dokshina, line 5: I think that we have to show more examples that shows `crud.pairs`, not `luafun`.

First of all, we need to show difference between calls with and without `use_tomap`. Moreover, I would like to see returned data, just like in `select` examples.

We also need to show pairs with pagination (be careful, reverse pagination isn't supported here).

BTW, what do you think about showing select without conditions before all examples to understand space contents. It can be helpful to understand filters and pagination.

doc/select.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Filtering
+
+The second parameter passed to ``crud.select`` is an array of conditions. Below are examples of filtering data using these conditions.
```
dokshina, line 5: Maybe smth like 

CRUD allows to filter tuples by conditions.
Each condition can use field name/number or index name. The first condition that uses index name is used to iterate over space (if there is no conditions that match index names, full scan is performed). Other conditions are used as additional filters.

doc/select.md
```diff
@@ -0,0 +1,204 @@
...
+The second parameter passed to ``crud.select`` is an array of conditions. Below are examples of filtering data using these conditions. 
+
+### Getting full space
+
+To get a full space without filtering, you need to pass ``nil`` as a condition.
+
+**Example:**
+
+```lua
+crud.select('customers', nil)
```
dokshina, line 14: This is a bad approach to select all space contents. I think we should show user that requests without conditions should be performed only with `first` specified.

You can write smth like:

Let's check `customers` space contents to make other examples more clear. Just select first 10 values without conditions.

doc/select.md
```diff
@@ -0,0 +1,204 @@
...
+
+You can also make a selection using a non-indexed field.
+
+**Example:**
+
+```lua
+res, err = crud.select('customers', {{'>=', 'id', 3}})
+res.rows
+---
+- - [3, 6517, 'Ronald', 'Dump', 77]
```
dokshina, line 85: Id `id` a non-indexed field?

dokshina: Usually, ID is a primary index.
Maybe we should show space indexes before all examples.

BTW, you have `surname` field that is non-indexed. And it can show that the different between using first and other parts of composite index (`name` vs `surname`).

doc/select.md
```diff
@@ -0,0 +1,204 @@
...
+res, err = crud.select('customers', {{'>=', 'age', 30}, {'>=', 'id', 3}})
+res.rows
+---
+- - [3, 6517, 'Ronald', 'Dump', 77]
+...
+```
+
+## Pagination
+
+The third (but optional) parameter in ``crud.select`` is array of [``options``](https://github.com/tarantool/crud#select). With this parameter, we can implement pagination.
```
dokshina, line 119: You can specify pagination (and other) options using `opts` argument.

Your description is good, but too long and detailed. `select` API is described in README, this document is used to show how it can be used.

doc/select.md
```diff
@@ -0,0 +1,204 @@
...
+With this request, we got objects behind the objects from the [previous example](https://github.com/crud/doc/select#first-parameter)
+
+### Combine ``first`` and ``after`` 
+
+To use pagination, we have to combine ``after`` and ``first`` parameters. 
+
+**Example:**
+
+```lua
+res, err = crud.select('developers', nil,  { first = 3 })
```
dokshina, line 167: Let's use one space for all examples (jfyi, I like developers more than customers).

doc/select.md
```diff
@@ -0,0 +1,204 @@
...
+  - [6, 8765, 'Artyom', 51]
+...
+res, err = crud.select('developers', nil, { after = res.rows[1], first = -3 })
+res.rows
+---
+- - [1, 7331, 'Alexey', 20]
+  - [2, 899, 'Sergey', 21]
+  - [3, 9661, 'Pavel', 27]
+...
+```
```
dokshina, line 204: For all examples - results of select are shown in the specific order. I want you to notice that. For example, selecting by `age >= x` (if `age` is index) leads to sorting tuples by age, but all your examples shows results sorted by id.

doc/pairs.md
```diff
@@ -0,0 +1,206 @@
...
+
+## ``Use_tomap`` parameter
+
+With ``use_tomap`` flag, you can choose to iterate over objects or over tuples. If ``use_tomap = false``, you will iterate over tuples. This parameter is false by default.
+
+**Example:**
+
+```lua
+tuples = {}
+for _, tuple in crud.pairs('developers', nil, { use_tomap = false, first = 3 }) do
```
dokshina, line 13: I'm not sure we need to show `use_tomap = false` separately.
I would like to show the simplest `pairs` in the first example and then say that you can iterate over objects using `use_tomap` flag.

doc/pairs.md
```diff
@@ -0,0 +1,206 @@
...
+    - 9661
+    - Pavel
+    - Adams
+    - 27
+  - - 4
+    - 501
+    - Mikhail
+    - Liston
+    - 31
+```
```
dokshina, line 118: But there is 4 tuples - it confuses.

doc/pairs.md
```diff
@@ -0,0 +1,206 @@
...
+
+Note that ``crud.pairs``, unlike ``crud.select``, **don't support reverse pagination.**
```
dokshina, line 120: doesn't?

doc/select.md
```diff
@@ -0,0 +1,296 @@
...
+  - [3, 9661, 'Pavel', 'Adams', 27]
+  - [4, 501, 'Mikhail', 'Liston', 31]
+  - [5, 1993, 'Dmitry', 'Jacobi', 16]
+  - [6, 8765, 'Alexey', 'Sidorov', 51]
+...
+```
+
+### Select using index
+
+Let's say we have a ``age`` index. Example below gets a list of ``customers`` over 30 years old.
```
dokshina, line 38: Let's show that tuples are sorted by age in this case (now the order is the same as if tuples are sorted by primary key).

doc/select.md
```diff
@@ -0,0 +1,296 @@
...
+  - {'name': 'surname', 'type': 'string'}
+  - {'name': 'age', 'type': 'number'}
+  rows:
+  - [1, 7331, 'Alexey', 'Adams', 20]
+...
+```
+
+### Select using partial key 
+
+Alternatively, you can use a partial key for a composite index.
```
dokshina, line 79: But if you specify `crud.select('developers', {{'==', 'full_name',  {nil, "Sidorov"}}})` than full scan by pk with filter by `full_name == {nil, "Sidorov"}` is performed.

doc/select.md
```diff
@@ -0,0 +1,296 @@
...
+  - {'name': 'age', 'type': 'number'}
+  rows:
+  - [1, 7331, 'Alexey', 'Adams', 20]
+  - [3, 9661, 'Pavel', 'Adams', 27]
+...
+```
+
+**Note:** in this case full scan is performed.
+
+Note that the search condition for the indexed field must be placed first to avoid a full scan.
```
dokshina, line 121: I think that such recommendations can be placed at the top of the document in the list with a links to related examples.

doc/select.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+### Reverse pagination
+
+Select also supports reverse pagination. To use it, pass a negative value to the ``first`` parameter.
+
+**Example:**
+
+```lua
+res, err = crud.select('developers', nil, { first = 3 })
```
dokshina, line 254: Maybe we can improve comments here

Smth like


```
-- Imagine that user looks at his friends list using very small pages
-- He opens first page 
-- He presses `->` and take next page
-- Then. he wants to return back and presses `<-`
```

doc/pairs.md
```diff
@@ -0,0 +1,205 @@
...
+    - 27
+  - - 4
+    - 501
+    - Mikhail
+    - Liston
+    - 51
+...
+```
+
+## ``Use_tomap`` parameter
```
dokshina, line 42: ``use_tomap`

doc/select.md
```diff
@@ -0,0 +1,301 @@
+# Select examples
+
+## Filtering
+
+``CRUD`` allows to filter tuples by conditions. Each condition can use field name (or number) or index name. The first condition that uses index name is used to iterate over space. If there is no conditions that match index names, full scan is performed. Other conditions are used as additional filters. Search condition for the indexed field must be placed first to avoid a full scan.
+
+**Note:** If you specify ``sharding key`` or ``bucket_id`` select will be performed on single node. Otherwise Map-Reduce over all nodes will be occurred.
```
dokshina, line 7: I think we don't need `` quotes for sharding key.

doc/select.md
```diff
@@ -0,0 +1,301 @@
...
+  - {'name': 'bucked_id', 'type': 'unsigned'}
+  - {'name': 'name', 'type': 'string'}
+  - {'name': 'surname', 'type': 'string'}
+  - {'name': 'age', 'type': 'number'}
+  rows:
+  - [6, 8765, 'Alexey', 'Sidorov', 31]
+  - [4, 501, 'Mikhail', 'Liston', 51]
+...
+```
+
```
dokshina, line 56: Please, add a note that results are sorted by age because first condition is `age` that is index name,

doc/select.md
```diff
@@ -0,0 +1,301 @@
...
+  rows:
+  - [3, 9661, 'Pavel', 'Adams', 27]
+...
+```
+
+## Pagination
+
+[See more](https://github.com/tarantool/crud#select) about ``opts`` parameter.
+
+### ``First`` parameter
```
dokshina, line 163: ``first``

doc/select.md
```diff
@@ -0,0 +1,301 @@
...
+  rows:
+  - [1, 7331, 'Alexey', 'Adams', 20]
+  - [2, 899, 'Sergey', 'Allred', 21]
+  - [3, 9661, 'Pavel', 'Adams', 27]
+...
+```
+
+Thus, we got the first three objects from the ``developers`` space.
+
+### ``After`` parameter
```
dokshina, line 187: ``after``

doc/pairs.md
```diff
@@ -0,0 +1,205 @@
+# Pairs examples
+
+With ``crud.pairs``, you can iterate across a distributed space. The arguments are the same as [``crud.select``](https://github.com/tarantool/crud/docs/select.md), except or the ``use_tomap`` parameter. Below are examples that may help you.
```
olegrok, line 3: Please split this line into several lines. Currently it's too long.

Also `except or` -> `except of`?

doc/pairs.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Getting space
+
+Let's check ``developers`` space contents to make other examples more clear. Just select first 4 values without conditions.
```
olegrok, line 7: `space contents` -> `space content`?

doc/pairs.md
```diff
@@ -0,0 +1,205 @@
...
+  table.insert(tuples, tuple)
+end
+
+tuples
+---
+- - - 1
+    - 7331
+    - Alexey
+    - Adams
+    - 20
```
olegrok, line 23: It would be great to describe fields and schema of space briefly. I see no information above.

doc/pairs.md
```diff
@@ -0,0 +1,205 @@
...
+    - 501
+    - Mikhail
+    - Liston
+    - 51
+...
+```
+
+## ``use_tomap`` parameter
+
+With ``use_tomap`` flag, you can choose to iterate over objects or over tuples. If ``use_tomap = true``, you will iterate over objects. This parameter is false by default.
```
olegrok, line 44: Please split this line into several

doc/pairs.md
```diff
@@ -0,0 +1,205 @@
...
+    bucket_id: 9661
+    name: Pavel
+    surname: Adams
+    age: 27
+...
+```
+
+## Pagination
+
+``crud.pairs``, like [``crud.select``](https://github.com/tarantool/crud/doc/select#pagination), supports pagination. To use it, combine the ``first`` and ``after`` parameters.
```
olegrok, line 76: Please split this line into several.

doc/select.md
```diff
@@ -0,0 +1,303 @@
...
+
+``CRUD`` allows to filter tuples by conditions. Each condition can use field name (or number) or index name. The first condition that uses index name is used to iterate over space. If there is no conditions that match index names, full scan is performed. Other conditions are used as additional filters. Search condition for the indexed field must be placed first to avoid a full scan.
+
+**Note:** If you specify sharding key or ``bucket_id`` select will be performed on single node. Otherwise Map-Reduce over all nodes will be occurred.
+
+Below are examples of filtering data using these conditions. 
+
+### Getting space
+
+Let's check ``developers`` space contents to make other examples more clear. Just select first 6 values without conditions.
```
olegrok, line 13: `space contents` -> `space content`?

doc/select.md
```diff
@@ -0,0 +1,303 @@
...
+...
+```
+
+## Pagination
+
+[See more](https://github.com/tarantool/crud#select) about ``opts`` parameter.
+
+### ``first`` parameter
+
+Using the ``first`` option we will get the first **N** results for the query.
```
olegrok, line 167: maybe `results of query`?

doc/select.md
```diff
@@ -0,0 +1,303 @@
...
+  rows:
+  - [4, 501, 'Mikhail', 'Liston', 51]
+  - [5, 1993, 'Dmitry', 'Jacobi', 16]
+  - [6, 8765, 'Alexey', 'Sidorov', 31]
+...
+```
+
+### Reverse pagination
+
+Select also supports reverse pagination. To use it, pass a negative value to the ``first`` parameter.
```
olegrok, line 253: As I remember you can't use reverse pagination without "after" option. Seems you could add small note about it here.

olegrok (commented):
Thanks for your changes. Consider several comments below.

dokshina (approved):
Please, fix @olegrok comments and then feel free to merge.

Thanks a lot for these examples!

Totktonada:
@mRrvz How bucket IDs are calculated? I can't get those values from permutations of given field values using `require('vshard.hash').<..strcrc32 or mpcrc32..>({<...>}) % <...> + 1`.

mRrvz:
@Totktonada hi! To be honest, I don't remember anymore. I think we need to ask @olegrok or @dokshina.

olegrok:
> @mRrvz How bucket IDs are calculated? I can't get those values from permutations of given field values using `require('vshard.hash').<..strcrc32 or mpcrc32..>({<...>}) % <...> + 1`.

I think it's mpcrc32 with bucket_count = 3000

Totktonada:
@olegrok I tried strcrc32 and mpcrc32 with bucket_count = 10000 for the primary key and different permutations of other fields. bucket_count = 3000 is impossible, because we have bucket_id = 7331. If bucket_count is not 10000, but X*1000, I would anyway see the same last three digits.

PR #111, opened 2021-01-22
Fix using outdated space schema

### LINKED ISSUE #98: [4pt] CRUD uses outdated schema from vshard connections
## Steps to reproduce:

Original issue: see https://github.com/vrogach2020/mdm-sandbox/pull/1

Given the following setup: a Cartridge application + tarantool/crud + tarantool/migrations + integration tests in luatest, an integration test trying to call `crud.insert('test', {...})` fails with error `Failed to insert test, error: Insert: Space "test" doesn't exist`, although the space is initialized from migrations and does exist at the time of call, which is checked beforehand with a code like:
```lua
--wait until migrations are applied
    for _, server in pairs(helper.cluster.servers) do
        --spaces may be created with a slight delay on replicas
        helper.cluster:retrying({ timeout = 1 }, function()
            t.assert_not(server.net_box:eval('return box.space.test == nil'), server.alias)
        end)
    end

    helper.cluster:wait_until_healthy()
```  

CRUD checks the space existence before performing operations on it using internal vshard connections, the simplified code looks like : 

```lua
_, replicaset = next(vshard.router.routeall())
replicaset.master.conn.space[space_name]
```
Investigated by @rosik:
Vshard connection is established before migrations, and until the first insert, there are no other communications. Thus crud uses outdated schema and the request fails.

The suggested W/A is explicitly calling `ping()` for forcing schema update: 
```lua
helper.cluster.main_server.net_box:eval([[
        local vshard = require('vshard')
        for _, r in pairs(vshard.router.routeall()) do
            r.master.conn:ping()
        end
    ]])
```

## Actual result

CRUD may fail to work with spaces if vshard schema in internal connections is outdated, returning a fake "space doesn't exist" error

## Expected result

CRUD doesn't fail to work with spaces when spaces actually exist

## Proposed solution

1. @racktear: do not check for space existence at all.
May lead to errors when attempting to use the space format/metadata.

2. Do not use vshard connections for checking the space existence.
Probably requires the cluster DDL to be implemented

dokshina:
The problem is that if schema is changed on storages, router still uses
outdated schema from vshard connection. It leads to errors like 
"non-existent space" when space is already created. 
The solution is quite simple - all such errors leads to `conn:reload_schema()` 
call and then crud retries. In case of error on the storage-side (e.g. in `*_object` 
functions when object was flattened using outdated space format) space schema
hash is returned. Router compares this hash with hash of schema that was used 
on router-side, and reloads schema if hashes are different.
 `conn:reload_schema()` is called by one fiber at time - if reloading is in progress,
other fibers just wait for it to end.

Closes #98

### DIFF
--- a/crud/common/call.lua
+++ b/crud/common/call.lua
@@ -160,6 +160,8 @@ function call.rw_single(bucket_id, func_name, func_args, options)
             )
         end
 
+        err = errors.wrap(err)
+
         return nil, CallError:new(utils.format_replicaset_error(
              replicaset.uuid, "Function returned an error: %s", err
         ))
--- a/crud/common/const.lua
+++ b/crud/common/const.lua
@@ -1,0 +1,6 @@
+local const = {}
+
+const.RELOAD_RETRIES_NUM = 1
+const.RELOAD_SCHEMA_TIMEOUT = 3 -- 3 seconds
+
+return const
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -1,0 +1,170 @@
+local fiber = require('fiber')
+local msgpack = require('msgpack')
+local digest = require('digest')
+local vshard = require('vshard')
+local errors = require('errors')
+local log = require('log')
+
+local ReloadSchemaError = errors.new_class('ReloadSchemaError',  {capture_stack = false})
+
+local const = require('crud.common.const')
+
+local schema = {}
+
+local function table_len(t)
+    local len = 0
+    for _ in pairs(t) do
+        len = len + 1
+    end
+    return len
+end
+
+local function call_reload_schema_on_replicaset(replicaset, channel)
+    replicaset.master.conn:reload_schema()
+    channel:put(true)
+end
+
+local function call_reload_schema(replicasets)
+    local replicasets_num = table_len(replicasets)
+    local channel = fiber.channel(replicasets_num)
+
+    local fibers = {}
+    for _, replicaset in pairs(replicasets) do
+        local f = fiber.new(call_reload_schema_on_replicaset, replicaset, channel)
+        table.insert(fibers, f)
+    end
+
+    for _ = 1,replicasets_num do
+        if channel:get(const.RELOAD_SCHEMA_TIMEOUT) == nil then
+            for _, f in ipairs(fibers) do
+                if fiber:status() ~= 'dead' then
+                    f:cancel()
+                end
+            end
+            return nil, ReloadSchemaError:new("Reloading schema timed out")
+        end
+    end
+
+    return true
+end
+
+local reload_in_progress = false
+local reload_schema_cond = fiber.cond()
+
+local function reload_schema(replicasets)
+    if reload_in_progress then
+        if not reload_schema_cond:wait(const.RELOAD_SCHEMA_TIMEOUT) then
+            return nil, ReloadSchemaError:new('Waiting for schema to be reloaded is timed out')
+        end
+    else
+        reload_in_progress = true
+
+        local ok, err = call_reload_schema(replicasets)
+        if not ok then
+            return nil, err
+        end
+
+        reload_schema_cond:broadcast()
+        reload_in_progress = false
+    end
+
+    return true
+end
+
+-- schema.wrap_func_reload calls func with specified arguments.
+-- func should return `res, err, need_reload`
+-- If function returned error and `need_reload` is true,
+-- then schema is reloaded and one more attempt is performed
+-- (but no more than RELOAD_RETRIES_NUM).
+-- This wrapper is used for functions that can fail if router uses outdated
+-- space schema. In case of such errors these functions returns `need_reload`
+-- for schema-dependent errors.
+function schema.wrap_func_reload(func, ...)
+    local i = 0
+
+    local res, err, need_reload
+    while true do
+        res, err, need_reload = func(...)
+
+        if err == nil or not need_reload then
+            break
+        end
+
+        local ok, reload_schema_err = reload_schema(vshard.router.routeall())
+        if not ok then
+            log.warn("Failed to reload schema: %s", reload_schema_err)
+            break
+        end
+
+        i = i + 1
+        if i > const.RELOAD_RETRIES_NUM then
+            break
+        end
+    end
+
+    return res, err
+end
+
+local function get_space_schema_hash(space)
+    if space == nil then
+        return ''
+    end
+
+    local indexes_info = {}
+    for i = 0, table.maxn(space.index) do
+        local index = space.index[i]
+        if index ~= nil then
+            indexes_info[i] = {
+                unique = index.unique,
+                parts = index.parts,
+                id = index.id,
+                type = index.type,
+                name = index.name,
+                path = index.path,
+            }
+        end
+    end
+
+    local space_info = {
+        format = space:format(),
+        indexes = indexes_info,
+    }
+
+    return digest.murmur(msgpack.encode(space_info))
+end
+
+-- schema.wrap_box_space_func_result pcalls some box.space function
+-- and returns its result as a table
+-- `{res = ..., err = ..., space_schema_hash = ...}`
+-- space_schema_hash is computed if function failed and
+-- `add_space_schema_hash` is true
+function schema.wrap_box_space_func_result(add_space_schema_hash, space, func_name, ...)
+    local result = {}
+
+    local ok, func_res = pcall(space[func_name], space, ...)
+    if not ok then
+        result.err = func_res
+        if add_space_schema_hash then
+            result.space_schema_hash = get_space_schema_hash(space)
+        end
+    else
+        result.res = func_res
+    end
+
+    return result
+end
+
+-- schema.result_needs_reload checks that schema reload can
+-- be helpful to avoid storage error.
+-- It checks if space_schema_hash returned by storage
+-- is the same as hash of space used on router.
+-- Note, that storage returns `space_schema_hash = nil`
+-- if reloading space format can't avoid the error.
+function schema.result_needs_reload(space, result)
+    if result.space_schema_hash == nil then
+        return false
+    end
+    return result.space_schema_hash ~= get_space_schema_hash(space)
+end
+
+return schema
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -1,12 +1,15 @@
 local errors = require('errors')
 local ffi = require('ffi')
+local vshard = require('vshard')
 
+local schema = require('crud.common.schema')
 local dev_checks = require('crud.common.dev_checks')
 
 local FlattenError = errors.new_class("FlattenError", {capture_stack = false})
 local UnflattenError = errors.new_class("UnflattenError", {capture_stack = false})
 local ParseOperationsError = errors.new_class('ParseOperationsError',  {capture_stack = false})
 local ShardingError = errors.new_class('ShardingError',  {capture_stack = false})
+local GetSpaceFormatError = errors.new_class('GetSpaceFormatError',  {capture_stack = false})
 
 local utils = {}
 
@@ -34,7 +37,19 @@ end
 function utils.get_space(space_name, replicasets)
     local replicaset = select(2, next(replicasets))
     local space = replicaset.master.conn.space[space_name]
+
     return space
+end
+
+function utils.get_space_format(space_name, replicasets)
+    local space = utils.get_space(space_name, replicasets)
+    if space == nil then
+        return nil, GetSpaceFormatError:new("Space %q doesn't exist", space_name)
+    end
+
+    local space_format = space:format()
+
+    return space_format
 end
 
 local system_fields = { bucket_id = true }
@@ -44,20 +59,30 @@ function utils.flatten(object, space_format, bucket_id)
 
     local tuple = {}
 
-    for fieldno, field_format in ipairs(space_format) do
-        local value = object[field_format.name]
+    local fieldnames = {}
 
-        if not system_fields[field_format.name] then
+    for fieldno, field_format in ipairs(space_format) do
+        local fieldname = field_format.name
+        local value = object[fieldname]
+
+        if not system_fields[fieldname] then
             if not field_format.is_nullable and value == nil then
-                return nil, FlattenError:new("Field %q isn't nullable", field_format.name)
+                return nil, FlattenError:new("Field %q isn't nullable", fieldname)
             end
         end
 
-        if bucket_id ~= nil and field_format.name == 'bucket_id' then
+        if bucket_id ~= nil and fieldname == 'bucket_id' then
             value = bucket_id
         end
 
         tuple[fieldno] = value
+        fieldnames[fieldname] = true
+    end
+
+    for fieldname in pairs(object) do
+        if not fieldnames[fieldname] then
+            return nil, FlattenError:new("Unknown field %q is specified", fieldname)
+        end
     end
 
     return tuple
@@ -124,7 +149,7 @@ local function determine_enabled_features()
     enabled_tarantool_features.uuids = major >= 2 and (minor > 4 or minor == 4 and patch >= 1)
 end
 
-local function tarantool_supports_fieldpaths()
+function utils.tarantool_supports_fieldpaths()
     if enabled_tarantool_features.fieldpaths == nil then
         determine_enabled_features()
     end
@@ -141,7 +166,7 @@ function utils.tarantool_supports_uuids()
 end
 
 function utils.convert_operations(user_operations, space_format)
-    if tarantool_supports_fieldpaths() then
+    if utils.tarantool_supports_fieldpaths() then
         return user_operations
     end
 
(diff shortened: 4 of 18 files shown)


crud/common/utils.lua
```diff
@@ -33,10 +37,36 @@ end
 
 function utils.get_space(space_name, replicasets)
     local replicaset = select(2, next(replicasets))
-    local space = replicaset.master.conn.space[space_name]
+
+    local master = replicaset.master
+    if not master.conn:ping({timeout = ping_timeout}) then
```
akudiyar, line 42: So, we'll do an extra PING call on each crud operation?
Although it is a working solution, we need to measure the performance impact. What about pinging once, saving the schema internally and update it when the schema version changes, as it is done inside the connection?
Also, it is possible to fetch schema from replicas too. Probably it may be faster and save time for the master a bit.

knazarov:
I don't believe that issuing ping requests on every operation is a good measure. Instead, we can do the same thing regular box operations do: send our schema ID along with the call and if they diverge with what is currently on the storage, return a special error and the router will then retry.

olegrok:
> send our schema ID along with the call and if they diverge with what is currently on the storage, return a special error and the router will then retry.

That's still not a general purpose solution. There are some examples when schema could be dynamic and schema will be changed quite often - this will lead to continuous retries.
Moreover you shouldn't send schema_id manually because in case of schema mismatch Tarantool fetches it automatically. Seems your suggestion could be simplified to

```lua
while true do
    local schema_id = conn.schema_version
    local data, err = conn:call(...)
    if err ~= nil and conn.schema_version == schema_id then
        return nil, err
    elseif data ~= nil then
        return data
    end
end
```

I don't like it. I could update some spaces that don't related to my queries but it will be cause some retries.
May be it's rare cases but anyway we shouldn't neglect them.
Seems that the most appropriate solution is support "DDL". I suggest to support it optionally I don't want to impose this on the user.
I don't see any reliable solutions until https://github.com/tarantool/tarantool/issues/802 won't be implemented.

Totktonada:
Ugh… I was sure that `net.box` sends `schema_id` in a request, performs retrying in case of `E_WRONG_SCHEMA_VERSION` error and so guarantees that a request is made against a schema that is not older than one that were actual at the moment when a `net.box` method was called. That is not so. I'll file an issue against `net.box` later (I briefly described the problem in Russian [here](https://t.me/tarantoolru/165945)).

So calling `<connection>:reload_schema()` (it is just wrapper for `:ping()`) when a space does not exist is not enough. And, in any circumstances, we should not issue a ping request before any other request: it doubles latency.

However we still can send a request without the client side checking of a space presence and it'll be the correct solution. You can even use the local `net.box`'s schema after receiving a response (to form a response metadata), because the response is returned only after the schema fetching (if it appears to be obsoleted). Okay, the schema is not guaranteed to correspond *the same* `schema_id`, but is at least not older that one against which the request was made. The only way to achieve the guarantee of *the exactly same* `schema_id` of a response and a local schema (and use current `net.box` implementation) is to send the metadata from the server (which is costly).

Sure, there are cases, when you need to use some schema information to produce a request, but I don't see such cases here.

crud/common/utils.lua
```diff
@@ -226,4 +253,29 @@ function utils.is_uuid(value)
...
 end
 
+function utils.format_result(rows, space)
+    return {
+        metadata = table.copy(space:format()),
+        rows = rows,
+    }
+end
+
+local function flatten_space_obj(space_name, obj)
```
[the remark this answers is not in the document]
dokshina: `flatten_obj_by_space_name(obj, space_name)`?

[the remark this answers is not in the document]
olegrok: It seems to be more clear

Why not just `flatten_object(space_name, obj)`? It will work as `do_something(type_name, object)`

crud/update.lua
```diff
@@ -21,14 +22,61 @@ local function update_on_storage(space_name, key, operations)
...
+    local space = utils.get_space(space_name, vshard.router.routeall())
+    if space == nil then
+        return nil, UpdateError:new("Space %q doesn't exist", space_name), true
+    end
+
+    local space_format = space:format()
+
+    if box.tuple.is(key) then
+        key = key:totable()
+    end
```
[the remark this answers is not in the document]
dokshina:
```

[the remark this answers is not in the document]
Comment (olegrok) [reply crud/update.lua:52]: Yes, my bad, sorry for chores

File: crud/common/schema.lua [hunk 4]
```diff
@@ -0,0 +1,107 @@
...
+        master.conn:reload_schema()
+    end
+end
+
+local reload_in_progress = false
+local reload_schema_cond = fiber.cond()
+
+local function reload_schema(replicasets)
+    if reload_in_progress then
+        reload_schema_cond:wait()
```
olegrok, line 22: It would be great to add some timeout here. It's easy to get deadlock here if something occasionally will throw or reload will hang for some reasons (e.g. network problem)

crud/common/schema.lua
```diff
@@ -0,0 +1,107 @@
...
+local const = require('crud.common.const')
+
+local schema = {}
+
+local function call_reload_schema(replicasets)
+    for _, replicaset in pairs(replicasets) do
+        local master = replicaset.master
+        master.conn:reload_schema()
+    end
+end
```
olegrok, line 15: reload_schema calls ping under the hood. It's not async. Seems you could perform such operation in several fibers.

crud/common/schema.lua
```diff
@@ -0,0 +1,107 @@
...
+            }
+        end
+    end
+
+    local space_info = {
+        format = space:format(),
+        indexes = indexes_info,
+    }
+
+    return digest.md5(msgpack.encode(space_info))
```
olegrok, line 81: md5 is cryptographic hash function. Please use something more lightweight.

crud/common/call.lua
```diff
@@ -160,6 +160,8 @@ function call.rw_single(bucket_id, func_name, func_args, options)
             )
         end
 
+        err = errors.wrap(err)
```
olegrok, line 163: Is it really needed?

crud/common/schema.lua
```diff
@@ -0,0 +1,107 @@
...
+        end
+    end
+
+    return res, err
+end
+
+local function get_space_schema_hash(space)
+    if space == nil then
+        return ''
+    end
```
olegrok, line 60: Are there some cases when `space == nil`? Looks like a mistake.

crud/common/schema.lua
```diff
@@ -0,0 +1,107 @@
...
+    for i = 0, table.maxn(space.index) do
+        local index = space.index[i]
+        if index ~= nil then
+            indexes_info[i] = {
+                unique = index.unique,
+                parts = index.parts,
+                id = index.id,
+                type = index.type,
+                name = index.name,
+            }
```
olegrok, line 72: Also there is a "path" option.

crud/common/schema.lua
```diff
@@ -0,0 +1,107 @@
...
+
+    local space_info = {
+        format = space:format(),
+        indexes = indexes_info,
+    }
+
+    return digest.md5(msgpack.encode(space_info))
+end
+
+function schema.wrap_box_space_func_result(add_space_schema_hash, space, func, ...)
```
olegrok, line 84: I assume we need to describe why every function in this module are needed in comment. Otherwise it's your personal secret knowledge.

crud/delete.lua
```diff
@@ -21,14 +22,55 @@ local function delete_on_storage(space_name, key)
         return nil, DeleteError:new("Space %q doesn't exist", space_name)
     end
 
-    local tuple = space:delete(key)
-    return tuple
+    return schema.wrap_box_space_func_result(false, space, space.delete, space, key)
```
olegrok, line 25: Why "false"? Please describe for all places where you use it.

crud/common/call.lua
dokshina: Yes, error is a table with overloaded `tostring` fuction. We need to reconstruct error received from connection to correctly show it.

crud/common/schema.lua
dokshina: Yes, there is not such cases, but this is just a function that computes space format hash and it knows nothing about this cases. Without this check we will got error on using `space.index` sometimes.

crud/common/schema.lua
dokshina: I don't understand what do you mean =(

crud/delete.lua
dokshina: `false`? What?

olegrok: What? Describe why did you pass "false" as the first argument here

crud/common/call.lua
olegrok: But why did it work ok before?

crud/common/schema.lua
olegrok: > Yes, there are no such cases

You need a put assert/error here

crud/common/call.lua
dokshina: It didn't =)

crud/common/schema.lua
olegrok: Currently you don't understand me. But without good comments you will not understand you in several weeks/months. And nobody won't understand it.

crud/common/call.lua
dokshina: Schema wasn't changed in tests before this patch, so all `Space doesn't exists` errors were returned directly by router.

crud/common/schema.lua
dokshina: In fact, it's a valid case - when space doesn't exist. Why do we need error here?

crud/delete.lua
dokshina: I think that it looks bad now (I mean passing this flag as a first argument).
I can add `schema.wrap_box_space_func_result_no_schema_hash` for such places, but it will require a comment anyway (to describe a logic "why it don't need space hash").

crud/common/schema.lua
olegrok: If it's ok, why do you return empty string but not e.g. nil?

dokshina: But what's the difference?)

crud/common/schema.lua
dokshina: `digest.murmur` is used now.

crud/common/schema.lua
olegrok: String even empty is a content that could be useful for some cases/reasons. `nil` is absence of anything. Just value that says you "I don't have nothing useful".

dokshina: See `schema.result_needs_reload` function. `nil` returned by storage means "I didn't returned any hash, you have nothing to compare with", but `""` means "My space schema is that: space doesn't exist".

crud/common/schema.lua
dokshina: Yes, I didn't understand that you mean adding comments for functions, sorry, that's my fault.

Anyway, I've added comments.

crud/delete.lua
dokshina: Added comments to all calls.

I've also changed `schema.wrap_box_space_func_result` signature to avoid passing `space, space:<func>, space`.

crud/common/const.lua
```diff
@@ -0,0 +1,6 @@
+local const = {}
+
+const.RELOAD_RETRIES_NUM = 1
+const.RELOAD_SCHEMA_TIMEOUT = 30 -- half of minute
```
olegrok, line 4: Even with 1k rps you will create 30k fibers. It's too huge amount. I think you need to decrease to 3-5 seconds timeout.

crud/common/schema.lua
```diff
@@ -0,0 +1,168 @@
...
+    channel:put(true)
+end
+
+local function call_reload_schema(replicasets)
+    local replicasets_num = table_len(replicasets)
+    local channel = fiber.channel(replicasets_num)
+
+    local fibers = {}
+    for _, replicaset in pairs(replicasets) do
+        local fiber = fiber.new(call_reload_schema_on_replicaset, replicaset, channel)
```
olegrok, line 33: Here you shadow upvalue "fiber". I think you could eliminate it.

crud/common/schema.lua
```diff
@@ -0,0 +1,168 @@
...
+        table.insert(fibers, fiber)
+    end
+
+    for _ = 1,replicasets_num do
+        if channel:get(const.RELOAD_SCHEMA_TIMEOUT ) == nil then
```
olegrok, line 38: `TIMEOUT )` -> `TIMEOUT)`

crud/delete.lua
```diff
@@ -21,14 +22,57 @@ local function delete_on_storage(space_name, key)
...
+    if storage_result.err ~= nil then
+        return nil, DeleteError:new("Failed to delete: %s", storage_result.err)
+    end
+
+    local tuple = storage_result.res
+
+    return {
+        metadata = table.copy(space:format()),
+        rows = {tuple},
+    }
```
olegrok, line 73: Finally, I don't understand why did you move "format_result" into separate function if you don't actually use it.

crud/get.lua
```diff
@@ -72,23 +56,59 @@ function get.call(space_name, key, opts)
...
+        return nil, GetError:new("Failed to get: %s", storage_result.err)
+    end
+
+    local tuple = storage_result.res
+
     -- protect against box.NULL
-    if result == nil then
-       result = nil
+    if tuple == nil then
+        tuple = nil
```
olegrok, line 75: `_tuple` -> `tuple`

crud/delete.lua
```diff
@@ -21,14 +22,57 @@ local function delete_on_storage(space_name, key)
...
 end
 
 function delete.init()
    _G._crud.delete_on_storage = delete_on_storage
 end
 
+-- returns result, err, need_reload
+-- need_reload indicates if reloading schema could help
+-- see crud.common.schema.wrap_func_reload()
+local function call_delete(space_name, key, opts)
```
olegrok, line 37: I suggest to use e.g. "*_on_router". I really don't like that use place router and storage function in single file but it's another question. "delete_on_storage" says that you perform some logic on "storage-side". "call_delete" says that you call some abstract delete - storage? router? Ok, router, but I need to read source code to understand it.

crud/update.lua
```diff
@@ -21,14 +22,63 @@ local function update_on_storage(space_name, key, operations)
...
+
+    local operations, err = utils.convert_operations(user_operations, space_format)
+    if err ~= nil then
+        return nil, UpdateError:new("Wrong operations are specified: %s", err), true
+    end
+
+    local bucket_id = sharding.key_get_bucket_id(key, opts.bucket_id)
+    local storage_result, err = call.rw_single(
+        bucket_id, UPDATE_FUNC_NAME, {space_name, key, operations},
+        {timeout=opts.timeout})
```
olegrok, line 64: Since you anyway change this piece of code, you have a great chance to add spaces around `=` as for other places.

dokshina: Who you are? Oleg always suggests doing such things in a separated patch.

crud/common/utils.lua
```diff
@@ -33,31 +36,55 @@ end
...
 
         tuple[fieldno] = value
+        fieldnames[fieldname] = true
+    end
+
+    for fieldname in pairs(object) do
+        if not fieldnames[fieldname] then
+            return nil, FlattenError:new("Unknown field %q is specified", fieldname)
+        end
     end
```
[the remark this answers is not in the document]
dokshina: It's covered by outdated schema test, but I will add unit test for this case.

crud/common/schema.lua
```diff
@@ -0,0 +1,168 @@
...
+end
+
+-- schema.result_needs_reload checks that schema reload can
+-- be helpful to avoid storage error.
+-- It checks if space_schema_hash returned by storage
+-- is the same as hash of space used on router.
+-- Note, that storage returns `space_schema_hash = nil`
+-- if reloading space format can't avoid the error.
+function schema.result_needs_reload(space, result)
+    if result.space_schema_hash == nil then
```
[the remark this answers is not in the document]
dokshina: No, `nil` hash means that reloading can't avoid the error.
Empty hash could mean that space doesn't exist on storage (but it can still exist on router). Now this case isn't actual, but in the future when we start to create spaces using crud it could be helpful.

crud/delete.lua
dokshina: Oh shit, seems I've missed this changes =(

crud/get.lua
dokshina: No, indentation was wrong, but now it's OK.

crud/common/utils.lua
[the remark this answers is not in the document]
dokshina: Added a test.

crud/common/schema.lua
```diff
@@ -0,0 +1,170 @@
...
+    while true do
+        res, err, need_reload = func(...)
+
+        if err == nil or not need_reload then
+            break
+        end
+
+        local ok, reload_schema_err = reload_schema(vshard.router.routeall())
+        if not ok then
+            log.debug("Failed to reload schema: %s", reload_schema_err)
```
olegrok, line 95: maybe verbose?

olegrok (approved):
"In CRUD we trust" say them.

Currently it's not so.

crud/common/schema.lua
dokshina: I suggest `warn`.

PR #118, opened 2021-02-11
Support partial results

### LINKED ISSUE #116: [2pt] Simple operations: Partial results
The first part of #53 

In this part we will support partial results for all operations except select/pairs.
This is much more simple for "1-tuple" operations than for "many-tuples" operations when we need to merge results on router, so I sure that it should be done in two patches.

AnaNek:
Closes #116

### DIFF
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -8,6 +8,7 @@ local log = require('log')
 local ReloadSchemaError = errors.new_class('ReloadSchemaError',  {capture_stack = false})
 
 local const = require('crud.common.const')
+local dev_checks = require('crud.common.dev_checks')
 
 local schema = {}
 
@@ -133,22 +134,18 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
-local function extract_fields(tuple, fields)
-    local extracted_fields = {}
-
-    for i, field in ipairs(fields) do
-        extracted_fields[i] = tuple[field]
-    end
-
-    return extracted_fields
-end
-
-local function get_partial_result(tuple, fields)
+local function filter_result_fields(tuple, fields)
     if fields == nil or tuple == nil then
         return tuple
-    else
-        return extract_fields(tuple, fields)
     end
+
+    local result = {}
+
+    for i, field in ipairs(fields) do
+        result[i] = tuple[field]
+    end
+
+    return result
 end
 
 -- schema.wrap_box_space_func_result pcalls some box.space function
@@ -157,6 +154,8 @@ end
 -- space_schema_hash is computed if function failed and
 -- `add_space_schema_hash` is true
 function schema.wrap_box_space_func_result(space, func_name, args, opts)
+    dev_checks('table', 'string', 'table', 'table')
+
     local result = {}
 
     opts = opts or {}
@@ -168,7 +167,7 @@ function schema.wrap_box_space_func_result(space, func_name, args, opts)
             result.space_schema_hash = get_space_schema_hash(space)
         end
     else
-        result.res = get_partial_result(func_res, opts.fields)
+        result.res = filter_result_fields(func_res, opts.fields)
     end
 
     return result
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -135,18 +135,18 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
-local function filter_result_fields(tuple, fields)
-    if fields == nil or tuple == nil then
+local function filter_result_fields(tuple, field_names)
+    if field_names == nil or tuple == nil then
         return tuple
     end
 
     local result = {}
 
-    for i, field in ipairs(fields) do
-        result[i] = tuple[field]
+    for i, field_name in ipairs(field_names) do
+        result[i] = tuple[field_name]
         if result[i] == nil then
             return nil, FilterFieldsError:new(
-                    'Space format doesn\'t contain field named %q', field
+                    'Space format doesn\'t contain field named %q', field_name
             )
         end
     end
@@ -174,7 +174,7 @@ function schema.wrap_box_space_func_result(space, func_name, args, opts)
             result.space_schema_hash = get_space_schema_hash(space)
         end
     else
-        result.res, err = filter_result_fields(func_res, opts.fields)
+        result.res, err = filter_result_fields(func_res, opts.field_names)
         if err ~= nil then
             return nil, err
         end
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -6,6 +6,7 @@ local errors = require('errors')
 local log = require('log')
 
 local ReloadSchemaError = errors.new_class('ReloadSchemaError',  {capture_stack = false})
+local FilterFieldsError = errors.new_class('FilterFieldsError',  {capture_stack = false})
 
 local const = require('crud.common.const')
 local dev_checks = require('crud.common.dev_checks')
@@ -143,6 +144,11 @@ local function filter_result_fields(tuple, fields)
 
     for i, field in ipairs(fields) do
         result[i] = tuple[field]
+        if result[i] == nil then
+            return nil, FilterFieldsError:new(
+                    'Space format doesn\'t contain field named %q', field
+            )
+        end
     end
 
     return result
@@ -157,6 +163,7 @@ function schema.wrap_box_space_func_result(space, func_name, args, opts)
     dev_checks('table', 'string', 'table', 'table')
 
     local result = {}
+    local err
 
     opts = opts or {}
 
@@ -167,7 +174,10 @@ function schema.wrap_box_space_func_result(space, func_name, args, opts)
             result.space_schema_hash = get_space_schema_hash(space)
         end
     else
-        result.res = filter_result_fields(func_res, opts.fields)
+        result.res, err = filter_result_fields(func_res, opts.fields)
+        if err ~= nil then
+            return nil, err
+        end
     end
 
     return result
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -133,23 +133,24 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
-local function get_partial_result(func_get_res, fields)
-    local result = {}
+local function extract_fields(tuple, fields)
+    local extracted_fields = {}
 
-    result.err = func_get_res.err
-    if func_get_res.res ~= nil then
-        if fields ~= nil then
-            result.res = {}
-            for i, field in ipairs(fields) do
-                result.res[i] = func_get_res.res[field]
-            end
-        else
-            result.res = func_get_res.res
-        end
+    for i, field in ipairs(fields) do
+        extracted_fields[i] = tuple[field]
     end
 
-    return result
+    return extracted_fields
 end
+
+local function get_partial_result(tuple, fields)
+    if fields == nil or tuple == nil then
+        return tuple
+    else
+        return extract_fields(tuple, fields)
+    end
+end
+
 -- schema.wrap_box_space_func_result pcalls some box.space function
 -- and returns its result as a table
 -- `{res = ..., err = ..., space_schema_hash = ...}`
@@ -167,10 +168,10 @@ function schema.wrap_box_space_func_result(space, func_name, args, opts)
             result.space_schema_hash = get_space_schema_hash(space)
         end
     else
-        result.res = func_res
+        result.res = get_partial_result(func_res, opts.fields)
     end
 
-    return get_partial_result(result, opts.fields)
+    return result
 end
 
 -- schema.result_needs_reload checks that schema reload can
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -133,25 +133,44 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
+local function get_partial_result(func_get_res, fields)
+    local result = {}
+
+    result.err = func_get_res.err
+    if func_get_res.res ~= nil then
+        if fields ~= nil then
+            result.res = {}
+            for i, field in ipairs(fields) do
+                result.res[i] = func_get_res.res[field]
+            end
+        else
+            result.res = func_get_res.res
+        end
+    end
+
+    return result
+end
 -- schema.wrap_box_space_func_result pcalls some box.space function
 -- and returns its result as a table
 -- `{res = ..., err = ..., space_schema_hash = ...}`
 -- space_schema_hash is computed if function failed and
 -- `add_space_schema_hash` is true
-function schema.wrap_box_space_func_result(add_space_schema_hash, space, func_name, ...)
+function schema.wrap_box_space_func_result(space, func_name, args, opts)
     local result = {}
 
-    local ok, func_res = pcall(space[func_name], space, ...)
+    opts = opts or {}
+
+    local ok, func_res = pcall(space[func_name], space, unpack(args))
     if not ok then
         result.err = func_res
-        if add_space_schema_hash then
+        if opts.add_space_schema_hash then
             result.space_schema_hash = get_space_schema_hash(space)
         end
     else
         result.res = func_res
     end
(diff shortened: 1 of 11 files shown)


crud/get.lua
```diff
@@ -24,7 +81,9 @@ local function get_on_storage(space_name, key)
 
     -- add_space_schema_hash is false because
     -- reloading space format on router can't avoid get error on storage
-    return schema.wrap_box_space_func_result(false, space, 'get', key)
```
dokshina, line 27: What do you think about delegating `fields` processing to `schema.wrap_box_space_func_result`?

You can change this function sugnature.

Now it is

`function schema.wrap_box_space_func_result(add_space_schema_hash, space, func_name, ...)`

If it become 

`function schema.wrap_box_space_func_result(space, func_name, args, opts)`, 
where `opts` is `{add_space_schema_hash = '?boolean', fields = '?table'}`, it can be used this way:

```lua
return schema.wrap_box_space_func_result(space, 'get', {key}, {
    add_space_schema_hash = false, 
   fields = fields,
})
```

crud/get.lua
```diff
@@ -77,7 +137,9 @@ local function call_get_on_router(space_name, key, opts)
         tuple = nil
     end
 
-    return utils.format_result({tuple}, space)
+    local formatted_result = utils.format_result({tuple}, space)
```
dokshina, line 140: What if `utils.format_result` accept `fields` argument?

`return utils.format_result({tuple}, space, fields)`

In this case you can just change this function call everywhere you need.

crud/get.lua
```diff
@@ -14,8 +14,65 @@ local get = {}
...
+    if func_get_res.res ~= nil then
+        if fields ~= nil then
+            result.res = {}
+            for i, field in ipairs(fields) do
+                result.res[i] = func_get_res.res[field]
+            end
+        else
+            result.res = func_get_res.res
+        end
+    end
```
olegrok, line 69: It confuses me a bit. I suggest to pass "tuple" and "fields" as arguments.

If fields are nil then function returns tuple immediately. Otherwise you can extract fields.

I'd recommend to untie your feature as max as possible from schema manipulations and retries logic.

crud/get.lua
```diff
@@ -14,8 +14,65 @@ local get = {}
...
-local function get_on_storage(space_name, key)
-    dev_checks('string', '?')
+local function get_field_metadata(full_metadata, field)
+    dev_checks('table', 'string')
+
+    for _, tuple in ipairs(full_metadata) do
+        if tuple['name'] == field then
+            return tuple
+        end
+    end
```
olegrok, line 24: This code looks a bit suboptimal. I suggest to cache field map.

Small snippet:

```lua
local cache = setmetabale({}, {__mode = 'k'}) -- https://www.lua.org/pil/17.html

...
local metadata_cache = cache[full_metadata]
if metadata_cache ~= then
    return metadata_cache[field]
end

cache[full_metadata] = {}
for i, field in ipairs(full_metadata) do 
    cache[full_metadata][field.name] = field
end

...
return cache[full_metadata][field]
```

olegrok (commented):
Thanks for your patch! Consider several comments below.

test/integration/simple_operations_test.lua
```diff
@@ -391,3 +391,233 @@ pgroup:add('test_upsert', function(g)
...
+        {name = 'id', type = 'unsigned'},
+        {name = 'bucket_id', type = 'unsigned'},
+        {name = 'name', type = 'string'},
+        {name = 'age', type = 'number'},
+    })
+    local objects = crud.unflatten_rows(result.rows, result.metadata)
+    t.assert_equals(objects, {{id = 1, name = 'Elizabeth', age = 24, bucket_id = 477}})
+
+    -- get
+    local result, err = g.cluster.main_server.net_box:call('crud.get', {'customers', 1, {fields={'id', 'name'}}})
```
mRrvz, line 411: Please, add space here (and below): 
```lua
fields = {'id', 'name'}
```

test/integration/simple_operations_test.lua
```diff
@@ -391,3 +391,233 @@ pgroup:add('test_upsert', function(g)
...
+        {fields={'id', 'age'}}
+    })
+
+    t.assert_equals(#result.rows, 0)
+    t.assert_equals(result.metadata, {
+        {name = 'id', type = 'unsigned'},
+        {name = 'age', type = 'number'},
+    })
+    t.assert_equals(err, nil)
+end)
```
mRrvz, line 623: Please, add a newline

test/integration/simple_operations_test.lua
```diff
@@ -391,3 +391,233 @@ pgroup:add('test_upsert', function(g)
...
+        'customers', {1, box.NULL, 'Elizabeth', 23}, {{'+', 'age', 1},},
+        {fields={'id', 'age'}}
+    })
+
+    t.assert_equals(#result.rows, 0)
+    t.assert_equals(result.metadata, {
+        {name = 'id', type = 'unsigned'},
+        {name = 'age', type = 'number'},
+    })
+    t.assert_equals(err, nil)
```
mRrvz, line 592: I think it's more logical to check for an error from beginning:

```lua
    t.assert_equals(err, nil)
    t.assert_equals(#result.rows, 0)
    t.assert_equals(result.metadata, {
        {name = 'age', type = 'number'},
    })
```

test/integration/simple_operations_test.lua
```diff
@@ -391,3 +391,233 @@ pgroup:add('test_upsert', function(g)
...
+
+    t.assert_equals(err, nil)
+    t.assert_equals(result.metadata, {
+        {name = 'id', type = 'unsigned'},
+        {name = 'bucket_id', type = 'unsigned'},
+        {name = 'name', type = 'string'},
+        {name = 'age', type = 'number'},
+    })
+    local objects = crud.unflatten_rows(result.rows, result.metadata)
+    t.assert_equals(objects, {{id = 1, name = 'Elizabeth', age = 24, bucket_id = 477}})
```
mRrvz, line 408: Why 477?

crud/common/schema.lua
```diff
@@ -133,22 +133,42 @@ local function get_space_schema_hash(space)
...
+
+    return extracted_fields
+end
+
+local function get_partial_result(tuple, fields)
+    if fields == nil or tuple == nil then
+        return tuple
+    else
+        return extract_fields(tuple, fields)
+    end
```
mRrvz, line 151: I think this ``else`` construction can be deleted

crud/common/schema.lua
```diff
@@ -133,22 +133,42 @@ local function get_space_schema_hash(space)
...
+end
+
 -- schema.wrap_box_space_func_result pcalls some box.space function
 -- and returns its result as a table
 -- `{res = ..., err = ..., space_schema_hash = ...}`
 -- space_schema_hash is computed if function failed and
 -- `add_space_schema_hash` is true
-function schema.wrap_box_space_func_result(add_space_schema_hash, space, func_name, ...)
+function schema.wrap_box_space_func_result(space, func_name, args, opts)
```
dokshina, line 159: Let's add `dev_checks` here.

crud/common/schema.lua
```diff
@@ -133,22 +133,42 @@ local function get_space_schema_hash(space)
...
+        extracted_fields[i] = tuple[field]
+    end
+
+    return extracted_fields
+end
+
+local function get_partial_result(tuple, fields)
+    if fields == nil or tuple == nil then
+        return tuple
+    else
```
dokshina, line 149: `else` block is redundant after return.

crud/common/utils.lua
```diff
@@ -251,11 +253,46 @@ function utils.is_uuid(value)
...
+local function get_field_metadata(full_metadata, field)
+    dev_checks('table', 'string')
+
+    local metadata = metadata_cache[full_metadata]
+    if metadata ~= nil then
+        return metadata[field]
+    end
+
+    metadata_cache[full_metadata] = {}
+    for _, tuple in ipairs(full_metadata) do
```
dokshina, line 265: Is it `tuple`? Looks like `field`.

test/integration/simple_operations_test.lua
dokshina, line 623: Please, add a newline.

crud/common/utils.lua
```diff
@@ -251,11 +253,46 @@ function utils.is_uuid(value)
...
+function utils.format_result(rows, space, fields)
+    local result = {}
+    local metadata = table.copy(space:format())
+    result.rows = rows
+
+    if fields ~= nil then
+        result.metadata = format_metadata(metadata, fields)
+    else
+        result.metadata = metadata
+    end
```
dokshina, line 293: I suggest renaming `metadata` to `space_format`. In fact, in becomes metadata only when we return it. But inside the function it's much more simple to think about `space format` than think about some `metadata`.

```lua
local space_format = table.copy(space:format())
result_fields_format = filter_format_fields(space_format)

result.metadata = result_fields_format
```

If this case function `filter_format_fields`(`format_metadata `) can check if `fields` option is specified and return full format if isn't.

crud/common/schema.lua
```diff
@@ -133,22 +133,42 @@ local function get_space_schema_hash(space)
...
+    local extracted_fields = {}
+
+    for i, field in ipairs(fields) do
+        extracted_fields[i] = tuple[field]
+    end
+
+    return extracted_fields
+end
+
+local function get_partial_result(tuple, fields)
```
dokshina, line 146: What about `filter_result_fields`?.

I'm not sure that we need `extract_fields` as a separated function.

It's ok to use early-return at the function begin and do other things after that:

```lua
local function filter_result_fields(tuple, fields)
    if fields == nil or tuple == nil then
        return tuple
    end

   for i, field in ipairs(fields) do
      ...
   end

   return smth
end
```

test/integration/simple_operations_test.lua
```diff
@@ -391,3 +391,233 @@ pgroup:add('test_upsert', function(g)
     t.assert_equals(err, nil)
     t.assert_equals(result.rows, {{67, 1143, 'Mikhail Saltykov-Shchedrin', 63}})
 end)
+
+pgroup:add('test_get_partial_result', function(g)
+    -- insert_object
+    local result, err = g.cluster.main_server.net_box:call(
+            'crud.insert_object', {'customers', {id = 1, name = 'Elizabeth', age = 24}})
```
dokshina, line 398: What about "bad input" tests? What if user specified some bad field name?

olegrok: > What about "bad input" tests? What if user specified some bad field name?

I believe that we should do as Tarantool:
```lua
tarantool> box.tuple.new({1, {a = {b = 'c'}}})
---
- [1, {'a': {'b': 'c'}}]
...

tarantool> box.tuple.new({1, {a = {b = 'c'}}})['[1]']
---
- 1
...

tarantool> box.tuple.new({1, {a = {b = 'c'}}})['[2]']
---
- {'a': {'b': 'c'}}
...

tarantool> box.tuple.new({1, {a = {b = 'c'}}})['[2].a.b']
---
- c
...
tarantool> box.tuple.new({1, {a = {b = 'c'}}})['K']
---
- null
...
```

This makes this feature quite powerful and allows to fetch not some tuple fields but some nested structures in tuple fields.

test/integration/simple_operations_test.lua
olegrok:
```
tarantool> vshard.router.bucket_id_strcrc32(1)
---
- 477
...
```

mRrvz: Okay, but maybe should do it like this?

```lua
t.assert_equals(objects, {{id = 1, name = 'Elizabeth', age = 24, bucket_id = vshard.router.bucket_id_strcrc32(1)}}) 
```

or maybe leave a comment on how `477` is formed

test/integration/simple_operations_test.lua
```diff
@@ -391,3 +391,286 @@ pgroup:add('test_upsert', function(g)
...
+    })
+
+    t.assert_equals(err, nil)
+    t.assert_equals(result.metadata, {
+        {name = 'id', type = 'unsigned'},
+        {name = 'bucket_id', type = 'unsigned'},
+        {name = 'name', type = 'string'},
+        {name = 'age', type = 'number'},
+    })
+    t.assert_equals(#result.rows, 0)
```
mRrvz, line 577:
```lua
t.assert_equals(err, nil)
t.assert_equals(#result.rows, 0)
t.assert_equals(result.metadata, {
    {name = 'bucket_id', type = 'unsigned'},
    {name = 'name', type = 'string'},
    {name = 'age', type = 'number'},
})
```

test/integration/simple_operations_test.lua
AnaNek: Issue #120 was opened

crud/common/utils.lua
```diff
@@ -251,11 +254,57 @@ function utils.is_uuid(value)
...
+    return metadata_cache[full_metadata][field_name]
+end
+
+local function filter_format_fields(full_metadata, fields)
+    dev_checks('table', 'table')
+
+    local metadata = {}
+
+    for i, field in ipairs(fields) do
+        metadata[i] = get_field_metadata(full_metadata, field)
```
dokshina, line 279: What is metadata here? 
`filter_format_fields` function accepts `(space_format, field_names)`.

`get_field_metadata` is `get_field_format` in fact. And `metadata_cache` is `space_format_cache`.

I mean that word "metadata" should exists only in `format_result` function, whet we pass *space format* as *metadata*, but all these functions know nothing about metadata, they operate with *space format*.

crud/common/utils.lua
```diff
@@ -251,11 +254,57 @@ function utils.is_uuid(value)
...
+    dev_checks('table', 'string')
+
+    local metadata = metadata_cache[full_metadata]
+    if metadata ~= nil then
+        return metadata[field_name]
+    end
+
+    metadata_cache[full_metadata] = {}
+    for _, field in ipairs(full_metadata) do
+        metadata_cache[full_metadata][field.name] = field
```
dokshina, line 267: Here you can see the difference between `field` and `field_name`, and variable name says that some variables are field objects (table with some field data), but others are `field_name` - a string with name of field.

But in `filter_format_fields` function `fields` is used as `field_names` and it confuses. 
I think that the only one place when we use `fields` name for field names array is CRUD functions API.
User says `crud.do_smth('space', {fields = {...}})`, we accept `opts.fields` here (because it's comfortable for user), but then we can use this value in internal functions as `field_names` to highlight that this is array of strings, not field objects (e.g. `*_on_storage` function can accept `field_names` since it's internal).

crud/common/utils.lua
```diff
@@ -282,11 +285,57 @@ function utils.is_uuid(value)
...
+        end
+    end
+
+    return metadata
+end
+
+function utils.format_result(rows, space, field_names)
+    local result = {}
+    local err
+    local space_format = table.copy(space:format())
```
olegrok, line 324: table.copy here is an attempt to protect space:format() value form user modifications.

If you need to create custom metadata you actually perform this copy manually. So you don't need to perform `table.copy` if fields are specified.

test/integration/simple_operations_test.lua
```diff
@@ -467,17 +467,17 @@ pgroup:add('test_object_with_nullable_fields', function(g)
...
 
     -- {3, 2804, NULL, NULL, NULL, NULL, NULL, true, NULL, true, true, NULL}
     -- Shouldn't failed because of https://github.com/tarantool/tarantool/issues/3378
     _, err = g.cluster.main_server.net_box:call(
-        'crud.upsert_object', {'tags', {id = 3, is_dirty = true}, {
-             {'=', 'is_useful', true},
-    }})
+            'crud.upsert_object', {'tags', {id = 3, is_dirty = true}, {
+                {'=', 'is_useful', true},
+            }})
```
dokshina, line 480: Seems that this changed are redundant.

dokshina:
Thanks a lot for your patch! Me and @olegrok left two little comments.

Please, also update CHANGELOG and describe new option in README and then we can merge this PR!

dokshina (approved):
Thanks a lot for your patch!

olegrok (approved):
Bless you

Don't forget to squash all your commit into single patch.

PR #121, opened 2021-02-20
Update function now adds intermediate nullable fields (if they needed)

### LINKED ISSUE #113: [2pt] Add intermediate nullable fields on update if not present in tuple
Found by customer (again).

Expected behavior

When adding a field via update() to the end of the tuple, if there are nullable fields in between which are not present yet, the operation fails with an error like `Field ''statistics'' was not found in the tuple`.

Actual behavior

Such operation completes successfully, as in the case if the user adds the absent intermediate fields.

Additional notes

Related to https://github.com/tarantool/tarantool/issues/3378 -- already affects SQL and there are no forecasts about its destiny. So, it is possible to add a w/a on the CRUD module level, for not multiplying the clusterfuck effect of the bug.

mRrvz:
Fixed not fining field in tuple on ``crud.update`` if there are ``is_nullable`` fields in front of it that were added when the schema was changed. Closes #113

### DIFF
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -261,7 +261,7 @@ local function convert_jsonpath_operation(operation)
     if #field_parts > 1 or field_id == nil then
         -- Checking '[4].a.b' case
         if field_name:sub(1, 1) == '[' and field_name:sub(-1, -1) == ']' then
-            local field_id = tonumber(field_name:sub(2, -2))
+            field_id = tonumber(field_name:sub(2, -2))
 
             if field_id == nil then
                 -- Checking '["field_name"]' case
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -199,23 +199,23 @@ function utils.tarantool_supports_uuids()
     return enabled_tarantool_features.uuids
 end
 
-local function exists_operations_id_map(operations)
-    local map = {}
-    for _, operation in ipairs(operations) do
-        map[operation[2]] = true
+local function map_to_table(map)
+    local tab = {}
+    for _, value in pairs(map) do
+        table.insert(tab, value)
     end
 
-    return map
+    return tab
 end
 
-local function add_nullable_fields_rec(operations, operations_id_map, space_format, tuple, id)
+local function add_nullable_fields_recursive(operations, space_format, tuple, id)
     if id < 2 or tuple[id - 1] ~= box.NULL then
         return operations
     end
 
-    if space_format[id - 1].is_nullable and not operations_id_map[id - 1] then
-        table.insert(operations, {'=', id - 1, box.NULL})
-        return add_nullable_fields_rec(operations, operations_id_map, space_format, tuple, id - 1)
+    if space_format[id - 1].is_nullable and not operations[id - 1] then
+        operations[id - 1] = {'=', id - 1, box.NULL}
+        return add_nullable_fields_recursive(operations, space_format, tuple, id - 1)
     end
 
     return operations
@@ -237,40 +237,58 @@ function utils.add_intermediate_nullable_fields(operations, space_format, tuple)
         return operations
     end
 
-    if utils.tarantool_supports_fieldpaths() then
-        local formatted_operations, err = utils.convert_operations(operations, space_format)
-        if err ~= nil then
-            return operations
-        end
-
-        operations = formatted_operations
+    -- We need this map to check existence of an operation in constant complexity
+    local operations_map, err = utils.get_operations_map(operations, space_format)
+    if err ~= nil then
+        return operations
     end
 
-    -- We need this map to check existence of an operation in constant complexity
-    local operations_id_map = exists_operations_id_map(operations)
-    for i = 1, #operations do
-        operations = add_nullable_fields_rec(
-            operations,
-            operations_id_map,
-            space_format,
-            tuple,
-            operations[i][2]
+    for key, _ in pairs(operations_map) do
+        operations_map = add_nullable_fields_recursive(
+            operations_map, space_format,
+            tuple, key
         )
     end
 
-    table.sort(operations, function(v1, v2) return v1[2] < v2[2] end)
-    return operations
+    return map_to_table(operations_map)
+end
+
+local function convert_jsonpath_operation(operation)
+    local field_parts = operation[2]:split('.', 1)
+    local field_name = field_parts[1]
+    local field_id = tonumber(field_name:sub(2, -2))
+
+    if #field_parts > 1 or field_id == nil then
+        -- Checking '[4].a.b' case
+        if field_name:sub(1, 1) == '[' and field_name:sub(-1, -1) == ']' then
+            local field_id = tonumber(field_name:sub(2, -2))
+
+            if field_id == nil then
+                -- Checking '["field_name"]' case
+                if field_name:sub(2, 2) == '"' and field_name:sub(-2, -2) == '"' then
+                    return field_name:sub(3, -3), nil
+                end
+            end
+
+            return nil, tonumber(field_name:sub(2, -2))
+        end
+    end
+
+    return field_name, nil
 end
 
 function utils.convert_operations(user_operations, space_format)
+    if utils.tarantool_supports_fieldpaths() then
+        return user_operations
+    end
+
     local converted_operations = {}
 
     for _, operation in ipairs(user_operations) do
         if type(operation[2]) == 'string' then
             local field_id
             for fieldno, field_format in ipairs(space_format) do
-                local operation_name = operation[2]:split('.', 1)[1]
-                if field_format.name == operation_name then
+                if field_format.name == operation[2] then
                     field_id = fieldno
                     break
                 end
@@ -290,6 +308,36 @@ function utils.convert_operations(user_operations, space_format)
     end
 
     return converted_operations
+end
+
+function utils.get_operations_map(user_operations, space_format)
+    local operations_map = {}
+
+    for _, operation in ipairs(user_operations) do
+        if type(operation[2]) == 'string' then
+            local field_name, field_id = convert_jsonpath_operation(operation)
+
+            if field_id == nil then
+                for fieldno, field_format in ipairs(space_format) do
+                    if field_format.name == field_name then
+                        field_id = fieldno
+                        break
+                    end
+                end
+
+                if field_id == nil then
+                    return nil, ParseOperationsError:new(
+                            "Space format doesn't contain field named %q", operation[2])
+                end
+            end
+
+            operations_map[field_id] = operation
+        else
+            operations_map[operation[2]] = operation
+        end
+    end
+
+    return operations_map
 end
 
 function utils.unflatten_rows(rows, metadata)
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -291,32 +291,6 @@ function utils.convert_operations(user_operations, space_format)
     return converted_operations
 end
 
-function utils.get_operations_map(user_operations, space_format)
-    local operations_map = {}
-
-    for _, operation in ipairs(user_operations) do
-        if type(operation[2]) == 'string' then
-            local field_id
-            for fieldno, field_format in ipairs(space_format) do
-                if field_format.name == operation[2] then
-                    field_id = fieldno
-                    break
-                end
-            end
-
-            if field_id == nil then
-                return nil, ParseOperationsError:new(
-                        "Space format doesn't contain field named %q", operation[2])
-            end
-            operations_map[field_id] = operation
-        else
-            operations_map[operation[2]] = operation
-        end
-    end
-
-    return operations_map
-end
-
 function utils.unflatten_rows(rows, metadata)
     if metadata == nil then
         return nil, UnflattenError:new('Metadata is not provided')
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -199,6 +199,77 @@ function utils.tarantool_supports_uuids()
     return enabled_tarantool_features.uuids
 end
 
+function utils.table_to_string(tbl)
+    local result = "{"
+    for k, v in pairs(tbl) do
+        -- Check the key type (ignore any numerical keys - assume its an array)
+        if type(k) == "string" then
+            result = result.."[\""..k.."\"]".."="
+        end
+
+        -- Check the value type
+        if type(v) == "table" then
+            result = result..utils.table_to_string(v)
+        elseif type(v) == "boolean" then
+            result = result..tostring(v)
+        else
+            result = result.."\""..v.."\""
+        end
+        result = result..","
+    end
+    -- Remove leading commas from the result
+    if result ~= "" then
+        result = result:sub(1, result:len()-1)
+    end
+    return result.."}"
+end
+
+function utils.add_intermediate_nullable_fields(user_operations, space_format)
+    local math = require('math')
+    local formatted_operations = {}
+    local max_field_id = 0
+    local file = io.open('test_form', 'a+')
+    io.output(file)
+
+    for _, operation in ipairs(user_operations) do
+        if type(operation[2]) == 'string' then
+            local field_id
+            for fieldno, field_format in ipairs(space_format) do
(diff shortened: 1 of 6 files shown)


crud/update.lua
```diff
@@ -62,6 +62,12 @@ local function call_update_on_router(space_name, key, user_operations, opts)
         return nil, UpdateError:new("Wrong operations are specified: %s", err), true
     end
 
+    if type(key) == 'table' or type(key) == 'number' then
```
olegrok, line 65: Why "number"? space:get(...) works fine even with strings and other types. Seems this check is redundant.

crud/update.lua
```diff
@@ -62,6 +62,12 @@ local function call_update_on_router(space_name, key, user_operations, opts)
...
+        local tuple = space:get{key}
```
olegrok, line 66: You tries to get tuple on router but actually it is stored on storage. Also this construction is a bit wrong `{key}` -> `(key)`: otherwise it won't work for multipart indexes.

Also seems you've missed my comment that I shared you privately.
Update should be *optimistic*. Persons that perform correct update should get the same result - only one operation. If you get `box.error.NO_SUCH_FIELD_NO` or `box.error.NO_SUCH_FIELD_NAME` you should start to think about "adding intermediate fields" - try to go "get" tuple, add missing fields and then update.

crud/common/utils.lua
```diff
@@ -199,11 +199,47 @@ function utils.tarantool_supports_uuids()
...
+        for i = 1, #operations do
+            operations = add_nullable_fields_rec(
+                operations,
+                space_format,
+                tuple,
+                operations[i][2]
+            )
+        end
+
+        table.sort(operations, function(v1, v2) return v1[2] < v2[2] end)
```
olegrok, line 236: You can't sort fields by number. Because it could be update by number or even jsonpath updates.

olegrok: Originally it was stated in issue - https://github.com/tarantool/crud/issues/113: "Field ''statistics'' was not found in the tuple"

crud/common/utils.lua
```diff
@@ -199,11 +199,47 @@ function utils.tarantool_supports_uuids()
...
+    return true
+end
+
+local function add_nullable_fields_rec(operations, space_format, tuple, id)
+    if id < 2 or tuple[id - 1] ~= box.NULL then
+        return operations
+    end
+
+    if space_format[id - 1].is_nullable and id_not_in_table(operations, id - 1) then
+        table.insert(operations, {'=', id - 1, box.NULL})
```
olegrok, line 218: What if I have tuple `{1}` and tried to do "tuple:update({{'=', 4, 'value'}})", adding of only single field won't work - you need to add two nulls.

olegrok: More complex case - `update({{'=', 4, value1}, {'=', "some_field", value2}})` and "some_field" is 6th field.

olegrok (commented):
Thanks for you patch consider several comments below.

I've put quite a lot comments, so I need to clarify some expectations from this patch.
- Updates should be optimistic - we shouldn't get tuple before we understand that some field is missing: error has code `box.error.NO_SUCH_FIELD_NO` or `box.error.NO_SUCH_FIELD_NAME`
- This logic should be performed on storage-side
- We should try to do it only for top-level fields (don't think about jsonpath updates)
- Tarantool allows to add fields after fields that declared in space format, but I think it's not general purpose case IMO - so we need to do *something* only if field name is in space format (but for some reasons some fields is missed before this field) or field number <= space format length.

olegrok (commented):
Thanks for your fixes. Consider a couple of comments below.

crud/common/utils.lua
```diff
@@ -199,11 +199,71 @@ function utils.tarantool_supports_uuids()
...
+        return add_nullable_fields_rec(operations, space_format, tuple, id - 1)
     end
 
+    return operations
+end
+
+-- Tarantool < 2 has no fields `box.error.NO_SUCH_FIELD_NO` and `box.error.NO_SUCH_FIELD_NAME`.
+function utils.is_field_not_found(err_code)
+    local patch_parts = _G._TARANTOOL:split('-', 1)[1]:split('.', 2)
+    local major = tonumber(patch_parts[1])
```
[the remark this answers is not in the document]
dokshina: Now we have several functions that checks Tarantool version, maybe it's tine to cache Tarantool major, minor, patch versions?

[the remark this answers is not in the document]
olegrok: I'd call it "parse Tarantool versions on module load". But anyway even you have three numbers - minor, major, patch you still need to perform some comparison operations that could be easily eliminated.

test/integration/simple_operations_test.lua
```diff
@@ -392,6 +392,93 @@ pgroup:add('test_upsert', function(g)
...
+        server.net_box:call('add_extra_field', {'extra_2'})
+        server.net_box:call('add_extra_field', {'extra_3'})
+        server.net_box:call('add_extra_field', {'extra_4'})
+        server.net_box:call('add_extra_field', {'extra_5'})
+        server.net_box:call('add_extra_field', {'extra_6'})
+        server.net_box:call('add_extra_field', {'extra_7'})
+        server.net_box:call('add_extra_field', {'extra_8'})
+        server.net_box:call('add_extra_field', {'extra_9'})
+        server.net_box:call('add_extra_field', {'extra_10'})
+        server.net_box:call('add_extra_field', {'extra_11'})
```
dokshina, line 419: What about a cycle?

crud/common/utils.lua
```diff
@@ -199,11 +199,64 @@ function utils.tarantool_supports_uuids()
...
+        return err_code == box.error.NO_SUCH_FIELD
+    end
+end
+
+function utils.add_intermediate_nullable_fields(operations, space_format, tuple)
+    if tuple == nil then
+        return operations
+    end
+
+    local formatted_operations, err = utils.convert_operations(operations, space_format)
```
dokshina, line 241: If `not utils.tarantool_supports_fieldpaths()`, we have already format the operations.

crud/common/utils.lua
```diff
@@ -238,22 +237,29 @@ function utils.add_intermediate_nullable_fields(operations, space_format, tuple)
         return operations
     end
 
-    local formatted_operations, err = utils.convert_operations(operations, space_format)
-    if err ~= nil then
-        return operations
+    if utils.tarantool_supports_fieldpaths() then
```
dokshina, line 240: Please, add a comment - why do we convert operations only in this case.

crud/common/utils.lua
```diff
@@ -199,24 +199,23 @@ function utils.tarantool_supports_uuids()
     return enabled_tarantool_features.uuids
 end
 
-local function id_not_in_table(operations, id)
+local function exists_operations_id_map(operations)
```
dokshina, line 202: I would like this function name to start via `get_`.

dokshina (commented):
Please, consider my comments and update a Changelog.

crud/common/utils.lua
olegrok: Also seems it's not `operations_id` it's field id.

crud/common/utils.lua
```diff
@@ -199,18 +199,78 @@ function utils.tarantool_supports_uuids()
...
+    return map
+end
+
+local function add_nullable_fields_rec(operations, operations_id_map, space_format, tuple, id)
+    if id < 2 or tuple[id - 1] ~= box.NULL then
+        return operations
+    end
+
+    if space_format[id - 1].is_nullable and not operations_id_map[id - 1] then
+        table.insert(operations, {'=', id - 1, box.NULL})
```
olegrok, line 217: maybe it's possible to use table.insert(operations, index, {'=', id - 1, box.NULL}) here to eliminate table.sort below?

crud/common/utils.lua
```diff
@@ -199,18 +199,78 @@ function utils.tarantool_supports_uuids()
...
+local function exists_operations_id_map(operations)
+    local map = {}
+    for _, operation in ipairs(operations) do
+        map[operation[2]] = true
+    end
+
+    return map
+end
+
+local function add_nullable_fields_rec(operations, operations_id_map, space_format, tuple, id)
```
olegrok, line 211: What does "rec" mean here?

crud/common/utils.lua
```diff
@@ -199,18 +199,78 @@ function utils.tarantool_supports_uuids()
...
+else
+    function utils.is_field_not_found(err_code)
+        return err_code == box.error.NO_SUCH_FIELD
+    end
+end
+
+function utils.add_intermediate_nullable_fields(operations, space_format, tuple)
+    if tuple == nil then
+        return operations
+    end
```
olegrok, line 238: Is this check really needed?

As I understand if you try to update tuple that doesn't actually exist you don't have any error at all:
```
tarantool> box.space.tasks:update({10000000000}, {{'=', 'abc.d.e.f.g.h', 1000}})
---
...

tarantool> box.space.tasks:update({10000000000}, {{'+', 'abc.d.e.f.g.h', 1000}})
---
...
```

crud/common/utils.lua
```diff
@@ -199,18 +199,78 @@ function utils.tarantool_supports_uuids()
...
+
+function utils.convert_operations(user_operations, space_format)
     local converted_operations = {}
 
     for _, operation in ipairs(user_operations) do
         if type(operation[2]) == 'string' then
             local field_id
             for fieldno, field_format in ipairs(space_format) do
-                if field_format.name == operation[2] then
+                local operation_name = operation[2]:split('.', 1)[1]
```
olegrok, line 272: It won't work.

Because tuple['["field_name"]'] is appropriate way to get/update tuple field

crud/common/utils.lua
```diff
@@ -199,18 +199,78 @@ function utils.tarantool_supports_uuids()
...
+
     if utils.tarantool_supports_fieldpaths() then
-        return user_operations
+        local formatted_operations, err = utils.convert_operations(operations, space_format)
+        if err ~= nil then
+            return operations
+        end
+
+        operations = formatted_operations
+    end
```
olegrok, line 247: Please don't do that. If Tarantool supports JSONpath updates you should pass user defined operation as is. Your changes should be invisible for users that perform correct update operation. Currently it's not so.

Additional logic should be performed only if you get "field number/name doesn't exist"

test/integration/simple_operations_test.lua
```diff
@@ -392,6 +392,99 @@ pgroup:add('test_upsert', function(g)
...
+            id = 1,
+            bucket_id = 477,
+            extra_1 = box.NULL,
+            extra_2 = box.NULL,
+            extra_3 = {a = {b = {}}},
+        }
+    })
+
+    result, err = g.cluster.main_server.net_box:call('crud.update',
+        {'developers', 1, {{'=', 'extra_3.a.b[1]', 2}, {'=', 'extra_5', 'extra_value_5'}}})
```
olegrok, line 429: Should be placed under `if _TARANTOOL >= 2.3` otherwise you will get red CI on 1.10.

crud/common/utils.lua
mRrvz: I need this check because in ``add_nullable_fields_recursive`` function I am referring to the tuple by index

crud/common/utils.lua
mRrvz: It's mean recursive, I change "rec" -> "recursive"

crud/common/utils.lua
```diff
@@ -199,6 +199,84 @@ function utils.tarantool_supports_uuids()
...
+                    return field_name:sub(3, -3), nil
+                end
+            end
+
+            return nil, tonumber(field_name:sub(2, -2))
+        end
+    end
+
+    return field_name, nil
+end
```
olegrok, line 278: Please don't try to parse jsonpaths. You could see json parser that compatible with Tarantool jsonpath here - https://github.com/olegrok/tarantool-jsonpath/blob/master/jsonpath.lua

Actually, you don't need here

crud/update.lua
```diff
@@ -46,17 +68,20 @@ local function call_update_on_router(space_name, key, user_operations, opts)
...
         return nil, UpdateError:new("Space %q doesn't exist", space_name), true
     end
 
-    local space_format = space:format()
-
     if box.tuple.is(key) then
         key = key:totable()
     end
 
+    local space_format = space:format()
```
olegrok, line 84: This changes are not related to your patch anyhow.

crud/common/utils.lua
```diff
@@ -199,6 +199,84 @@ function utils.tarantool_supports_uuids()
...
+        return operations
+    end
+
+    if space_format[id - 1].is_nullable and not operations[id - 1] then
+        operations[id - 1] = {'=', id - 1, box.NULL}
+        return add_nullable_fields_recursive(operations, space_format, tuple, id - 1)
+    end
+
+    return operations
+end
```
olegrok, line 222: Required logic if you get an error:
  - Get tuple
  - Find field in format
  - If field is in format - check are there any field before this
  - JSONpath is not in format - just skip it and don't try to parse it
  - If there is no field in tuple - just call `table.insert(updates_list, index, {'=', field_in_tuple, box.NULL})`

crud/update.lua
```diff
@@ -24,10 +24,32 @@ local function update_on_storage(space_name, key, operations, field_names)
...
+    if err ~= nil then
+        return nil, err
+    end
+
+    if res.err == nil then
+        return res, nil
+    end
+
+    -- We can only add fields to end of the tuple.
+    -- If schema is updated and intermediate null fields are added, then we will get error.
```
olegrok, line 41: `null` -> `nullable`

 `intermediate` ... Not sure it's correct. You can only append fields to format. They are trailing.

crud/update.lua
```diff
@@ -24,10 +24,32 @@ local function update_on_storage(space_name, key, operations, field_names)
...
+    -- Therefore, we need to add filling of intermediate null fields.
```
olegrok, line 42: `null` -> `nullable`.

Here "intermediate" is correct because such fields are between target and first non-null field.

crud/update.lua
```diff
@@ -24,10 +24,32 @@ local function update_on_storage(space_name, key, operations, field_names)
...
+    -- More details: https://github.com/tarantool/crud/issues/113
```
olegrok, line 43: Even better to refer to https://github.com/tarantool/tarantool/issues/3378

crud/common/utils.lua
```diff
@@ -232,6 +286,32 @@ function utils.convert_operations(user_operations, space_format)
...
+
+    for _, operation in ipairs(user_operations) do
+        if type(operation[2]) == 'string' then
+            local field_id
+            for fieldno, field_format in ipairs(space_format) do
+                if field_format.name == operation[2] then
+                    field_id = fieldno
+                    break
+                end
+            end
```
olegrok, line 300: Still you have O(n^2) for operations_map calculation if update by name is used.

crud/common/utils.lua
```diff
@@ -199,6 +199,60 @@ function utils.tarantool_supports_uuids()
     return enabled_tarantool_features.uuids
 end
 
+local function map_to_table(map)
```
olegrok, line 202: map_to_array?

crud/common/utils.lua
```diff
@@ -199,6 +199,60 @@ function utils.tarantool_supports_uuids()
...
+    local tab = {}
+    for _, value in pairs(map) do
+        table.insert(tab, value)
+    end
```
olegrok, line 206: Here you use an assumption that pairs will return you values ordered by indexes. I'm not sure that it's correct.

olegrok: Maybe we don't need to use "recursive" approach here. It's really not obvious why `operations[id - 1] = {'=', id - 1, box.NULL}` can't rewrite some user operation.

olegrok (commented):
Thanks for your changes and fixes. Sorry for some controversial comments before.

Currently I basically agree with your patch and with your approach. I put some comments about some details. Please ask @dokshina for the second review.

dokshina (approved):
Nice work!

PR #122, opened 2021-02-25
Support partial results for select/pairs

### LINKED ISSUE #117: [4pt] Select/Pairs: Partial results
The second and complex part of #53

AnaNek:
Closes #117

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 * Support for UUID field types and UUID values
-* `fields` option for simple operations to get partial result
-* `fields` option for select/pairs with pagination support
+* `fields` option for simple operations and select/pairs calls with pagination support to get partial result
 
 ## [0.4.0] - 2020-12-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 * Support for UUID field types and UUID values
 * `fields` option for simple operations to get partial result
+* `fields` option for select/pairs with pagination support
 
 ## [0.4.0] - 2020-12-02
 
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -136,7 +136,6 @@ local function get_space_schema_hash(space)
 end
 
 local function filter_tuple_fields(tuple, field_names)
-
     if field_names == nil or tuple == nil then
         return tuple
     end
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -135,6 +135,37 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
+function schema.truncate_tuple_trailing_fields(tuple, field_names)
+    dev_checks('table|cdata', '?table')
+
+    if field_names == nil then
+        return tuple
+    end
+
+    local index = #field_names + 1
+    local len_tuple = #tuple
+
+    for i = index, len_tuple do
+        tuple[i] = nil
+    end
+
+    return tuple
+end
+
+function schema.truncate_tuples_trailing_fields(tuples, field_names)
+    dev_checks('table|cdata', '?table')
+
+    if field_names == nil then
+        return tuples
+    end
+
+    for i, _ in ipairs(tuples) do
+        tuples[i] = schema.truncate_tuple_trailing_fields(tuples[i], field_names)
+    end
+
+    return tuples
+end
+
 function schema.filter_tuple_fields(tuple, field_names)
     dev_checks('?table|cdata', '?table')
 
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -135,7 +135,7 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
-local function filter_result_fields(tuple, field_names)
+function schema.filter_result_fields(tuple, field_names)
     if field_names == nil or tuple == nil then
         return tuple
     end
@@ -149,6 +149,28 @@ local function filter_result_fields(tuple, field_names)
                     'Space format doesn\'t contain field named %q', field_name
             )
         end
+    end
+
+    return result
+end
+
+function schema.filter_tuples_fields(tuples, field_names)
+    dev_checks('?table', '?table')
+
+    if field_names == nil then
+        return tuples
+    end
+
+    local result = {}
+
+    for _, tuple in ipairs(tuples) do
+        local filtered_tuple, err = schema.filter_result_fields(tuple, field_names)
+
+        if err ~= nil then
+            return nil, err
+        end
+
+        table.insert(result, filtered_tuple)
     end
 
     return result
@@ -174,7 +196,7 @@ function schema.wrap_box_space_func_result(space, func_name, args, opts)
             result.space_schema_hash = get_space_schema_hash(space)
         end
     else
-        result.res, err = filter_result_fields(func_res, opts.field_names)
+        result.res, err = schema.filter_result_fields(func_res, opts.field_names)
         if err ~= nil then
             return nil, err
         end
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -135,7 +135,9 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
-function schema.filter_result_fields(tuple, field_names)
+function schema.filter_tuple_fields(tuple, field_names)
+    dev_checks('?table|cdata', '?table')
+
     if field_names == nil or tuple == nil then
         return tuple
     end
@@ -164,7 +166,7 @@ function schema.filter_tuples_fields(tuples, field_names)
     local result = {}
 
     for _, tuple in ipairs(tuples) do
-        local filtered_tuple, err = schema.filter_result_fields(tuple, field_names)
+        local filtered_tuple, err = schema.filter_tuple_fields(tuple, field_names)
 
         if err ~= nil then
             return nil, err
@@ -196,7 +198,7 @@ function schema.wrap_box_space_func_result(space, func_name, args, opts)
             result.space_schema_hash = get_space_schema_hash(space)
         end
     else
-        result.res, err = schema.filter_result_fields(func_res, opts.field_names)
+        result.res, err = schema.filter_tuple_fields(func_res, opts.field_names)
         if err ~= nil then
             return nil, err
         end
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -136,7 +136,7 @@ local function get_space_schema_hash(space)
 end
 
 function schema.truncate_tuple_trailing_fields(tuple, field_names)
-    dev_checks('table|cdata', '?table')
+    dev_checks('table', '?table')
 
     if field_names == nil then
         return tuple
@@ -153,7 +153,7 @@ function schema.truncate_tuple_trailing_fields(tuple, field_names)
 end
 
 function schema.truncate_tuples_trailing_fields(tuples, field_names)
-    dev_checks('table|cdata', '?table')
+    dev_checks('table', '?table')
 
     if field_names == nil then
         return tuples
@@ -166,8 +166,7 @@ function schema.truncate_tuples_trailing_fields(tuples, field_names)
     return tuples
 end
 
-function schema.filter_tuple_fields(tuple, field_names)
-    dev_checks('?table|cdata', '?table')
+local function filter_tuple_fields(tuple, field_names)
 
     if field_names == nil or tuple == nil then
         return tuple
@@ -197,7 +196,7 @@ function schema.filter_tuples_fields(tuples, field_names)
     local result = {}
 
     for _, tuple in ipairs(tuples) do
-        local filtered_tuple, err = schema.filter_tuple_fields(tuple, field_names)
+        local filtered_tuple, err = filter_tuple_fields(tuple, field_names)
 
         if err ~= nil then
             return nil, err
@@ -229,7 +228,7 @@ function schema.wrap_box_space_func_result(space, func_name, args, opts)
             result.space_schema_hash = get_space_schema_hash(space)
         end
     else
-        result.res, err = schema.filter_tuple_fields(func_res, opts.field_names)
+        result.res, err = filter_tuple_fields(func_res, opts.field_names)
         if err ~= nil then
             return nil, err
         end
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -135,38 +135,6 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
-function schema.convert_tuple_to_space_format(space_format, field_names, tuple)
-    dev_checks('table', '?table', '?table')
-
-    if tuple == nil then
-        return nil
-    end
-
-    if field_names == nil then
-        return tuple
-    end
-
-    local positions = {}
-    local transformed_tuple = {}
-
-    for i, field in ipairs(space_format) do
-        positions[field.name] = i
-    end
-
-    for i, field_name in ipairs(field_names) do
-        local fieldno = positions[field_name]
-        if fieldno == nil then
-            return nil, FilterFieldsError:new(
-                    'Space format doesn\'t contain field named %q', field_name
-            )
-        end
-
-        transformed_tuple[fieldno] = tuple[i]
-    end
-
-    return transformed_tuple
-end
-
 local function filter_tuple_fields(tuple, field_names)
 
     if field_names == nil or tuple == nil then
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -135,35 +135,36 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
-function schema.truncate_tuple_trailing_fields(tuple, field_names)
-    dev_checks('table', '?table')
+function schema.convert_tuple_to_space_format(space_format, field_names, tuple)
+    dev_checks('table', '?table', '?table')
+
+    if tuple == nil then
+        return nil
+    end
 
     if field_names == nil then
         return tuple
     end
 
-    local index = #field_names + 1
-    local len_tuple = #tuple
+    local positions = {}
+    local transformed_tuple = {}
 
-    for i = index, len_tuple do
-        tuple[i] = nil
+    for i, field in ipairs(space_format) do
+        positions[field.name] = i
     end
 
-    return tuple
-end
+    for i, field_name in ipairs(field_names) do
+        local fieldno = positions[field_name]
+        if fieldno == nil then
+            return nil, FilterFieldsError:new(
+                    'Space format doesn\'t contain field named %q', field_name
+            )
+        end
 
-function schema.truncate_tuples_trailing_fields(tuples, field_names)
-    dev_checks('table', '?table')
-
-    if field_names == nil then
-        return tuples
+        transformed_tuple[fieldno] = tuple[i]
     end
 
-    for i, _ in ipairs(tuples) do
-        tuples[i] = schema.truncate_tuple_trailing_fields(tuples[i], field_names)
-    end
-
-    return tuples
+    return transformed_tuple
 end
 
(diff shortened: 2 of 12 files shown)


crud/common/schema.lua
```diff
@@ -135,7 +135,7 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
-local function filter_result_fields(tuple, field_names)
+function schema.filter_result_fields(tuple, field_names)
```
olegrok, line 138: nit: Since you add this function public, let's add dev_checks.

test/integration/select_test.lua
```diff
@@ -763,3 +763,68 @@ pgroup:add('test_multipart_primary_index', function(g)
...
+    local conditions = {{'>=', 'age', 33}}
+    local result, err = g.cluster.main_server.net_box:call('crud.select',
+            {'customers', conditions, {fields = {'id', 'name'}}}
+    )
+
+    t.assert_equals(err, nil)
+    local objects = crud.unflatten_rows(result.rows, result.metadata)
+    t.assert_equals(objects, expected_customers)
+end)
+
```
olegrok, line 830: This empty line seems to be redundant

test/integration/select_test.lua
```diff
@@ -763,3 +763,68 @@ pgroup:add('test_multipart_primary_index', function(g)
...
+            age = 81, city = "Chicago",
+        },
+    })
+
+    -- in age order
+    local expected_customers = {
+        {id = 3, name = "David"},
+        {id = 2, name = "Mary"},
+        {id = 4, name = "William"},
+    }
```
olegrok, line 817: side note: it looks really strange. You could fetch some field subset but you lose an opportunity to use pagination.

test/integration/pairs_test.lua
```diff
@@ -319,3 +319,49 @@ pgroup:add('test_luafun_compatipility', function(g)
...
+        local objects = {}
+        for _, object in crud.pairs('customers', conditions,  {use_tomap = true, fields = {'id', 'name'}}) do
+            table.insert(objects, object)
+        end
+
+        return objects
+    ]], {conditions})
+    t.assert_equals(objects, expected_customers)
+end)
+
```
olegrok, line 367: Redundant empty line

crud/select.lua
```diff
@@ -305,16 +361,37 @@ function select_module.call(space_name, user_conditions, opts)
             break
         end
 
+        if opts.fields then
+            tuple, err = utils.unflatten(tuple, iter.space_format)
+            if err ~= nil then
+                error(string.format("Failed to unflatten next object: %s", err))
+            end
+        end
```
olegrok, line 369: Why do you perform "unflatten" if "opts.fields" specified?

Seems you need just nullify trailing merged fields.

crud/select.lua
```diff
@@ -247,9 +279,31 @@ function select_module.pairs(space_name, user_conditions, opts)
...
+        if err ~= nil then
+            return nil, err
+        end
+
+        if opts.fields then
+            tuple, err = utils.unflatten(tuple, iter.space_format)
+            if err ~= nil then
+                error(string.format("Failed to unflatten next object: %s", err))
+            end
+        end
```
olegrok, line 296: The same as below. After you fetch and sort tuples you should assign `nil`s to trailing fields that were merged.

crud/common/utils.lua
```diff
@@ -338,6 +357,49 @@ function utils.format_result(rows, space, field_names)
...
+            key_field_names[field_name] = true
+        end
+    end
+
+    for _, field_name in ipairs(field_names) do
+        if not key_field_names[field_name] then
+            table.insert(merged_field_names, field_name)
+            key_field_names[field_name] = true
+        end
+    end
```
olegrok, line 398: I suggest to merge cmp_parts to field_names but not vice versa. Details I'll describe below - it helps to avoid some complex transformations. Finally you need to truncate "redundant" parts of tuple.

olegrok (commented):
Hi! Thanks for your patch. Consider several comments below.

crud/common/schema.lua
```diff
@@ -154,6 +154,28 @@ local function filter_result_fields(tuple, field_names)
...
+    dev_checks('?table', '?table')
+
+    if field_names == nil then
+        return tuples
+    end
+
+    local result = {}
+
+    for _, tuple in ipairs(tuples) do
+        local filtered_tuple, err = schema.filter_result_fields(tuple, field_names)
```
dokshina, line 167: Seems that `filter_result_fields` can be renamed to `filter_tuple_fields`.

crud/select.lua
```diff
@@ -180,6 +189,17 @@ local function build_select_iterator(space_name, user_conditions, opts)
     local scan_index = space.index[plan.index_id]
     local primary_index = space.index[0]
     local cmp_key_parts = utils.merge_primary_key_parts(scan_index.parts, primary_index.parts)
+
+    local field_names = utils.merge_comparison_fields(
+            space_format,
```
dokshina, line 194: Redundant ident.

test/integration/pairs_test.lua
```diff
@@ -319,3 +319,49 @@ pgroup:add('test_luafun_compatipility', function(g)
...
+    })
+
+    -- in age order
+    local expected_customers = {
+        {id = 3, name = "David"},
+        {id = 2, name = "Mary"},
+        {id = 4, name = "William"},
+    }
+
+    table.sort(customers, function(obj1, obj2) return obj1.id < obj2.id end)
```
dokshina, line 347: Redundant line.

test/integration/select_test.lua
```diff
@@ -763,3 +763,68 @@ pgroup:add('test_multipart_primary_index', function(g)
...
+
+    table.sort(customers, function(obj1, obj2) return obj1.id < obj2.id end)
```
dokshina, line 819: Redundant.

test/integration/select_test.lua
```diff
@@ -763,3 +763,68 @@ pgroup:add('test_multipart_primary_index', function(g)
...
+
+    local conditions = {{'>=', 'age', 33}}
+    local result, err = g.cluster.main_server.net_box:call('crud.select',
+            {'customers', conditions, {fields = {'id', 'name'}}}
```
dokshina, line 823: Is `age` indexed field? 

There are some simplest cases that we need to test at the first time:

* condition by indexed non-unique non-primary field: result doesn't contain primary key, result tuples are sorted by field+primary (x2 cases - condition field is id opts.fields and not).
* condition by non-indexed non-unique non-primary field: result doesn't contain primary key, result tuples are sorted by primary (x2 cases - condition field is id opts.fields and not).

I suggest you to test all these cases to be completely sure that everything works fine.
Moreover, it would be great to add short comments for each test to make it clear for the future developers.

crud/common/utils.lua
```diff
@@ -338,6 +357,49 @@ function utils.format_result(rows, space, field_names)
...
+    local updated_key_parts = {}
+    start_fieldno = start_fieldno or 1
+    local current_fieldno = start_fieldno
+
+    for _, part in ipairs(key_parts) do
+        local updated_part = table.copy(part)
+        updated_part.fieldno = current_fieldno
+        current_fieldno = current_fieldno + 1
+        table.insert(updated_key_parts, updated_part)
+    end
```
mRrvz, line 370: Maybe do it like this?

```lua
    local updated_key_parts = {}
    local current_fieldno = start_fieldno or 0

    for i, part in ipairs(key_parts) do
        local updated_part = table.copy(part)
        updated_part.fieldno = current_fieldno + i
        table.insert(updated_key_parts, updated_part)
    end
```

crud/select.lua
```diff
@@ -64,6 +65,8 @@ local function select_on_storage(space_name, index_id, conditions, opts)
         return nil, SelectError:new("Failed to execute select: %s", err)
     end
 
+    tuples = schema.filter_tuples_fields(tuples, opts.field_names)
+
     return tuples
```
mRrvz, line 70: I think there is no need to assign ``schema.filter_tuples_fields(tuples, opts.field_names)`` ​​in ``tuples``:

```lua
return schema.filter_tuples_fields(tuples, opts.field_names)
```

crud/select.lua
```diff
@@ -188,9 +208,18 @@ local function build_select_iterator(space_name, user_conditions, opts)
         return nil, SelectError:new("Failed to generate comparator function: %s", err)
     end
 
+    local filtered_space_format, err = utils.get_fields_format(
+            space_format,
+            field_names
+    )
```
mRrvz, line 214: Maybe just this?

```lua
local filtered_space_format, err = utils.get_fields_format(space_format, field_names)
```

and below.

crud/select.lua
dokshina: Sorry, indent.

crud/common/schema.lua
```diff
@@ -135,7 +135,40 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
-local function filter_result_fields(tuple, field_names)
+function schema.truncate_tuple_trailing_fields(tuple, field_names)
+    dev_checks('table|cdata', '?table')
```
olegrok, line 139: `cdata` -> `tuple`

crud/common/schema.lua
```diff
@@ -135,7 +135,40 @@ local function get_space_schema_hash(space)
...
+
+    for i = index, len_tuple do
+        tuple[i] = nil
+    end
+
+    return tuple
+end
+
+function schema.truncate_tuples_trailing_fields(tuples, field_names)
+    dev_checks('table|cdata', '?table')
```
olegrok, line 156: tuples (array) can't have cdata type

crud/common/schema.lua
```diff
@@ -135,7 +135,40 @@ local function get_space_schema_hash(space)
...
+
+    for i, _ in ipairs(tuples) do
+        tuples[i] = schema.truncate_tuple_trailing_fields(tuples[i], field_names)
+    end
+
+    return tuples
+end
+
+function schema.filter_tuple_fields(tuple, field_names)
+    dev_checks('?table|cdata', '?table')
```
olegrok, line 170: `cdata` -> `tuple`

crud/common/schema.lua
olegrok: Also it won't work for box.tuples - box.tuple is immutable.

```
tarantool> box.tuple.new({1, 2, 3})[3] = nil
---
- error: '[string "box.tuple.new({1, 2, 3})[3] = nil"]:1: ''struct tuple'' cannot
    be indexed with ''number'''
...
```

It works only for tables.

olegrok: If you want to support it also for tuples consider `truncate` function. But I don't think that currently it's needed.

crud/select.lua
```diff
@@ -312,9 +335,17 @@ function select_module.call(space_name, user_conditions, opts)
         utils.reverse_inplace(tuples)
     end
 
+    local filtered_space_format, err = utils.get_fields_format(iter.space_format, opts.fields)
+
```
dokshina, line 339: Remove empty line.

Such `err ~= nil` checks relate to previous line, so it's convenient way to keep them together.

crud/common/utils.lua
```diff
@@ -338,6 +354,44 @@ function utils.format_result(rows, space, field_names)
     return result
 end
 
+function utils.merge_comparison_fields(space_format, key_parts, field_names)
```
dokshina, line 357: This is not "merge" - we crop space_format to field_names and add adds `cmp_keys_parts`.

This function is also specific for select plan, so it can be moved to `select.plan`. 
Moreover, can it be refactored?

I mean, we can transform `field_names` and use it later.

```
local field_names = enrich_field_names_with_cmp_key(field_names, cmp_key_parts)
```

But `updated_space_metadata.key_parts` is used only by `gen_tuples_comparator`, so I suggest you go pass `field_names` to this function and take care about `updated_space_metadata.key_parts`  inside. And it would be great to add a comments why does `fields` optsion affects creating `tuples_comparator`.

dokshina (commented):
Great! Please, consider my comments.

crud/common/schema.lua
```diff
@@ -135,7 +135,8 @@ local function get_space_schema_hash(space)
     return digest.murmur(msgpack.encode(space_info))
 end
 
-local function filter_result_fields(tuple, field_names)
+local function filter_tuple_fields(tuple, field_names)
+
```
olegrok, line 139: Do you really need to add this empty line?

crud/select/comparators.lua
```diff
@@ -83,6 +83,33 @@ local function gen_array_cmp_func(target, key_parts)
...
+
+    for _, part in ipairs(key_parts) do
+        local field_name = space_format[part.fieldno].name
+        if not fields_positions[field_name] then
+            fields_positions[field_name] = last_position
+            last_position = last_position + 1
+        end
+        local updated_part = table.copy(part)
+        updated_part.fieldno = fields_positions[field_name]
+        table.insert(updated_key_parts, updated_part)
```
olegrok, line 107: Could you just make something like `table.insert(updated_key_parts, {field_name = ..., fieldno = ..., is_nullable = ...})`?
"part" is not so huge and complex table.

crud/select/plan.lua
```diff
@@ -106,10 +107,65 @@ local function extract_sharding_key_from_scan_value(scan_value, scan_index, shar
...
+            return nil, FilterFieldsError:new(
+                    'Space format doesn\'t contain field named %q', field_name
+            )
+        end
+
+        transformed_tuple[fieldno] = tuple[i]
+    end
+
+    return transformed_tuple
+end
```
olegrok, line 138: Please, describe your logic in PR description. I don't understand why do you need to construct "after tuple".

Just pass it as the second argument from select.call_on_storage.

```lua
tuples, after_tuple = call()
if tuples == nil then
  -- after_tuple is error here
  return nil, after_tuple 
end

...

return {
  metadata = ...,
  rows = transform(tuples),
  after_tuple = after_tuple,
}
```

Anyway we will need to have some kind of cursor for each storage result when we will implement merger (https://github.com/tarantool/crud/pull/38/files#diff-3c146e04832d06f8726bb2eabde2e134695f8924f7847d3058d05ada80e837b5L114-R79)

olegrok: So you could really simplify reviewer's life if describe all transformations of single tuple: where, why.

E.g. "on storage we strip all fields except user defined fields and fields that are needed for comparison on router", "on router: ..."

olegrok:
I think this patch is quite close to merge.
So consider several final comments:
  - Please add a changelog note
  - Let's extend our [examples](https://github.com/tarantool/crud/blob/master/doc) - add some info about this feature.
  - Squash all commit into singl and add good description of this feature (motivation, restrictions, ...)

crud/select/plan.lua
olegrok: Well, ignore my first comment. But there is a lack of information about tuple lifecycle in case when "fields" is specified - so, the second comment is still relevant.

doc/pairs.md
```diff
@@ -207,3 +207,65 @@ tuples
     - 51
 ...
 ```
+## ``fields`` parameter
+
+With ``fields`` parameter you can choose the value of which fields you would like to get.
+Result contains fields specified by ``fields`` parameter, primary key and scan key. We need primary and scan keys in result to support pagination.
```
dokshina, line 213: Result contains only fields specified by ``fields`` parameter, but scan key and primary key values are merged to the result fields to support pagination (any tuple from result can be simply passed to `after` option).
Using `fields` parameters allows to reduce amount of data transferred from storage.

doc/pairs.md
```diff
@@ -207,3 +207,65 @@ tuples
...
+
+**Example:**
+
```
dokshina, line 216: Seems that we can list space fields here to show that some fields were cut.

doc/pairs.md
```diff
@@ -207,3 +207,65 @@ tuples
...
+objects
+---
+- - id: 6
+    name: Alexey
+    age: 31
+  - id: 4
+    name: Mikhail
+    age: 51
+...
+```
```
dokshina, line 233: Add a note to describe what happened, smth like "we got `name` field as it was specified, `age` field because space was scanned by `age` index and primary key `id`".

doc/pairs.md
```diff
@@ -207,3 +207,65 @@ tuples
...
+
+You can use pagination and ``fields`` option together if you pass tuple gotten with ``fields`` option to ``after``.
```
dokshina, line 235: We need to highlight that `after` tuple should contain the same fields as we receive on `pairs` call with such `fields` parameters.

Smth like

Wrong:

```lua
tuples = {}
for _, tuple in crud.pairs('developers', {{'>=', 'age', 27}} }) do  -- `fields` isn't specified
    table.insert(tuples, tuple)
end

-- THIS WOULD FAIL
-- call `pairs` with `fields` option specified 
-- and pass to `after` tuple that were got without `fields` option
crud.pairs('developers', {{'>=', 'age', 27}}, {fields = {'name'}}, after = tuples[1] })
```

doc/select.md
```diff
@@ -301,3 +301,41 @@ res
...
+res, err = crud.select('developers', {{'>=', 'age', 27}}, { fields = {'id', 'name'}, after = res.rows[1] })
+res
+- metadata:
+  - {'name': 'id', 'type': 'unsigned'}
+  - {'name': 'name', 'type': 'string'}
+  - {'name': 'age', 'type': 'number'}
+  rows:
+  - [6, 'Alexey', 31]
+  - [4, 'Mikhail', 51]
+```
```
dokshina, line 341: All comments I've left for `pairs` relate to `select`.

crud/select/comparators.lua
```diff
@@ -127,15 +158,18 @@ function comparators.gen_func(cmp_operator, key_parts)
     return func
 end
 
-function comparators.gen_tuples_comparator(cmp_operator, key_parts)
-    local keys_comparator, err = comparators.gen_func(cmp_operator, key_parts)
+function comparators.gen_tuples_comparator(cmp_operator, key_parts, field_names, space_format)
```
dokshina, line 161: It would be great to add unit test for comparators to fix the behaviour. See https://github.com/tarantool/crud/blob/master/test/unit/select_comparators_test.lua.

crud/select/plan.lua
```diff
@@ -106,10 +107,70 @@ local function extract_sharding_key_from_scan_value(scan_value, scan_index, shar
...
```
dokshina, line 143: This function also needs unit test. It can be tested [here](https://github.com/tarantool/crud/blob/master/test/unit/select_plan_test.lua) as a part of `select.new` tests.

Moreover, now plan contains `field_names`, please add this check to plan tests.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 * Support for UUID field types and UUID values
 * `fields` option for simple operations to get partial result
+* `fields` option for select/pairs with pagination support
```
olegrok, line 20: Maybe it's reasonable to merge this two sections into one?

doc/pairs.md
```diff
@@ -209,12 +209,21 @@ tuples
...
+format = box.space.developers:format()
+format
+- {'name': 'id', 'type': 'unsigned'}
+- {'name': 'bucked_id', 'type': 'unsigned'}
+- {'name': 'name', 'type': 'string'}
+- {'name': 'surname', 'type': 'string'} 
+- {'name': 'age', 'type': 'number'}
+...
 objects = {}
 -- condition by indexed non-unique non-primary field
```
dokshina, line 228: Change to
`-- get names of users that are 31 year old or older`

doc/pairs.md
```diff
@@ -268,4 +287,17 @@ new_tuples
     - Mikhail
     - 51
 ...
```
dokshina, line 289: Add smth like `** THIS WOULD FAIL**` here. We need to highlight it as much as it's possible.

doc/select.md
```diff
@@ -338,4 +369,12 @@ res
...
   - [4, 'Mikhail', 51]
+...
+-- 'fields' isn't specified
+res, err = crud.select('developers', {{'>=', 'age', 27}})
+
+-- THIS WOULD FAIL
+-- call 'select' with 'fields' option specified 
+-- and pass to 'after' tuple that were got without 'fields' option
+res, err = crud.select('developers', {{'>=', 'age', 27}}, { fields = {'id', 'name'}, after = res.rows[1] })
 ```
```
dokshina, line 380: The same as for pairs.

dokshina (approved):
Brilliant!

PR #125, opened 2021-03-01
use is_async feature for calls

olegrok:
Maybe it's the most useful part of merger intgration patch. Before
this patch we spawn fiber for every netbox call. But after
"is_async" feature is introduced we could reduce CPU pressure with
fiber spawning and GC pressure when create channels.

Before patch:
```
SELECT
Fibers count - 200
Connection count - 10
Timeout  - 30.000000
Requests - 87503
Errors - 0
RPS - 2916.766667
```

After patch
```
SELECT
Fibers count - 200
Connection count - 10
Timeout  - 30.000000
Requests - 90827
Errors - 0
RPS - 3027.566667
```

### DIFF
--- a/crud/common/call.lua
+++ b/crud/common/call.lua
@@ -1,9 +1,9 @@
-local fiber = require('fiber')
 local vshard = require('vshard')
 local errors = require('errors')
 
 local dev_checks = require('crud.common.dev_checks')
 local utils = require('crud.common.utils')
+local fiber_clock = require('fiber').clock
 
 local CallError = errors.new_class('Call')
 local NotInitializedError = errors.new_class('NotInitialized')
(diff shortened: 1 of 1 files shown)


crud/common/call.lua
```diff
@@ -52,39 +32,42 @@ local function call_impl(vshard_call, func_name, func_args, opts)
         end
     end
 
-    local nodes_count = utils.table_count(replicasets)
-    local channel = fiber.channel(nodes_count)
-
+    local futures = {}
```
dokshina, line 35: Maybe `futures_by_replicasets`?

crud/common/call.lua
```diff
@@ -52,39 +32,42 @@ local function call_impl(vshard_call, func_name, func_args, opts)
...
-            if channel:is_closed() then
-                return nil, CallError:new("Channel is closed")
-            end
-
-            return nil, CallError:new("Timeout was reached")
+    local deadline = fiber_clock() + timeout
+    for replicaset_uuid, future in pairs(futures) do
+        local wait_timeout = deadline - fiber_clock()
+        if wait_timeout < 0 then
+            wait_timeout = 0
```
dokshina, line 47: Should we perform `future:wait_result(0)` call anyway, or simply return error?

dokshina (commented):
Thanks for your patch!

Please consider several comments.
BTW, I think that we can notice this change in Changelog.

olegrok:
> BTW, I think that we can notice this change in Changelog.

No, there are no any user-visible changes

crud/common/call.lua
olegrok: You should return special kind of box.error in this case box.error.timeout. I don't want to overcomplicate this logic - let's simply pass 0 to timeout.

crud/common/call.lua
olegrok: No problem. Fixed.

dokshina (approved):
Brilliant patch (again)!

PR #132, opened 2021-03-16
Add read strategy parameters opts

### LINKED ISSUE #124: [4pt] Add an option to choose callro/callre/callbro/callbre on read requests
callro - the best choise for default behaviour, but in some cases on highload we may need to balance read requests, or reduce requests to replicaset masters

dokshina:
Added `mode`, `prefer_replica` and `balance` options for read operations
(`get`, `select`, `pairs`). According to this parameters one of vshard
calls (`callrw`, `callro`, `callbro`, `callre`, `callbre`) is selected.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #124

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+* `mode`, `prefer_replica` and `balance` options for read operations
+  (get, select, pairs). According to this parameters one of vshard
+  calls (`callrw`, `callro`, `callbro`, `callre`, `callbre`) is selected
+
 ## [0.5.0] - 2021-03-10
 
 ### Fixed
--- a/crud/common/call.lua
+++ b/crud/common/call.lua
@@ -12,13 +12,69 @@ local call = {}
 
 local DEFAULT_VSHARD_CALL_TIMEOUT = 2
 
-local function call_impl(vshard_call, func_name, func_args, opts)
-    dev_checks('string', 'string', '?table', {
+local function get_vshard_call_name(mode, prefer_replica, balance)
+    dev_checks('string', '?boolean', '?boolean')
+
+    if mode == 'write' then
+        return 'callrw'
+    end
+
+    if not prefer_replica and not balance then
+        return 'callro'
+    end
+
+    if not prefer_replica and balance then
+        return 'callbro'
+    end
+
+    if prefer_replica and not balance then
+        return 'callre'
+    end
+
+    -- prefer_replica and balance
+    return 'callbre'
+end
+
+local function wrap_vshard_err(err, func_name, replicaset_uuid, bucket_id)
+    if err.type == 'ClientError' and type(err.message) == 'string' then
+        if err.message == string.format("Procedure '%s' is not defined", func_name) then
+            if func_name:startswith('_crud.') then
+                err = NotInitializedError:new("crud isn't initialized on replicaset: %q", replicaset_uuid)
+            else
+                err = NotInitializedError:new("Function %s is not registered", func_name)
+            end
+        end
+    end
+
+    if replicaset_uuid == nil then
+        local replicaset, _ = vshard.router.route(bucket_id)
+        if replicaset == nil then
+            return CallError:new(
+                "Function returned an error, but we couldn't figure out the replicaset: %s", err
+            )
+        end
+
+        replicaset_uuid = replicaset.uuid
+    end
+
+    err = errors.wrap(err)
+
+    return CallError:new(utils.format_replicaset_error(
+        replicaset_uuid, "Function returned an error: %s", err
+    ))
+end
+
+function call.map(func_name, func_args, opts)
+    dev_checks('string', '?table', {
+        mode = 'string',
+        prefer_replica = '?boolean',
+        balance = '?boolean',
         timeout = '?number',
         replicasets = '?table',
     })
-
     opts = opts or {}
+
+    local vshard_call_name = get_vshard_call_name(opts.mode, opts.prefer_replica, opts.balance)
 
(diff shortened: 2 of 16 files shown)


crud/common/call.lua
```diff
@@ -12,7 +12,59 @@ local call = {}
...
+        return 'callbro'
+    end
+
+    if prefer_replica and not balance then
+        return 'callre'
+    end
+
+    -- prefer_replica and balance
+    return 'callbre'
+end
```
olegrok, line 36: Sorry, I don't like it.
`vshard.router.map_callrw` (https://github.com/tarantool/vshard/issues/147) is coming soon. And it will be a single way to guarantee consistency when we perform map-reduce over all storages.
It will lead us to introducing new option?

So, anyway we should stay a way to perform "rw" call for read operations.

test/integration/read_calls_strategies_test.lua
```diff
@@ -0,0 +1,155 @@
+local fio = require('fio')
+
+local t = require('luatest')
+
+local helpers = require('test.helper')
+
+local pgroup = helpers.pgroup.new('read_calls_strategies', {
+    engine = {'memtx'},
+})
```
olegrok, line 9: Maybe you don't need parametrization at all?

test/integration/read_calls_strategies_test.lua
```diff
@@ -0,0 +1,155 @@
...
+        for _, func_name in ipairs(call_impl_funcs) do
+            call.internal.vshard_calls[func_name] = {}
+
+            local current_func = call.internal[func_name]
+            call.internal[func_name] = function(...)
+                local vshard_call = ...
+                table.insert(call.internal.vshard_calls[func_name], vshard_call)
+                return current_func(...)
+            end
+
```
olegrok, line 42: Maybe you can use `box.stat().CALL.total` for such purpose. I don't like approach with internals monkeypatching

olegrok (commented):
Hi! Thanks for your patch. I left several thoughts.

crud/common/call.lua
dokshina: Added `mode` option that can be `read` or `write`. `write` mode performs `rw`, `read` performs `callr*` according to `prefer_replica` and `balance`.

test/integration/read_calls_strategies_test.lua
dokshina: I've monkeypatched `router_mt`.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+* `prefer_replica` and `balance` options for read operations
```
olegrok, line 12: `mode`?

crud/common/call.lua
```diff
@@ -53,101 +113,33 @@ local function call_impl(vshard_call, func_name, func_args, opts)
...
-    local res, err = vshard.router.callrw(bucket_id, func_name, func_args, options)
-
-    -- This is a workaround, until vshard supports telling us where the error happened
-    if err ~= nil then
-        if type(err) == 'table' and err.type == 'ClientError' and type(err.message) == 'string' then
-            if err.message == string.format("Procedure '%s' is not defined", func_name) then
-               err = NotInitializedError:new("crud isn't initialized on replicaset")
-            end
-        end
+    local vshard_call_name = get_vshard_call_name(opts.mode, opts.prefer_replica, opts.balance, opts.mode)
```
olegrok, line 133: So... I'm not sure that you should do it in such way.

Seems you can use just vshard.router.call('name', {mode=, balance=, prefer_replica=}, args, opts).
So you don't need to perform such logic.

(https://www.tarantool.io/en/doc/latest/reference/reference_rock/vshard/vshard_api/#lua-function.vshard.router.call)

crud/select.lua
```diff
@@ -26,6 +26,14 @@ local SELECT_FUNC_NAME = '_crud.select_on_storage'
 
 local DEFAULT_BATCH_SIZE = 100
 
+local available_modes = {
+    read = true,
+    write = true,
+}
+function checkers.call_mode(p)
```
olegrok, line 33: `vshard_router_call_mode`?

crud/select.lua
```diff
@@ -26,6 +26,14 @@ local SELECT_FUNC_NAME = '_crud.select_on_storage'
...
+    return available_modes[p] ~= nil
```
olegrok, line 34: Maybe it's even faster to perform naive check: `mode == 'read' or mode == 'write'`

crud/common/call.lua
```diff
@@ -12,14 +12,74 @@ local call = {}
...
+
+    return CallError:new(utils.format_replicaset_error(
+        replicaset_uuid, "Function returned an error: %s", err
+    ))
+end
+
+function call.map(func_name, func_args, opts)
+    dev_checks('string', '?table', {
+        mode = 'string',
+        prefer_replica = '?boolean',
```
olegrok, line 74: Maybe `mode='write' + prefer_replica=true` should cause an error.

But feel free to ignore.

crud/common/call.lua
dokshina: It wouldn't work for `replicaset:call`, so we decided to don't change it now.

crud/common/call.lua
dokshina: `vshard` ignores `prefer_replica` and `balance` options if `write` mode is specified. Let's do the same.

crud/common/call.lua
```diff
@@ -12,14 +12,74 @@ local call = {}
 
 local DEFAULT_VSHARD_CALL_TIMEOUT = 2
 
-local function call_impl(vshard_call, func_name, func_args, opts)
-    dev_checks('string', 'string', '?table', {
+
+-- for testing
+call.internal = {}
```
olegrok, line 17: Is it still needed?

olegrok (approved):

Issue #135, opened 2021-03-25
CRUD not select information from vshard


seet61:
We have cartridge application.

```
dependencies = {
        'tarantool == 2.5.3',
        'lua >= 5.1',
        'checks',
    'cartridge == 2.5.0',
    'metrics == 0.7.0',
    'cartridge-cli-extensions',
    'crud == 0.5.0',
    'http',
    'gis',
}
```

At storage role we have space:
```
        -- authTemplates
        local authTemplates_space = box.schema.space.create('authTemplates', {
            engine = 'memtx',
            if_not_exists = true,
        })
        log.info('authTemplates space was configured')

        authTemplates_space:format({
            {name='msisdn',type='string'},
            {name='channel',type='string'},
            {name='password',type='string'},
            {name='counter',type='unsigned'},
            {name='create_date',type='unsigned'},
            {name='ttl',type='unsigned'},
            {name='bucket_id', type='unsigned'},
        })
        log.info('authTemplates was formatted')

        authTemplates_space:create_index('authTemplates_msisdn_channel_idx', {parts={{field='msisdn'}, {field='channel'}},
            type = 'HASH',
            if_not_exists=true})
        log.info('authTemplates_msisdn_channel_idx')

        authTemplates_space:create_index('bucket_id', {parts={{field='bucket_id'}},
            unique=false,
            if_not_exists=true})
        log.info('authTemplates_bucket_id')

        authTemplates_space:create_index('ttl', {parts={{field='ttl'}},
            unique=false,
            if_not_exists=true})
        log.info('authTemplates_ttl')
```

Topology:
We have 2 servers.
At each server started 3 roles: router, master and replica.

3 replica set: for router and 2 storages. Master storage reserving by replica from another server.
![image](https://user-images.githubusercontent.com/6140289/112429891-6536fc80-8d4e-11eb-8225-f172ff51d901.png)

Saving this entry by crud:
```
{"1":"79774120882","2":"app4t2","3":"671418","4":3,"5":1716591611,"6":1716592211,"space":"authTemplates"}


crud.insert('authTemplates', {"1":"79774120882","2":"app4t2","3":"671418","4":3,"5":1716591611,"6":1716592211}, {timeout = 5})
```

After that we trying to get this information from first router:
```
localhost:3301> crud.select('authTemplates', {{'==', 'authTemplates_msisdn_channel_idx', {'79774120882', 'app4t2'}}}, {timeout = 5})
---
[...]

PR #142, opened 2021-04-02
Add test for 135 issue, simple select from two routers

filonenko-mikhail:

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Closes #???


olegrok:
Issue https://github.com/tarantool/crud/issues/135 is closed because it doesn't related to crud at all. The problem is fixed in metrics 0.9.0.

Issue #140, opened 2021-03-31
[1pt] filter_tuple_fields returns error on attempt to extract null field


olegrok:
https://github.com/tarantool/crud/blob/c961e166645aeca897895f8513cc71e6a69c408c/crud/common/schema.lua#L147-L151

1. Such check is true if field contain box.NULL. It could be solved with `type(field) ~= 'nil'`.
1. Such check is true if I add some new nullable field at the end of tuple. There is such field in space format but physically this field absent. It could not be solved with `type(field) ~= 'nil'`.

PR #143, opened 2021-04-06
Cut result returned by select

### LINKED ISSUE #128: [1pt] Add a function to "cut" result returned by select
Since #122 it's possible to get a part of fields. 
BUT scan key and primary key are always merged to result (for correct pagination). 
We need to introduce a function to cut unnecessary fields, smth like:

```lua
res, err = crud.select('developers', {{'>=', 'age', 27}}, { fields = {'id', 'name'} })
res
- metadata:
  - {'name': 'id', 'type': 'unsigned'}
  - {'name': 'name', 'type': 'string'}
  - {'name': 'age', 'type': 'number'}
  rows:
  - [3, 'Pavel', 27]
  - [6, 'Alexey', 31]
  - [4, 'Mikhail', 51]

crud.cut_rows(rows, {'id', 'name'})
- metadata:
  - {'name': 'id', 'type': 'unsigned'}
  - {'name': 'name', 'type': 'string'}
  rows:
  - [3, 'Pavel']
  - [6, 'Alexey']
  - [4, 'Mikhail']
```

AnaNek:
Using select with `field` option we have unnecessary fields in select result for correct pagination. 
This patch introduces a function to cut unnecessary fields.

Closes #128


crud.lua
```diff
@@ -69,6 +70,10 @@ crud.unflatten_rows = utils.unflatten_rows
 -- @function truncate
 crud.truncate = truncate.call
 
+-- @refer cut_rows.call
+-- @function cut_rows
+crud.cut_rows = cut_rows.call
```
olegrok, line 75: Maybe it's better to place this function directly in `utils`? In the same way as for `unflatten_rows `.

crud/common/schema.lua
```diff
@@ -176,6 +176,19 @@ function schema.filter_tuples_fields(tuples, field_names)
     return result
 end
 
+function schema.truncate_row_trailing_fields(tuple, field_names)
```
olegrok, line 179: Seems it won't work for box.tuple (cdata).

crud/cut_rows.lua
```diff
@@ -0,0 +1,32 @@
...
+
+    if err ~= nil then
+        return nil, err
+    end
+
+    local truncated_rows = {}
+
+    for _, row in ipairs(select_result.rows) do
+        local truncated_row = schema.truncate_row_trailing_fields(row, field_names)
+        table.insert(truncated_rows, truncated_row)
```
olegrok, line 23: Since this function performs in-place modification (it modifies user data).
Maybe it's better to eliminate `truncated_rows ` creation. Just return `select_result.rows`.

crud/common/schema.lua
olegrok: FYI, merger:pairs() returns tuples

crud/common/schema.lua
```diff
@@ -135,6 +135,25 @@ local function get_space_schema_hash(space)
...
+
+    local result = {}
+
+    for _, field_name in ipairs(field_names) do
+        result[field_name] = obj[field_name]
+        if result[field_name] == nil then
+            return nil, FilterFieldsError:new(
+                    'Object doesn\'t contain field named %q', field_name
+            )
+        end
```
olegrok, line 151: Will it work for nullable fields?

See similar issue - https://github.com/tarantool/crud/issues/140

crud/common/utils.lua
```diff
@@ -437,6 +441,45 @@ function utils.truncate_tuple_metadata(tuple_metadata, field_names)
...
+
+    opts = opts or {}
+
+    local truncated_metadata, err = truncate_tuple_metadata(metadata, field_names)
+
+    if err ~= nil then
+        return nil, err
+    end
+
+    if opts.mapped then
```
olegrok, line 457: `cut_rows(..., {mapped = box.NULL})` will be considered as true.

crud/common/utils.lua
```diff
@@ -437,6 +441,45 @@ function utils.truncate_tuple_metadata(tuple_metadata, field_names)
...
+            if err ~= nil then
+                return nil, err
+            end
+
+        end
+
+        return {
+            metadata = truncated_metadata,
+            rows = rows,
+        }
```
olegrok, line 470: It's the same as below. You could remove this part and place  `chema.truncate_row_trailing_fields` under `else` branch.

crud/common/schema.lua
```diff
@@ -144,7 +144,7 @@ function schema.filter_obj_fields(obj, field_names)
 
     for _, field_name in ipairs(field_names) do
         result[field_name] = obj[field_name]
-        if result[field_name] == nil then
+        if type(result[field_name]) == nil then
```
olegrok, line 147: This condition is always false. Because type(...) returns a string. It could return `"nil"` (as string). But anyway as I mentioned in issue https://github.com/tarantool/crud/issues/140 this check won't protect you from cases when nullable field is not box.NULL it's just `nil`. It's possible to omit trailing nils in tuples or user could add some nullable fields at the end of a tuple during migration. My proposal is to remove this check at all.

crud/common/utils.lua
```diff
@@ -413,6 +413,68 @@ function utils.format_result(rows, space, field_names)
...
+
+    opts = opts or {}
+
+    local truncated_metadata, err = truncate_tuple_metadata(metadata, field_names)
+
+    if err ~= nil then
+        return nil, err
+    end
+
+    if type(opts.mapped) ~= 'cdata' and opts.mapped then
```
olegrok, line 457: Just add `opts.mapped == true`. Please don't overcomplicate :)

crud/common/utils.lua
```diff
@@ -413,6 +413,63 @@ function utils.format_result(rows, space, field_names)
...
+            )
+        end
+
+        table.insert(truncated_metadata, tuple_metadata[i])
+    end
+
+    return truncated_metadata
+end
+
+function utils.cut_rows(rows, metadata, field_names, opts)
```
olegrok, line 444: If it's public method you should provide some documentation.
Especially if you a refer this part above https://github.com/tarantool/crud/pull/143/files#diff-24b635b35abaadcc2951d808149b57dcb8fe2325d9b08d2c861eaffed77ce167R72-R74

olegrok (approved):
Please, provide changelog entry and documentation for new method (arguments part is the most interesting). Also I concern about function name (maybe it's "truncate", not "cut" IDK). @dokshina please also take a look.

dokshina (commented):
Thanks, @olegrok!

Please, add a doc and update Changelog.

test/integration/pairs_test.lua
```diff
@@ -543,3 +543,49 @@ pgroup:add('test_pairs_partial_result', function(g)
...
+        {name = "William", city = "Chicago"},
+    }
+
+    local objects = g.cluster.main_server.net_box:eval([[
+        local crud = require('crud')
+
+        local conditions, fields = ...
+
+        local objects = {}
+        for _, object in crud.pairs('customers', conditions, {use_tomap = true, fields = fields}) do
```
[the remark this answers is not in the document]
AnaNek: Will pagination work if we cut off the fields inside `pairs`?

README.md
```diff
@@ -390,6 +391,37 @@ end
...
+* `fields` (`table`) - field names of fields that should be contained in the result
+
+Returns metadata and array of rows, error.
+
+See more examples of `crud.cut_rows` usage [here](https://github.com/tarantool/crud/blob/master/doc/select.md) and [here.](https://github.com/tarantool/crud/blob/master/doc/pairs.md)
+
+### Cut_objects
+
+```lua
+local new_objects = crud.cut_rows(objects, fields)
```
dokshina, line 413: `crud.cut_rows` -> `crud.cut_objects`

README.md
```diff
@@ -390,6 +391,37 @@ end
 
 See more examples of pairs queries [here.](https://github.com/tarantool/crud/blob/master/doc/pairs.md)
 
+### Cut_rows
```
dokshina, line 394: `Cut extra rows`? I think that it would be good to add a note that this function is useful with `fields` parameter.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 * Fixed error for partial result option if field contains box.NULL.
 
+### Added
+
+* `cut_rows` and `cut_objects` functions to cut off scan key and primary key values that were merged to the select/pairs partial result.
```
dokshina, line 16: Generally, not only select/pairs.

... merged to the result of operations with `fields` specified?

dokshina (approved):
Thanks for your patch!

Please, consider one comment before merge.

PR #145, opened 2021-04-09
Hot-reload work check

### LINKED ISSUE #107: [1pt] Verify that cartridge hot-reload works
Verify (and fix if it's not so) cartridge roles (and crud as separate module) supports hot-reload.

See also https://github.com/tarantool/cartridge/pull/1105

mRrvz:
Added hot-reload health test. 
Added ``stop()`` functions for ``crud`` roles.
Closes #107 #148

### DIFF
--- a/crud.lua
+++ b/crud.lua
@@ -103,12 +103,11 @@ function crud.init_router()
    rawset(_G, 'crud', crud)
 end
 
-
-function crud.router_stop()
+function crud.stop_router()
     rawset(_G, 'crud', nil)
 end
 
-function crud.storage_stop()
+function crud.stop_storage()
     rawset(_G, '_crud', nil)
 end
 
--- a/crud.lua
+++ b/crud.lua
@@ -103,4 +103,13 @@ function crud.init_router()
    rawset(_G, 'crud', crud)
 end
 
+
+function crud.router_stop()
+    rawset(_G, 'crud', nil)
+end
+
+function crud.storage_stop()
+    rawset(_G, '_crud', nil)
+end
+
 return crud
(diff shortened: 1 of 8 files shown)


rosik:
I'd suggest cleaning up globals `_G.crud` and `_G._crud`:

https://github.com/tarantool/crud/blob/c961e166645aeca897895f8513cc71e6a69c408c/crud.lua#L79-L96

You can use cartridge `stop` callback. See for example https://github.com/tarantool/cartridge/blob/master/cartridge/roles/vshard-storage.lua#L48

no1seman:
There is another obe case that must be checked: during roles reload version of crud (and also all lua-files) may be changed on disk. Test must check, that new version fully reloaded.

crud.lua
```diff
@@ -95,4 +95,13 @@ function crud.init_router()
    rawset(_G, 'crud', crud)
 end
 
+
+function crud.router_stop()
```
dokshina, line 99: But you use `crud.stop_router()` above/

test/integration/reload_test.lua
```diff
@@ -0,0 +1,141 @@
...
+    end)
+
+    local cnt = #g.insertions_passed
+    local ok = pcall(g.cluster.retrying, g.cluster, {timeout = 2}, function()
+        helpers.assert_ge(#g.insertions_passed, cnt+1)
+    end)
+
+    if not ok then
+        g.highload_fiber:cancel()
+        fiber.sleep(10000)
```
dokshina, line 133: ???

dokshina (approved):

Issue #120, opened 2021-02-17
Fetching nested structures in tuple fields


AnaNek:
> What about "bad input" tests? What if user specified some bad field name?

I believe that we should do as Tarantool does:
```lua
tarantool> box.tuple.new({1, {a = {b = 'c'}}})
---
- [1, {'a': {'b': 'c'}}]
...

tarantool> box.tuple.new({1, {a = {b = 'c'}}})['[1]']
---
- 1
...

tarantool> box.tuple.new({1, {a = {b = 'c'}}})['[2]']
---
- {'a': {'b': 'c'}}
...

tarantool> box.tuple.new({1, {a = {b = 'c'}}})['[2].a.b']
---
- c
...
tarantool> box.tuple.new({1, {a = {b = 'c'}}})['K']
---
- null
...
```

This makes this feature quite powerful and allows to fetch not only some tuple fields but some nested structures in tuple fields.

_Originally posted by @olegrok in https://github.com/tarantool/crud/pull/118#discussion_r576381992_

Totktonada:
Am I understood right: the proposal is to support JSON path in `fields` option of crud operations and in `crud.cut_rows()` and `crud.cut_objects()` function?

olegrok:
> the proposal is to support JSON path in fields option of crud operations

Yes

> in crud.cut_rows() and crud.cut_objects() function

Initially there are no such functions in public API. However since currently it's so it should be done for consistency.

Totktonada:
It's hard to evaluate, but I don't hear requests about this feature from customers, the product team or project developers (except you). I guess that the demain is small.

The implementation would either depend on https://github.com/tarantool/tarantool/issues/5203 (which is not implemented at the moment of writing) or would use an external module like https://github.com/olegrok/tarantool-jsonpath (which has pros and cons).

I'll put the task to the wishlist. It means that if someone is ready to work on it, we're open for design discussions (let's start them here) and pull requests. Of course, we can accept the implementation only if it is carefully designed, implemented, documented and tested.

The priority may raise in a future, if something important will depend on the feature.

DifferentialOrange:
It's not clear how to work with such API. For now, asking a tuple with some fields would return the following row: `{field_1, nil, nil, field_4, nil, field_6}`. But what row should be returned for `[2].a.b` field? We can't return `[2].a.b` in the place of `2` since there could be `'[2].a.b'` and `'[2].c.d'` fields at the same time.

DifferentialOrange:
[...]

PR #147, opened 2021-04-12
Partial result bug with extracting nullable fields

### LINKED ISSUE #140: [1pt] filter_tuple_fields returns error on attempt to extract null field

AnaNek:
FilterTupleField check is true if field contain box.NULL. In this case we get an error. We get an error when there are nullable fields at the end of a tuple. But this cases are valid and should be performed without this error.

Closes #140

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+* Fixed error for partial result option if field contains box.NULL.
+
 ## [0.6.0] - 2021-03-29
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


olegrok:
Thanks for your patch! This PR should also resolve https://github.com/tarantool/crud/issues/120. Please add "Closes #120" to description and add some tests for this case.

dokshina (commented):
Nice patch. 
Don't forget about Changelog!

dokshina (approved):
Good!

PR #151, opened 2021-04-21
Optional flag to disable the bucket_id computation from primary key

### LINKED ISSUE #136: Add an optional flag to disable the bucket_id computation from primary key.
> By default, bucket_id is computed as vshard.router.bucket_id_strcrc32(key), where key is the primary key value. Custom bucket ID can be specified as opts.bucket_id for each operation. For operations that accepts tuple/object bucket ID can be specified as tuple/object field as well as opts.bucket_id value.

Please add an optional flag to disable the bucket_id computation.
Map-reduce select is impossible if primary key used in condition.
I want select to be called on each storage with primary key condition.

AnaNek:
Closes #136


test/helper.lua
```diff
@@ -192,4 +192,49 @@ function helpers.assert_ge(actual, expected, message)
     end
 end
 
+function helpers.get_replicasets_with_equal_key(cluster, key)
```
dokshina, line 195: Seems that this function can be replced by `get_other_storage_bucket_id` function from `custom_bucket_id_test`.

Moreover, `get_other_storage_bucket_id` name looks better.
This function does get replica sets, but in fact, only one of them corresponds to the specified key.
The main idea of this function is to get bucket_id that corresponds to other replicas than the specified key (in fact, bucket_id).

I think that this function can be common for your test and `custom_bucket_id_test`.
The signature can be changed to `get_other_storage_bucket_id(cluster, bucket_id)` since you need to use both bucket IDs.

crud/select/compat/select.lua
```diff
@@ -23,6 +23,7 @@ local function build_select_iterator(space_name, user_conditions, opts)
         first = '?number',
         batch_size = '?number',
         bucket_id = '?number|cdata',
+        bucket_optimization = '?boolean',
```
dokshina, line 26: I don't like this option name.
I would like it to be "negative". Like "don't use optimization".  @olegrok, what do you think?

But if you left this flag "positive", you can simply use `opts.bucket_optimization != false` instead of setting it to `true` if it's `nil`.

dokshina (commented):
Don't forget about documentation and changelog.

crud/select/compat/select.lua
olegrok: To be honest I also don't like this option name. Also amount of already existed options makes me frustrating.

I suggest to reconsider approach to solution of this problem.

My thoughts:
  * Add option that states that following call will we a map call (I also don't like it. It's still one more option and it looks like a crutch).
  * Take 2 steps back to take a step forward. Initially such optimization protects user from redundant network hops if it's possible to calculate bucket_id from user defined filters. If user uses custom sharding it would be great to consider it - we need to provide a way to change rules of calculation of bucket_id. It's a bit complex and ambitious feature but I like it.
  * Allow to put "box.NULL" to bucket_id option to block bucket_id calculation (It makes `box.NULL` a kind system/special value. And I also hates such approaches. But it could be solution).

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * `cut_rows` and `cut_objects` functions to cut off scan key and
   primary key values that were merged to the select/pairs partial result.
 * Functions ``stop()`` for the roles ``crud-storage`` and ``crud-router``.
+* Option flag `block_bucket_id_computation` for `select()`/`pairs()` to disable the `bucket_id` computation from primary key.
```
mRrvz, line 19: Please, make the lines shorter in the Changelog.

dokshina:
I suggest renaming `block_bucket_id_computation` to `force_map_call`.
Don't forget to rename corresponding tests.

README.md
```diff
@@ -321,6 +321,8 @@ where:
   * `after` (`?table`) - tuple after which objects should be selected
   * `batch_size` (`?number`) - number of tuples to process per one request to storage
   * `bucket_id` (`?number|cdata`) - bucket ID
+  * `force_map_call` (`?boolean`) - if `true` 
+    then the `bucket_id` computation from primary key is disabled
```
dokshina, line 325: Maybe `... then the map call is performed without any optimizations even if full primary key equal condition is specified?`

dokshina (approved):
Thanks for your patch.

Consider one comment in the doc before merging.

PR #155, opened 2021-05-24
Fix reverse pagination incorrect result

mRrvz:
``crud.select`` query using reverse pagination and without
specifying ``batch_size``, the [``limit``](https://github.com/tarantool/crud/blob/144f9035d44ceb242a3bf664e41d553690133dc2/crud/select/compat/select.lua#L94) became negative, 
which causes errors.

### DIFF
--- a/crud/select/compat/select.lua
+++ b/crud/select/compat/select.lua
@@ -75,16 +75,16 @@ local function build_select_iterator(space_name, user_conditions, opts)
         end
     end
 
-    local first = opts.first
-    if first ~= nil then
-        first = math.abs(opts.first)
+    local tuples_limit = opts.first
+    if tuples_limit ~= nil then
+        tuples_limit = math.abs(tuples_limit)
     end
 
-    -- If opts.batch_size is missed we should specify it to min(first, DEFAULT_BATCH_SIZE)
+    -- If opts.batch_size is missed we should specify it to min(tuples_limit, DEFAULT_BATCH_SIZE)
     local batch_size
     if opts.batch_size == nil then
-        if first ~= nil and first < common.DEFAULT_BATCH_SIZE then
-            batch_size = first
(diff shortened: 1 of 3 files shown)


crud/select/compat/select.lua
```diff
@@ -79,7 +79,7 @@ local function build_select_iterator(space_name, user_conditions, opts)
     local batch_size
     if opts.batch_size == nil then
         if opts.first ~= nil and opts.first < common.DEFAULT_BATCH_SIZE then
```
dokshina, line 81: If `opts.first` is negative, `opts.first < common.DEFAULT_BATCH_SIZE` is always true.

crud/select/compat/select.lua
```diff
@@ -75,6 +75,10 @@ local function build_select_iterator(space_name, user_conditions, opts)
         end
     end
 
+    if opts.first ~= nil then
+        opts.first = math.abs(opts.first)
```
olegrok, line 79: It's a side-effect. You modifies user's variable. Probably it's better to save "first" into separate variable and work with it.

crud/select/compat/select.lua
```diff
@@ -75,11 +75,16 @@ local function build_select_iterator(space_name, user_conditions, opts)
         end
     end
 
+    local first = opts.first
+    if first ~= nil then
+        first = math.abs(opts.first)
+    end
```
olegrok, line 81: Probably we could declare `local first` a bit above - https://github.com/tarantool/crud/pull/155/files#diff-b78f92602289b081e8f7213e3f7f546dd3148362696b13715667d1e795262a6fR56

Also `first = math.abs(first)` seems to be better than `first = math.abs(opts.first)`.

olegrok (approved):

dokshina (approved):

PR #158, opened 2021-05-26
Select: jsonpath indexes

### LINKED ISSUE #12: [2pt] Support JSON path
Even Tarantool 1.10 supports value extraction via jsonpath. So we could patch our filter generator and add jsonpath's support of jsonpath conditions.

mRrvz:
Added jsonpath indexes support for queries. Closes #12

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Ignoring `opts.first` on `crud.pairs` call
 * External `keydef` compatibility with built-in `merger`
 
+### Added
+
+* Added support for jsonpath indexes for ``crud.select``
+
 ## [0.7.0] - 2021-05-27
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-* Added support for jsonpath indexes for ``crud.select``
+* Added jsonpath indexes support for queries.
 
 ## [0.7.0] - 2021-05-27
 
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -152,10 +152,6 @@ function utils.extract_key(tuple, key_parts)
 end
 
 function utils.extract_jsonpath_keys(tuple, index_parts, condition_map)
-    if tuple == nil then
-        return nil
-    end
-
     local key_def = nil
     local extracted_values
     local curr_jsonpath_ind = nil
@@ -568,15 +564,6 @@ end
 
 function utils.flatten_obj_reload(space_name, obj)
     return schema.wrap_func_reload(flatten_obj, space_name, obj)
-end
-
-function utils.get_index_paths(index_parts)
-    local parts = {}
-    for _, v in ipairs(index_parts) do
-        table.insert(parts, v.path)
-    end
-
-    return parts
 end
 
 function utils.get_condition_values_map(conditions)
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -12,9 +12,6 @@ local ShardingError = errors.new_class('ShardingError',  {capture_stack = false}
 local GetSpaceFormatError = errors.new_class('GetSpaceFormatError',  {capture_stack = false})
 local FilterFieldsError = errors.new_class('FilterFieldsError',  {capture_stack = false})
 
-local compat = require('crud.common.compat')
-local keydef_lib = compat.require('tuple.keydef', 'key_def')
-
 local utils = {}
 
 local space_format_cache = setmetatable({}, {__mode = 'k'})
@@ -152,20 +149,26 @@ function utils.extract_key(tuple, key_parts)
 end
 
 function utils.extract_jsonpath_keys(tuple, index_parts, key_def)
-    key_def = key_def or keydef_lib.new(index_parts)
+    local keys = {}
 
     -- We have case, when we get already extracted values in
     -- tuple variable, and we will get error like this:
     -- 'Tuple field [5] required by space format is missing'.
-    local ok, extracted_values = pcall(function()
-        return key_def:extract_key(tuple)
-    end)
-
-    if ok == false then
-        return tuple
+    -- For example, we can get tuple {'Yellow'} (already extrated jsonpath key)
+    -- and index parts like this: {{fieldno = 3, path = data.color}}.
+    -- When we try to key_def:extract_key, we wil get error.
+    for _, part in ipairs(index_parts) do
+        if part.path ~= nil and type(tuple[part.fieldno]) == 'table' then
+            keys = key_def:extract_key(tuple)
+            break
+        end
     end
 
-    return extracted_values
+    if #keys == 0 then
+        return utils.extract_key(tuple, index_parts)
+    end
+
+    return keys
 end
 
 function utils.merge_primary_key_parts(key_parts, pk_parts)
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -12,6 +12,9 @@ local ShardingError = errors.new_class('ShardingError',  {capture_stack = false}
 local GetSpaceFormatError = errors.new_class('GetSpaceFormatError',  {capture_stack = false})
 local FilterFieldsError = errors.new_class('FilterFieldsError',  {capture_stack = false})
 
+local compat = require('crud.common.compat')
+local keydef_lib = compat.require('tuple.keydef', 'key_def')
+
 local utils = {}
 
 local space_format_cache = setmetatable({}, {__mode = 'k'})
@@ -145,6 +148,49 @@ function utils.extract_key(tuple, key_parts)
     for i, part in ipairs(key_parts) do
         key[i] = tuple[part.fieldno]
     end
+    return key
+end
+
+function utils.extract_jsonpath_keys(tuple, index_parts, condition_map)
+    if tuple == nil then
+        return nil
+    end
+
+    local key_def = nil
+    local extracted_values
+    local curr_jsonpath_ind = nil
+    local key = {}
+
+    for i, part in ipairs(index_parts) do
+        if part.path ~= nil and type(tuple[part.fieldno]) == 'table' then
+            if curr_jsonpath_ind == nil then
+                curr_jsonpath_ind = i
+            end
+
+            if key_def == nil then
+                key_def = keydef_lib.new(index_parts)
+                extracted_values = key_def:extract_key(tuple)
+            end
+
+            -- We need this check (and also condition map) for composite
+            -- jsonpath indexes. For example, we have to filtering values
+            -- by partial jsonpath index, but index is composite and fields
+            -- are jsonpaths to the same space field. In this case, we
+            -- have to extract only needed value, specified in condtion.
+            if condition_map ~= nil then
+                if condition_map[extracted_values[curr_jsonpath_ind]] then
+                    key[i] = extracted_values[curr_jsonpath_ind]
+                end
+            else
+                key[i] = extracted_values[curr_jsonpath_ind]
+            end
+
+            curr_jsonpath_ind = curr_jsonpath_ind + 1
+        else
+            key[i] = tuple[part.fieldno]
+        end
+    end
+
     return key
 end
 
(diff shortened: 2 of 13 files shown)


crud/common/utils.lua
```diff
@@ -148,6 +151,49 @@ function utils.extract_key(tuple, key_parts)
     return key
 end
 
+function utils.extract_jsonpath_keys(tuple, index_parts, condition_map)
+    if tuple == nil then
+        return nil
+    end
```
olegrok, line 157: Is it possible? tuple is nil looks like an error.

crud/common/utils.lua
```diff
@@ -512,4 +570,26 @@ function utils.flatten_obj_reload(space_name, obj)
...
 end
 
+function utils.get_index_paths(index_parts)
+    local parts = {}
+    for _, v in ipairs(index_parts) do
+        table.insert(parts, v.path)
+    end
+
+    return parts
+end
```
olegrok, line 580: This function seems to be never used

crud/common/utils.lua
```diff
@@ -148,6 +151,49 @@ function utils.extract_key(tuple, key_parts)
...
+                end
+            else
+                key[i] = extracted_values[curr_jsonpath_ind]
+            end
+
+            curr_jsonpath_ind = curr_jsonpath_ind + 1
+        else
+            key[i] = tuple[part.fieldno]
+        end
+    end
```
olegrok, line 192: I don't get why this part is needed. key_def.new(...) + key_def:extact_key/key_def:compare should be enough.

olegrok (commented):
Thanks for your patch. See several comments below.

crud/select/executor.lua
```diff
@@ -49,19 +50,6 @@ function executor.execute(space, index, filter_func, opts)
     local tuples_count = 0
 
     local value = opts.scan_value
-    if opts.after_tuple ~= nil then
```
olegrok, line 52: No, we shouldn't do it. I will cause some performance issues.

olegrok: See https://github.com/tarantool/crud/pull/169/files#r652086030

crud/common/utils.lua
```diff
@@ -148,6 +151,11 @@ function utils.extract_key(tuple, key_parts)
     return key
 end
 
+function utils.extract_jsonpath_keys(tuple, index_parts)
+    local key_def = keydef_lib.new(index_parts)
+    return key_def:extract_key(tuple)
+end
```
olegrok, line 157: It's a bit weird to create key_def for each(!) comparison.

crud/compare/comparators.lua
```diff
@@ -153,8 +154,8 @@ function comparators.gen_tuples_comparator(cmp_operator, key_parts, field_names,
     local keys_comparator = comparators.gen_func(cmp_operator, updated_key_parts)
 
     return function(lhs, rhs)
```
olegrok, line 156: You can create keydef instance here if you really need it.

crud/common/utils.lua
```diff
@@ -148,6 +151,23 @@ function utils.extract_key(tuple, key_parts)
     return key
 end
 
+function utils.extract_jsonpath_keys(tuple, index_parts, key_def)
+    key_def = key_def or keydef_lib.new(index_parts)
```
olegrok, line 155: I don't like such logic. Why can't you use just key_def instance?

Anyway such logic won't prevent a case when you'll recreate key_def instance if the third argument is skipped.

crud/common/utils.lua
```diff
@@ -148,6 +151,23 @@ function utils.extract_key(tuple, key_parts)
...
+
+    -- We have case, when we get already extracted values in
+    -- tuple variable, and we will get error like this:
+    -- 'Tuple field [5] required by space format is missing'.
+    local ok, extracted_values = pcall(function()
+        return key_def:extract_key(tuple)
+    end)
```
olegrok, line 162: I don't understand you. It looks like a crutch. Also you will recreate closure for each call. It's too expensive for such simple logic. Usage of pcall is too expensive as well.

olegrok:
I suggest to rebase create some temporary branch based on https://github.com/tarantool/crud/tree/perf-test and verify your patch against performance degradation.

crud/select/executor.lua
```diff
@@ -53,10 +57,10 @@ function executor.execute(space, index, filter_func, opts)
         if value == nil then
             value = opts.after_tuple
         else
-            local cmp_operator = select_comparators.get_cmp_operator(opts.tarantool_iter)
-            local scan_comparator = select_comparators.gen_tuples_comparator(cmp_operator, index.parts)
-            local after_tuple_key = utils.extract_key(opts.after_tuple, index.parts)
-            if scan_comparator(after_tuple_key, opts.scan_value) then
+            local key_def = keydef_lib.new(index.parts)
+            local after_tuple_key = key_def:extract_key(opts.after_tuple)
```
olegrok, line 61: Please move this line under `key_def:compare_with_key(opts.after_tuple, opts.scan_value)` - you don't need to extract anything if condition below will be false.

 Also I `== 0` means "equal" I but `scan_comparator(after_tuple_key, opts.scan_value)` returned `true` in case of "less than". So I'm not sure that your changes are correct.

CHANGELOG.md
```diff
@@ -7,9 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-### Fixed
-
 * Compatibility with Tarantool versions that don't support key_def and merger modules.
```
olegrok, line 10: `Compatibility with Tarantool versions that don't support key_def and merger modules.` should be in `Fixed` section. `Added support for jsonpath indexes for ...` should be in `Added`. Also I think we don't need to specify crud.select directly... because there is `crud.pairs`.

crud/select/filters.lua
```diff
@@ -40,11 +40,15 @@ local function is_early_exit_possible(index, tarantool_iter, condition)
...
 end
 
-local function get_index_fieldnos(index)
+local function get_index_fieldnos(index, space_format)
     local index_fieldnos = {}
 
     for _, part in ipairs(index.parts) do
-        table.insert(index_fieldnos, part.fieldno)
+        if part.path ~= nil then
+            table.insert(index_fieldnos, string.format("%s.%s", space_format[part.fieldno].name, part.path))
```
olegrok, line 48: Hmmm... Could you please check some complex "path". I assume that path could contain spaces and special characters... Probably correct variant is `string.format(`[%q][%q]`, space_format[part.fieldno].name, part.path))`

olegrok: Also I think you don't need `space_format` here.  `part` contains field (or fieldno) and it should be enough.

mRrvz: If we get ``part.path`` = car.model and ``fieldno`` = 5 (part doesn't contain``field``), we'll get ["5"]["car.model"]. We can split ``part.path`` by point, but I don't that's a gooid idea...

olegrok: Maybe such way could consider each possible combination...

```lua
tarantool> t = box.tuple.new({1, 2, {car = {model = 'bmw'}}})
---
...

tarantool> t['[3]car.model']
---
- bmw
...

tarantool> t['[3].car.model']
---
- bmw
...

tarantool> t['[3]["car"]["model"]']
---
- bmw
...
```

Anyway suggested solution seems doesn't solve a problem with such paths:
```lua
tarantool> box.space.space:create_index('json2', {parts = {{field = 2, path = '["hello world"]', type = 'string'}}})
---
- unique: true
  parts:
  - type: string
    is_nullable: false
    path: '["hello world"]'
    fieldno: 2

tarantool> box.space.space:create_index('json', {parts = {{field = 2, path = '["car"]["model"]', type = 'string'}}})
---
- unique: true
  parts:
  - type: string
    is_nullable: false
    path: '["car"]["model"]'
    fieldno: 2
```

crud/select/filters.lua
```diff
@@ -44,7 +44,11 @@ local function get_index_fieldnos(index)
     local index_fieldnos = {}
 
     for _, part in ipairs(index.parts) do
-        table.insert(index_fieldnos, part.fieldno)
+        if part.path ~= nil then
+            table.insert(index_fieldnos, string.format("[%d]%s", part.fieldno, part.path))
+        else
+            table.insert(index_fieldnos, part.fieldno)
+        end
```
olegrok, line 51: I think this change should be covered with test https://github.com/tarantool/crud/blob/master/test/unit/select_filters_test.lua

olegrok (commented):
Thanks for your fixes. Consider last (I hope) comment below. And fix backward compatibility - I see that test for 1.10.6-1-g52c786b fails.

crud/compare/comparators.lua
```diff
@@ -150,12 +154,23 @@ function comparators.gen_tuples_comparator(cmp_operator, key_parts, field_names,
     local updated_key_parts = comparators.update_key_parts_by_field_names(
             space_format, field_names, key_parts
     )
+
     local keys_comparator = comparators.gen_func(cmp_operator, updated_key_parts)
 
+    local key_def
+    if has_keydef then
+        key_def = keydef_lib.new(updated_key_parts)
+    end
```
olegrok, line 163:
```lua
if has_keydef then
    local key_def = keydef_lib.new(updated_key_parts)
    return function(lhs, rhs)
          local lhs_key = key_def:extract_key(lhs)
          local rhs_key = key_def:extract_key(rhs)
          return keys_comparator(lhs_key, rhs_key)
    end
else
    return function(lhs, rhs)
        local lhs_key = utils.extract_key(lhs, updated_key_parts)
        local rhs_key = utils.extract_key(rhs, updated_key_parts)
        return keys_comparator(lhs_key, rhs_key)
    end
end
```

I think it looks better and eliminates runtime `has_keydef` flag check inside comparator.

crud/select.lua
```diff
@@ -69,6 +69,7 @@ local function select_on_storage(space_name, index_id, conditions, opts)
         after_tuple = opts.after_tuple,
         tarantool_iter = opts.tarantool_iter,
         limit = opts.limit,
+        conditions = conditions,
```
olegrok, line 72: Do we really need it? I don't see any new usages.

crud/select/executor.lua
```diff
@@ -37,6 +39,7 @@ function executor.execute(space, index, filter_func, opts)
         after_tuple = '?table',
         tarantool_iter = 'number',
         limit = '?number',
+        conditions = '?table'
```
olegrok, line 42: My previous comment is also relevant for this `conditions`.

olegrok (approved):
Hooooray! 🥇 

Thanks for your patch. @dokshina take a look please.

crud/select/executor.lua
```diff
@@ -2,6 +2,8 @@ local errors = require('errors')
 
 local dev_checks = require('crud.common.dev_checks')
 local select_comparators = require('crud.compare.comparators')
+local compat = require('crud.common.compat')
+local has_keydef, keydef_lib = pcall(compat.require, 'tuple.keydef', 'key_def')
```
dokshina, line 6: I don't like the idea of using `pcall(compat.require)`. It can lead to ignoring some errors.
What about simple function `compare.exists('tuple.keydef', 'key_def')` that makes smth like this

```lua
    if module_cached ~= nil then
       return true
    end

    if package.search(module_name) then
        return true
    end

     if package.loaded[builtin_module_name] ~= nil then
         return true
    end

    return false
```

crud/select/executor.lua
```diff
@@ -53,11 +55,18 @@ function executor.execute(space, index, filter_func, opts)
...
         else
-            local cmp_operator = select_comparators.get_cmp_operator(opts.tarantool_iter)
-            local scan_comparator = select_comparators.gen_tuples_comparator(cmp_operator, index.parts)
-            local after_tuple_key = utils.extract_key(opts.after_tuple, index.parts)
-            if scan_comparator(after_tuple_key, opts.scan_value) then
-                value = after_tuple_key
+            if has_keydef then
+                local key_def = keydef_lib.new(index.parts)
+                if key_def:compare_with_key(opts.after_tuple, opts.scan_value) < 0 then
+                    value = key_def:extract_key(opts.after_tuple)
```
dokshina, line 61: Seems that this `if`s with `key_def:extract_key` or `utils.extract_key` can be moved to a separated function that knows if key def can be used or not.

dokshina (approved):
Bless you! Thanks a lot for this patch.

crud/select/executor.lua
olegrok: I left a comment for this topic https://github.com/tarantool/crud/commit/eccaff5123ba729574afbdf4583ee23bc6e5109b#r52462548

dokshina (approved):
Bless you!

PR #160, opened 2021-05-27
fix key_def cache invalidation

olegrok:
When "fields" feature was introduced we change caching strategy
and started to keep "string" key for index + fields.
This approach led to the cases when nobody could invalidate
key_def cache and if user changes index schema and changes e.g.
fields order in his/her index it won't reflect sort order. Old
value stuck in cache.

This patch fixes such behaviour. We don't consider fields in
key_def cache - yes we should recreate key_def each time. But at
least it's safe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
     by calling `package.search`, built-in `merger` and `keydef` modules are used
     if present in `package.loaded`. It allows to avoid ignoring errors on checking
     modules existing via `pcall(require, '<module_name>')`.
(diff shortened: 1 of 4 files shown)


dokshina (approved):
Don't forget about Changelog!

Totktonada:
How big the performance impact?

Can we just add a schema version into the cache key?

olegrok:
> How big the performance impact?

I think it's not so huge. But I think it's better than nothing.

```
local start = clock.time()
for _ = 1, 1e6 do
    key_def_lib.new({{type = 'number', field = 1}})
end
print(clock.time() - start)
```

```
tarantool init.lua
1.3106219768524
```

> Can we just add a schema version into the cache key?

Here we interested in "index" version, not whole schema. But I think in general it's possible.

PR #165, opened 2021-06-02
Imply vshard roles

rosik:
Two new cartridge role properties were introduced in https://github.com/tarantool/cartridge/pull/1217:

```
implies_router = true,
implies_storage = true,
```

They're necessary to preserve Cartridge WebUI operability (bootstrap vshard button visible) even if vshard roles are hidden explicitly.

This patch adds these properties to the crud roles, as they already have corresponding dependencies.


olegrok (approved):

dokshina (approved):
Please, update CHANGELOG before the merge.

PR #169, opened 2021-06-15
WIP: Multipart index single field query

olegrok:
Only tests from this patch have value. This patch can't be merged as is because introduces performance degradation.

### DIFF
--- a/crud/select/executor.lua
+++ b/crud/select/executor.lua
@@ -49,19 +49,6 @@ function executor.execute(space, index, filter_func, opts)
     local tuples_count = 0
 
     local value = opts.scan_value
-    if opts.after_tuple ~= nil then
-        if value == nil then
-            value = opts.after_tuple
-        else
-            local cmp_operator = select_comparators.get_cmp_operator(opts.tarantool_iter)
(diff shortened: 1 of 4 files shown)


crud/select/executor.lua
```diff
@@ -49,19 +49,6 @@ function executor.execute(space, index, filter_func, opts)
...
-    if opts.after_tuple ~= nil then
-        if value == nil then
-            value = opts.after_tuple
-        else
-            local cmp_operator = select_comparators.get_cmp_operator(opts.tarantool_iter)
-            local scan_comparator = select_comparators.gen_tuples_comparator(cmp_operator, index.parts)
-            local after_tuple_key = utils.extract_key(opts.after_tuple, index.parts)
-            if scan_comparator(after_tuple_key, opts.scan_value) then
-                value = after_tuple_key
-            end
```
olegrok, line 61: Actually after this change `scroll_to_after_tuple` makes fullscan. So funny that our tests don't catch it.

olegrok:
Will be fixed in https://github.com/tarantool/crud/pull/177

PR #175, opened 2021-06-21
Fix external keydef compatibility with builtin merger

### LINKED ISSUE #171: External key_def is not compatible with built-in merger
```lua
tarantool> key_def_lib = require('tuple.keydef')
---
...

tarantool> require('merger').new(key_def_lib.new({{field = 1, type = 'string'}}), {})
---
- error: 'Bad params, use: merger.new(key_def, {source, source, ...}[, {reverse =
    <boolean> or <nil>}])'
...
```

```lua
tarantool> key_def_lib = require('key_def')
---
...

tarantool> require('merger').new(key_def_lib.new({{field = 1, type = 'string'}}), {})
---
- 'cdata<struct merge_source &>: 0x7fe8afac94c0'
...

```

dokshina:
Closes #171

### DIFF
--- a/crud/compare/keydef.lua
+++ b/crud/compare/keydef.lua
@@ -7,6 +7,10 @@ local compat = require('crud.common.compat')
 local keydef_lib = compat.require('tuple.keydef', 'key_def')
 local merger_lib = compat.require('tuple.merger', 'merger')
 
+-- When built-in merger is used with external keydef, `merger_lib.new(keydef)`
+-- fails. It's simply fixed by casting `keydef` to 'struct key_def&'.
+-- This cast can be performed only when keydef is external, but merger is built-in.
+-- Otherwise, it leads to different errors (e.g. for Tarantool 1.10).
 local keyef_needs_cast = keydef_lib.__is_external and not merger_lib.__is_external
 
 -- As "tuple.key_def" doesn't support collation_id
--- a/crud/compare/keydef.lua
+++ b/crud/compare/keydef.lua
@@ -1,3 +1,5 @@
+local ffi = require('ffi')
+
 local comparators = require('crud.compare.comparators')
 local collations = require('crud.common.collations')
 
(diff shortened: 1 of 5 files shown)


crud/compare/keydef.lua
```diff
@@ -55,6 +58,10 @@ local function new(space, field_names, index_id)
         keydef = keydef:merge(keydef_lib.new(normalize_parts(updated_parts)))
     end
 
+    if keydef_lib.__is_external and not merger_lib.__is_external then
```
olegrok, line 61: Maybe it's better to save result to `need_cast` variable to check only one flag?

dokshina: I don't think that it makes sense, these modules are required separately, `__is_external ` is set for each of them. This condition is checked only once here.

olegrok: > This condition is checked only once here.

This condition is checked for each key_def.new

Totktonada: Why don't cast always (and add a comment that would describe the situation)? It would simplify logic.

dokshina: See https://github.com/tarantool/crud/actions/runs/957320219. It fails for 1.10 with `'struct key_def' has no member named 'compare'` error.

olegrok: Sometimes it shoots.



On 1.10 when merger is external.

olegrok (approved):
Thanks for your patch. Bless you.

crud/compare/keydef.lua
Totktonada: Ah, it seems, got it. When there is no built-in key_def module, we have no metatype with __index.compare for `cdata<struct key_def &>`. But crud calls this method somewhere.

I guess there should not be a problem, if we'll cast right before passing key_def to merger. It also better, because we'll not loss the link to the original key_def, so it'll not garbage collected.

dokshina: Thanks for your idea!

PR #177, opened 2021-06-29
Fix pagination + multipart indexes

### LINKED ISSUE #170: [8pt] Pagination [+ multipart indexes] is broken
Following tests should be fixed https://github.com/tarantool/crud/pull/169/files#diff-6ea5935828108f3cd61997766f7c37952a79debc13a01a31a4c168ed343c843f

mRrvz:
Fixed invalid results for pagination queries when filtering on
a parts of a composite index. Now, instead of "scrolling" 
to the desired tuple (during pagination) using the 
``scroll_to_after_tuple`` function, we change the iterators 
(``EQ`` -> ``GE``, ``REQ`` -> ``LT``) to use them in 
``index:pairs`` function, together with a combination with 
additional filtration. Closes #170


crud/select/executor.lua
```diff
@@ -83,6 +83,16 @@ function executor.execute(space, index, filter_func, opts)
...
 
+    if opts.after_tuple ~= nil then
+        if opts.tarantool_iter == box.index.LE then
+            opts.tarantool_iter = box.index.LT
+        elseif opts.tarantool_iter == box.index.EQ then
+            opts.tarantool_iter = box.index.GE
+        elseif opts.tarantool_iter == box.index.REQ then
+            opts.tarantool_iter = box.index.LT
+        end
+    end
```
olegrok, line 94: I think it could be moved to the plan

crud/select/filters.lua
```diff
@@ -649,11 +666,13 @@ function filters.gen_func(space, conditions, opts)
     dev_checks('table', '?table', {
         tarantool_iter = 'number',
         scan_condition_num = '?number',
+        after_tuple = '?table',
```
olegrok, line 669: I don't see any usages. Is it really needed?

crud/select/filters.lua
```diff
@@ -649,11 +666,13 @@ function filters.gen_func(space, conditions, opts)
...
     })
 
     local filter_conditions, err = parse(space, conditions, {
         scan_condition_num = opts.scan_condition_num,
         tarantool_iter = opts.tarantool_iter,
+        after_tuple = opts.after_tuple,
```
olegrok, line 675: I don't see any usages. Is it really needed?

crud/select/filters.lua
olegrok: My bad. It's used.

crud/select/filters.lua
```diff
@@ -95,6 +96,22 @@ local function parse(space, conditions, opts)
     end
 
     local filter_conditions = {}
+
+    if opts.after_tuple ~= nil and #conditions > 0 then
+        if opts.tarantool_iter == box.index.EQ or opts.tarantool_iter == box.index.REQ then
+            local index = space_indexes[conditions[1].operand]
```
olegrok, line 102: I think it should be `opts.scan_condition_num` not just `1`.

ligurio (changes requested):
@mRrvz Alexey, thanks for the patch!

Few comments:

- could you add explanations to commit messages "Fix unnecessary sharding key", "Remove unnecessary test". For me it is not clear what was fixed regarding to sharding key and why we need to remove test.
- I suppose the item to changelog should be added in the same commit where fix was added.

crud/select/plan.lua
```diff
@@ -208,7 +208,7 @@ function select_plan.new(space, conditions, opts)
 
     -- get sharding key value
     local sharding_key
-    if scan_value ~= nil and (scan_iter == box.index.EQ or scan_iter == box.index.REQ) then
+    if scan_value ~= nil and scan_after_tuple == nil and (scan_iter == box.index.EQ or scan_iter == box.index.REQ) then
```
olegrok, line 211: Why?

It seems only first query could be executed on single shard if full sharding key is specified.

crud/select/filters.lua
olegrok: Or plan contains index id. It could be also used.

crud/select/executor.lua
dokshina: I was sure that we already have such code in the plan...

olegrok: > I was sure that we already have such code in the plan...

We have only pain

crud/select/plan.lua
```diff
@@ -206,6 +212,16 @@ function select_plan.new(space, conditions, opts)
...
 
+    if scan_after_tuple ~= nil then
+        if scan_iter == box.index.LE then
+            scan_iter = box.index.LT
+        elseif scan_iter == box.index.EQ then
+            scan_iter = box.index.GE
+        elseif scan_iter == box.index.REQ then
+            scan_iter = box.index.LT
+        end
+    end
```
olegrok, line 223: nit: I think this change should be done before `local sharding_index = primary_index` line. 

`local sharding_index = primary_index` and `local sharding_key ...` are in "sharding" group of code.

crud/select/plan.lua
```diff
@@ -182,6 +182,12 @@ function select_plan.new(space, conditions, opts)
         return nil, err
     end
 
+    if scan_after_tuple ~= nil then
+        if scan_iter == box.index.REQ or scan_iter == box.index.EQ then
+            table.insert(conditions, conditions[scan_condition_num])
+        end
+    end
```
olegrok, line 189: Patch is more laconic and it's great but this change is not trivial. Please put a comment here.

test/unit/select_plan_test.lua
```diff
@@ -256,7 +256,7 @@ g.test_first = function()
     t.assert_equals(plan.scan_value, {777}) -- after_tuple id
     t.assert_equals(plan.after_tuple, after_tuple)
     t.assert_equals(plan.scan_condition_num, nil)
-    t.assert_equals(plan.tarantool_iter, box.index.LE) -- inverted iterator
+    t.assert_equals(plan.tarantool_iter, box.index.LT) -- inverted iterator
```
olegrok, line 259: Maybe it worth to add more unit test that shows that EQ iterator is changed to GT if after tuple is passed?

crud/select/plan.lua
```diff
@@ -182,6 +182,18 @@ function select_plan.new(space, conditions, opts)
         return nil, err
     end
 
+    -- We duplicate this condition so that it works in `filters.gen_func`
+    -- (see function `parse`) - therefore it does not process `conditions`
+    -- (and not converted them to `filtter_conditions` array) with an array
```
olegrok, line 187: `filtter_conditions` -> `filter_conditions `

crud/select/plan.lua
```diff
@@ -182,6 +182,18 @@ function select_plan.new(space, conditions, opts)
...
+    -- index equal to opts.scan_condition_num. This condition, together with
+    -- the cobination "inversion" of the scan iterator (see lines below),
+    -- is needed for correct pagination when using a partial index.
```
olegrok, line 190: To be honest I don't understand it.

I think that:
```
Since we use pagination we should continue iteration since after tuple.
Such iterator could be run only with index:pairs(key(after_tuple), 'GT/LT').
To preserve original condition we should manually inject it in filter_conditions
(see ...).
```

I think also we can't use "inversion" term for iterator. Inversion is transformation LT -> GT, GE -> LE, but here it is changed just to continue iterator in direct and reverse order.

olegrok (approved):
Thanks for your patch. Please resolve the rest of not my comments.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+* Pagination with composite indexes
```
dokshina, line 12: Please describe a kind of error. It wasn't working at all? Or bad result was returned?

dokshina (approved):
Please make Changelog more informative.

dokshina:
> @mRrvz Alexey, thanks for the patch!
> 
> Few comments:
> 
> * could you add explanations to commit messages "Fix unnecessary sharding key", "Remove unnecessary test". For me it is not clear what was fixed regarding to sharding key and why we need to remove test.
> * I suppose the item to changelog should be added in the same commit where fix was added.

.

ligurio (changes requested):

ligurio (approved):

dokshina (approved):

PR #180, opened 2021-07-05
Refactoring code in select modules

### LINKED ISSUE #144: [1pt] Seems that error can be ignored in some cases of `crud.select`
There is tuple iteration
And here we first check tuple and than error.

https://github.com/tarantool/crud/blob/master/crud/select/compat/select_old.lua#L211

In other places it's done better way:
```lua
        local tuple, err = iter:get()
        if err ~= nil then
            return nil, SelectError:new("Failed to get next object: %s", err)
        end

        if tuple == nil then
            break
        end
```

mRrvz:
Checking error in the select_old module when
receiving iterator is now called earlier than
checking tuple. Closes #144

### DIFF
--- a/crud/select/compat/select_old.lua
+++ b/crud/select/compat/select_old.lua
@@ -218,12 +218,12 @@ function select_module.pairs(space_name, user_conditions, opts)
 
     local gen = function(_, iter)
         local tuple, err = iter:get()
-        if tuple == nil then
-            return nil
-        end
-
         if err ~= nil then
             error(string.format("Failed to get next object: %s", err))
+        end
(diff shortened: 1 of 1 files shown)


olegrok (approved):
Thanks for your patch. Changes looks ok for me. Is it possible to test it somehow?

Also current commit message `Refactoring code in select modules` doesn't say anything about a problem you solve and it's worth to add `Closes #144` directly to commit message.

ligurio (approved):
Thanks for the patch!

dokshina (approved):

PR #182, opened 2021-07-11
Len implementation

### LINKED ISSUE #164: [2pt] Implement `crud.len()` to count space size in cluster
Similar to all other cluster functionality, `crud.len()` will iterate over each shard and count space size

AnaNek:
Now there is `crud.len()` function to iterate over each shard and count space size
Closes #164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-* `crud.len()` function to iterate over each shard and count space size
+* `crud.len()` function to calculate the number of tuples
+  in the space for memtx engine and calculate the maximum
+  approximate number of tuples in the space for vinyl engine.
 
 ## [0.8.0] - 02-07-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+* `crud.len()` function to iterate over each shard and count space size
+
 ## [0.8.0] - 02-07-21
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -497,7 +497,8 @@ local result, err = crud.len(space_name, opts)
 
 where:
 
-* `space_name` (`string`) - name of the space
+* `space_name` (`string|number`) - name of the space as well
+  as numerical id of the space
 * `opts`:
   * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
 
--- a/README.md
+++ b/README.md
@@ -487,6 +487,34 @@ crud.truncate('customers', {timeout = 2})
 ...
 ```
 
+### Len
+
+```lua
+-- Calculates the number of tuples in the space
+local result, err = crud.len(space_name, opts)
+```
+
+where:
+
+* `space_name` (`string`) - name of the space
+* `opts`:
+  * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
+
+Returns number or nil with error.
+
+**Example:**
+
+```lua
+#crud.select('customers')
+---
+- 5
+...
+crud.len('customers', {timeout = 2})
+---
+- 5
+...
+```
+
 ## Cartridge roles
 
 `cartridge.roles.crud-storage` is a Tarantool Cartridge role that depends on the
--- a/README.md
+++ b/README.md
@@ -490,7 +490,8 @@ crud.truncate('customers', {timeout = 2})
 ### Len
 
 ```lua
--- Calculates the number of tuples in the space
+-- Calculates the number of tuples in the space for memtx engine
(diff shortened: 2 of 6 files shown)


test/integration/len_test.lua
```diff
@@ -0,0 +1,83 @@
...
+    local result, err = g.cluster.main_server.net_box:call('crud.len', {'non_existent_space'})
+
+    t.assert_equals(result, nil)
+    t.assert_str_contains(err.err, "Space \"non_existent_space\" doesn't exist")
+end)
+
+pgroup:add('test_len', function(g)
+    local bucket_id = 1
+    local other_bucket_id, err = helpers.get_other_storage_bucket_id(g.cluster, bucket_id)
+    t.assert(other_bucket_id ~= nil, err)
```
olegrok, line 43: I think it's redundant overcomplication. Let's imagine that sequential ids uniformally distributed across storages.

I think simple loop gives you really high probability that you have at least one tuple on each storage.
```
local customers = {}
for i = 1, 100 do 
    table.insert(customers, {
    id = i, name = tostring(i), last_name = tostring(i), ...
})
end
helpers.insert_objects(g, 'customers', customers)
```

README.md
```diff
@@ -487,6 +487,34 @@ crud.truncate('customers', {timeout = 2})
...
+Returns number or nil with error.
+
+**Example:**
+
+```lua
+#crud.select('customers')
+---
+- 5
+...
+crud.len('customers', {timeout = 2})
```
olegrok, line 512: It's relevant only if space engine is memtx

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+* `crud.len()` function to iterate over each shard and count space size
```
olegrok, line 12: `count space size`. The same. Len for vinyl has a bit different name - " maximum approximate number of tuples in the space".

https://www.tarantool.io/en/doc/latest/book/box/engines/

olegrok (commented):
Thanks for your patch. See my comments below.

crud/len.lua
```diff
@@ -0,0 +1,93 @@
...
+    opts = opts or {}
+
+    local space = utils.get_space(space_name, vshard.router.routeall())
+    if space == nil then
+        return nil, LenError:new("Space %q doesn't exist", space_name), true
+    end
+
+    local replicasets = vshard.router.routeall()
+    local call_opts = {
+        mode = 'read',
```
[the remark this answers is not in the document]
no1seman: + seems that len must be calculated only on masters

ligurio (changes requested):

crud/len.lua
```diff
@@ -0,0 +1,105 @@
+local checks = require('checks')
+local errors = require('errors')
+local vshard = require('vshard')
+
+local call = require('crud.common.call')
+local utils = require('crud.common.utils')
+local dev_checks = require('crud.common.dev_checks')
+local schema = require('crud.common.schema')
+
+local LenError = errors.new_class('Len',  {capture_stack = false})
```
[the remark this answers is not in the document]
ligurio: (Other name consistency problems fixed in https://github.com/tarantool/crud/pull/184)

crud/len.lua
```diff
@@ -0,0 +1,72 @@
...
+    })
+
+    opts = opts or {}
+
+    local replicasets = vshard.router.routeall()
+    local results, err = call.map(LEN_FUNC_NAME, {space_name}, {
+        mode = 'write',
+        replicasets = replicasets,
+        timeout = opts.timeout,
+    })
```
olegrok, line 56: You can use `vshard.rouer.map_callrw` (https://github.com/tarantool/vshard/commit/a394e3f4008a7436f011c81028dcaadc270eebf6) for such purpose. It gives more guarantees in case when rebalanser is active.

test/integration/len_test.lua
AnaNek: But why is the use of a loop more preferable here than in the other tests?(https://github.com/tarantool/crud/blob/master/test/integration/borders_test.lua#L282)

olegrok: I never like such approach to tests - test should be clear and simple as a stick. An attempt to add some "read world sense" is redundant. Moreover currently you need only call "len" and have some guarantees that data is distributed across storages. This loop is shorter than current variant and gives such guarantees.

olegrok (approved):

README.md
```diff
@@ -487,6 +487,34 @@ crud.truncate('customers', {timeout = 2})
...
+### Len
+
+```lua
+-- Calculates the number of tuples in the space
+local result, err = crud.len(space_name, opts)
+```
+
+where:
+
+* `space_name` (`string`) - name of the space
```
ligurio, line 500: `space_name` can be a numerical id too. Please fix the description for argument and types here and in code.

ligurio (changes requested):

ligurio (approved):

PR #183, opened 2021-07-12
Move luacheck to a separate pipeline

ligurio:
**PLEASE MERGE IT WITH REBASE** (without squash)

> What has been done?

`luacheck` moved to a separate pipeline in Github Actions

> Why? What problem is being solved?

Imagine we have an unused variable in pushed source code. All tasks in Github Actions will be failed because `luacheck` run in all of them and blocks regression testing in case of fail.

I didn't forget about

- ~~[ ] Tests~~
- ~~[ ] Changelog~~
- ~~[ ] Documentation~~

Closes #???

### DIFF
--- a/.github/workflows/check_on_push.yaml
+++ b/.github/workflows/check_on_push.yaml
@@ -1,0 +1,25 @@
+name: Run static analysis
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  run-static-analysis:
+    if: |
+      github.event_name == 'push' ||
+      github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Setup Tarantool
+      uses: tarantool/setup-tarantool@v1
+      with:
+        tarantool-version: '2.6'
+
+    - name: Setup luacheck
+      run: tarantoolctl rocks install luacheck 0.25.0
+
+    - name: Run luacheck
+      run: .rocks/bin/luacheck .
--- a/.github/workflows/test_on_push.yaml
+++ b/.github/workflows/test_on_push.yaml
@@ -8,7 +8,7 @@ jobs:
   run-tests-ce:
     if: |
       github.event_name == 'push' ||
-      github.event_name == 'pull_request' && github.event.pull_request.head.repo.owner.login != 'tarantool'
+      github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository
     strategy:
       matrix:
         tarantool-version: ["1.10", "2.2", "2.3", "2.4", "2.5", "2.6", "2.7"]
--- a/.github/workflows/test_on_push.yaml
+++ b/.github/workflows/test_on_push.yaml
@@ -21,10 +21,13 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
+      - name: Setup Tarantool CE
+        uses: tarantool/setup-tarantool@v1
+        with:
+          tarantool-version: ${{ matrix.tarantool-version }}
+
       - name: Install requirements for community
         run: |
-          curl -L https://tarantool.io/installer.sh | sudo VER=${{ matrix.tarantool-version }} bash
-          sudo apt install -y tarantool-dev
           tarantool --version
           ./deps.sh
 
--- a/.github/workflows/test_on_push.yaml
+++ b/.github/workflows/test_on_push.yaml
@@ -39,9 +39,6 @@ jobs:
       - name: Stop Mono server
         run: sudo kill -9 $(sudo lsof -t -i tcp:8084) || true
 
-      - name: Run linter
-        run: .rocks/bin/luacheck .
-
       - name: Run tests
         run: .rocks/bin/luatest -v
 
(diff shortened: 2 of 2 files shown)


deps.sh
```diff
@@ -5,7 +5,6 @@ set -e
 
 # Test dependencies:
 tarantoolctl rocks install luatest 0.5.3
-tarantoolctl rocks install luacheck 0.25.0
```
mRrvz, line 8: Do we really need this change?

ligurio: `deps.sh` used in YAML files for GH Actions for dependencies installation. With my proposed patch luacheck is unused in tasks where crud tested. In a new pipeline luacheck installed directly using `tarantoolctl`, using `deps.sh` I believe will take more time due to `cartridge` and `luatest`.

ligurio: So the answer is "Yes".

mRrvz (approved):

AnaNek (approved):

.github/workflows/check_on_push.yaml
```diff
@@ -0,0 +1,24 @@
...
+      github.event_name == 'push' ||
+      github.event_name == 'pull_request' && github.event.pull_request.head.repo.owner.login != 'tarantool'
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@master
+
+      - name: Setup luacheck
+        run: |
+          curl -L https://tarantool.io/installer.sh | sudo bash
+          sudo apt install -y tarantool-dev
```
Totktonada, line 19: Just tarantool (without headers) should be okay. However I suggest to use [setup-tarantool](https://github.com/tarantool/setup-tarantool) for such simple workflows. I employs github's caching infrastructure to make pipelines incredibly fast.

.github/workflows/check_on_push.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  run-static-analysis:
+    if: |
+      github.event_name == 'push' ||
+      github.event_name == 'pull_request' && github.event.pull_request.head.repo.owner.login != 'tarantool'
```
Totktonada, line 11: It seems, we can do it without hardcoding 'tarantool' organization name. Please, consider this [code](https://github.com/tarantool/smtp/blob/556393ac5a8906a47910331e70db17481f60326a/.github/workflows/testing.yml#L12-L17) and the link inside it.

ligurio: This condition can be useful when we decide to use our own workers.
Anyway removed it now.

.github/workflows/check_on_push.yaml
ligurio: Yep, totally forget about this action. Now luacheck pipeline uses it.

.github/workflows/test_on_push.yaml
```diff
@@ -22,8 +22,10 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Install requirements for community
+        uses: tarantool/setup-tarantool@v1
+        with:
+          tarantool-version: ${{ matrix.tarantool-version }}
         run: |
-          curl -L https://tarantool.io/installer.sh | sudo VER=${{ matrix.tarantool-version }} bash
           sudo apt install -y tarantool-dev
```
Totktonada, line 29: tarantool-dev is installed by setup-tarantool.

.github/workflows/check_on_push.yaml
Totktonada: I didn't vote for removal, but for changing of `github.event.pull_request.head.repo.owner.login != 'tarantool'` to `github.event.pull_request.head.repo.full_name != github.repository`.

.github/workflows/check_on_push.yaml
Totktonada: Nice!

.github/workflows/check_on_push.yaml
Totktonada: The positive aspect is that it should work as expected for a pull request to a non-mainline repository (from one fork to another fork).

ligurio: This repository has only 3 forks and all of them are made by Tarantool's people. Why do you care about forks?

.github/workflows/test_on_push.yaml
ligurio: Removed `tarantool-dev` installation.

.github/workflows/check_on_push.yaml
Totktonada: That's the short recipe, which should work for all cases. Such recipes often walk from a repository to repository, so it would be nice to use a good default everywhere (if there are no specific requirements).

Totktonada: Seems resolved.

.github/workflows/test_on_push.yaml
```diff
@@ -22,9 +22,10 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Install requirements for community
+        uses: tarantool/setup-tarantool@v1
+        with:
+          tarantool-version: ${{ matrix.tarantool-version }}
```
Totktonada, line 27: >  Error : .github#L1
> a step cannot have both the `uses` and `run` keys

https://github.com/tarantool/crud/actions/runs/1029121910

deps.sh
olegrok: This change is not a part of GH actions primary. This file should make life simple for developers.

For local development you need to call ./deps.sh - and you will get all needed toolchain for local testing. Please revert this change.

ligurio: reverted

.github/workflows/test_on_push.yaml
ligurio: Fixed, thanks.

olegrok (approved):

PR #184, opened 2021-07-12
Number of fixes (typos, error names unification)

ligurio:
**PLEASE MERGE IT WITH REBASE** (without squash)

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Closes #???


crud/borders.lua
```diff
@@ -186,7 +186,7 @@ end
 -- @param string space_name
 --  A space name
 --
--- @param ?string index_name
+-- @param ?string index_id
 --  An index name (by default, primary index is used)
```
mRrvz, line 190: The parameter is called ``index_id``, but below it is described as ``An index name (by default, primary index is used)``. This is confusing

ligurio: Agree. Fixed and pushed.

mRrvz (approved):

AnaNek (approved):

rosik (approved):

crud/borders.lua
```diff
@@ -163,8 +163,8 @@ end
 -- @param string space_name
 --  A space name
 --
--- @param ?string index_name
---  An index name (by default, primary index is used)
+-- @param ?string index_id
+--  An index identifier (by default, primary index is used)
```
Totktonada, line 167: Names are inconsistent: space_name, index_id. AFAIS, both accept as space/index string name as well as its numeric ID.

crud/insert.lua
```diff
@@ -8,7 +8,7 @@ local sharding = require('crud.common.sharding')
 local dev_checks = require('crud.common.dev_checks')
 local schema = require('crud.common.schema')
 
-local InsertError = errors.new_class('Insert',  {capture_stack = false})
+local InsertError = errors.new_class('InsertError',  {capture_stack = false})
```
Totktonada, line 11: I'm okay generally, but are there clients, which lean on the old naming? @akudiyar, how about Java connector?

akudiyar: We don't stick to any error names at the moment yet apart from the netbox' `ClientError`. But @ArtDu is doing some research now about the tarantool/crud errors in the scope of  https://github.com/tarantool/cartridge-java/pull/102.

@ArtDu please, take a look at changes in this PR.

ArtDu: These changes have no effect on the code I was writing in tarantool/cartridge-java#102. Research with errors is still ongoing. So the changes are ok

ligurio: Thanks, Alexander, Alexey and Artem!

crud/borders.lua
ligurio: `borders.min()` do nothing with arguments and pass them to `get_border`, `get_border` too and calls `call_get_border_on_router` that uses `space_name` as a name of space with type string in its body. The same with `index_name`. I suggest using `{index, space}_name` because using the name is more friendly for humans instead of indices. Additionally document functions arguments and mention that space and index ID's accepted too.

Totktonada: Ouch, I missed the `checks()` call. Maybe it worth to allow numerical IDs here?

Generally speaking `box.space[<...>]` accepts both numerical IDs and string names, so upward APIs usualy reflects this property. I would use `*_name` names and comment that a numerical ID is okay here too.

ligurio: Added `numeric` type too:
```lua
--- a/crud/borders.lua
+++ b/crud/borders.lua
@@ -17,7 +17,7 @@ local STAT_FUNC_NAME = '_crud.get_border_on_storage'
 
 
 local function get_border_on_storage(border_name, space_name, index_name, field_names)
-    dev_checks('string', 'string', 'string|number', '?table')
+    dev_checks('string', 'string|number', 'string|number', '?table')
 
     assert(border_name == 'min' or border_name == 'max')
 
@@ -67,7 +67,7 @@ else
 end
 
 local function call_get_border_on_router(border_name, space_name, index_name, opts)
-    checks('string', 'string', '?string|number', {
+    checks('string', 'string|number', '?string|number', {
         timeout = '?number',
         fields = '?table',
     })
```

ligurio: And I made types of `space_name` consistent in a separate commit.

README.md
```diff
@@ -66,7 +66,7 @@ local result, err = crud.insert_object(space_name, object, opts)
 
 where:
 
-* `space_name` (`string`) - name of the space to insert an object
+* `space_name` (`string|number`) - name or numerical id of the space to insert an object
```
olegrok, line 69: Is it tested anyhow?

Won't it be a problem because on different replicasets the same space could have different id?

crud/borders.lua
olegrok: When we consider index_id as a number we probably have more guarantees that spaces on different replicasets have the same index_id. And it's not so in case of spaces. In case of cartridge we could run some role with storage role on some replicaset we have different space_id with "plain" storage because "additional" role on the first storage could initialize its own spaces before storage spaces. Also there is local and temporary spaces that could be replicaset-specific and have own space_id.

I think space_id is not so much useful in distributed systems.

ligurio: Well, I agree with your concerns about index_id and space_id, so I reverted changes regarding index_id and space_id.

README.md
ligurio: No actually, I have reverted these changes.

crud/borders.lua
Totktonada: Oleg's points convinced me too. Thanks!

mRrvz:
Please, update the CHANGELOG.

olegrok (approved):

ligurio:
@mRrvz Updated.

mRrvz (approved):

PR #189, opened 2021-07-31
Doc: described examples schema and info about select queries priority

### LINKED ISSUE #187: Add full space format to "Select examples"
It's not obviously that "full_name" is an index and seems that full space format including all indexes must be added to: https://github.com/tarantool/crud/blob/master/doc/select.md

### LINKED ISSUE #188: Add note of priority of index against field if names are equal in crud.select conditions

mRrvz:
Closes #187 and Closes #188

### DIFF
--- a/doc/pairs.md
+++ b/doc/pairs.md
@@ -3,6 +3,7 @@
 With ``crud.pairs``, you can iterate across a distributed space.  
 The arguments are the same as [``crud.select``](https://github.com/tarantool/crud/blob/master/doc/select.md), except of the ``use_tomap`` parameter.  
 Below are examples that may help you.
+Examples schema is similar to the [select documentation](select.md/#examples-space-format)
 
 ## Getting space
 
--- a/doc/select.md
+++ b/doc/select.md
@@ -6,7 +6,52 @@
 
 **Note:** If you specify sharding key or ``bucket_id`` select will be performed on single node. Otherwise Map-Reduce over all nodes will be occurred.
 
-Below are examples of filtering data using these conditions. 
+Below are examples of filtering data using these conditions.
+
+### Examples schema
+
+```lua
+box.space.developers:format()
+---
+- {'name': 'id', 'type': 'unsigned'}
+- {'name': 'bucket_id', 'type': 'unsigned'}
+- {'name': 'name', 'type': 'string'}
+- {'name': 'surname', 'type': 'string'}
+- {'name': 'age', 'type': 'number'}
+...
+box.space.developers.index
+- 0: &0
+    unique: true
+    parts:
+    - type: unsigned
+      is_nullable: false
+      fieldno: 1
+    id: 0
+    type: TREE
+    name: primary_index
+  1: &1
+    unique: false
+    parts:
+    - type: number
+      is_nullable: false
+      fieldno: 5
+    id: 1
+    type: TREE
+    name: age_index
+  2: &2
+    unique: false
+    parts:
+    - type: string
+      is_nullable: false
+      fieldno: 3
+    - type: string
+      is_nullable: false
+      fieldno: 4
+    id: 2
+    type: TREE
+    name: full_name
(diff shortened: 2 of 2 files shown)


ligurio (changes requested):
Thanks for the patch! See my comments inline.

ligurio:
@NickVolynkin could you (or your teammate) review changes too?

doc/pairs.md
```diff
@@ -1,8 +1,9 @@
 # Pairs examples
 
-With ``crud.pairs``, you can iterate across a distributed space.  
-The arguments are the same as [``crud.select``](https://github.com/tarantool/crud/blob/master/doc/select.md), except of the ``use_tomap`` parameter.  
-Below are examples that may help you.
+With ``crud.pairs``, you can iterate across a distributed space.
+The arguments are the same as [``crud.select``](https://github.com/tarantool/crud/blob/master/doc/select.md), except of the ``use_tomap`` parameter. Below are examples that may help you.
+
+Examples space format is similar to the [select documentation](https://github.com/tarantool/crud/blob/master/doc/select.md/#examples-space-format)
```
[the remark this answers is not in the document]
mRrvz: Fixed

doc/pairs.md
```diff
@@ -43,7 +44,7 @@ tuples
 
 ## ``use_tomap`` parameter
 
-With ``use_tomap`` flag, you can choose to iterate over objects or over tuples.  
+With ``use_tomap`` flag, you can choose to iterate over objects or over tuples.
```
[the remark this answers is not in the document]
mRrvz: Removed

doc/select.md
```diff
@@ -81,7 +81,7 @@ crud.select('developers', nil, { first = 6 })
 
 ### Select using index
 
-Let's say we have a ``age`` index. Example below gets a list of ``customers`` over 30 years old.
+We have a ``age_index`` index. Example below gets a list of ``customers`` over 30 years old.
```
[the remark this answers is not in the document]
mRrvz: Fixed

ligurio (approved):
LGTM

Totktonada:
#188 is about the case, when a field name is the same as an index name. In this case, it is unclear, whether the name will be interpreted as the index name or the field name (which may have another suitable index, it does not matter). I looked over the diff and I don't see where this situation is described.

mRrvz (commented):
> **Note**: index is named ``age_index``, but we can still query by the field name ``age``
and the search will be done using index without a full scan.

I think from this it is clear that in the case of a match of names, the search will be performed by the index. Anyway, added clarification

@Totktonada

doc/select.md
```diff
@@ -54,6 +99,10 @@ crud.select('developers', {{'>=', 'age', 30}})
 ...
 ```
 
+**Note**: index is named ``age_index``, but we can still query by the field name ``age``
+and the search will be done using index without a full scan. If the names of index and
+field match, the search will also be performed using index.
```
Totktonada, line 104: It would be more intuitive if we would have an example of using an index name in a condition. Now it is not obvious that it is possible to scan over an index not only using a field name.

mRrvz: Fixed

Totktonada (approved):
I would tweak wording here and there (including the existing paragraphs), but the change looks okay in general.

AnaNek (approved):

doc/select.md
```diff
@@ -6,7 +6,52 @@
 
 **Note:** If you specify sharding key or ``bucket_id`` select will be performed on single node. Otherwise Map-Reduce over all nodes will be occurred.
 
-Below are examples of filtering data using these conditions. 
+Below are examples of filtering data using these conditions.
+
+### Examples space format
```
olegrok, line 11: It's better to call it "schema"

mRrvz: Fixed

PR #196, opened 2021-08-12
test: implement simple perf tests

olegrok:
This patch introduces simple performance tests for select and
insert. Tests are really simple and could catch only huge
regressions. The purpose to merge it into master to have simple
way to check performance and such tests shouldn't be broken if
they work in CI.
It worth to note that CI is public and tests run in contains that
mean that result could vary from run to run.

### DIFF
--- a/test/entrypoint/srv_select.lua
+++ b/test/entrypoint/srv_select.lua
@@ -192,6 +192,9 @@ package.preload['customers-storage'] = function()
     }
 end
 
+local box_opts = {
+    readahead = 10 * 1024 * 1024,
+}
 local ok, err = errors.pcall('CartridgeCfgError', cartridge.cfg, {
     advertise_uri = 'localhost:3301',
     http_port = 8081,
@@ -200,8 +203,9 @@ local ok, err = errors.pcall('CartridgeCfgError', cartridge.cfg, {
         'cartridge.roles.crud-router',
         'cartridge.roles.crud-storage',
         'customers-storage',
-    },
-})
+    }},
+    box_opts
+)
 
 if not ok then
     log.error('%s', err)
--- a/test/performance/select_perf_test.lua
+++ b/test/performance/select_perf_test.lua
@@ -1,0 +1,177 @@
+local fio = require('fio')
+local fiber = require('fiber')
+local errors = require('errors')
+local net_box = require('net.box')
+local log = require('log')
+
+local t = require('luatest')
+local g = t.group('perf')
+
+local helpers = require('test.helper')
+
+g.before_all = function()
+    g.cluster = helpers.Cluster:new({
+        datadir = fio.tempdir(),
+        server_command = helpers.entrypoint('srv_select'),
+        use_vshard = true,
+        replicasets = {
+            {
+                uuid = helpers.uuid('a'),
+                alias = 'router',
+                roles = { 'crud-router' },
+                servers = {
+                    { instance_uuid = helpers.uuid('a', 1), alias = 'router' },
+                },
+            },
+            {
+                uuid = helpers.uuid('b'),
+                alias = 's-1',
+                roles = { 'customers-storage', 'crud-storage' },
+                servers = {
+                    { instance_uuid = helpers.uuid('b', 1), alias = 's1-master' },
+                    { instance_uuid = helpers.uuid('b', 2), alias = 's1-replica' },
+                },
+            },
+            {
+                uuid = helpers.uuid('c'),
+                alias = 's-2',
+                roles = { 'customers-storage', 'crud-storage' },
+                servers = {
+                    { instance_uuid = helpers.uuid('c', 1), alias = 's2-master' },
+                    { instance_uuid = helpers.uuid('c', 2), alias = 's2-replica' },
+                },
+            },
+            {
+                uuid = helpers.uuid('d'),
+                alias = 's-2',
+                roles = { 'customers-storage', 'crud-storage' },
+                servers = {
+                    { instance_uuid = helpers.uuid('d', 1), alias = 's3-master' },
+                    { instance_uuid = helpers.uuid('d', 2), alias = 's3-replica' },
+                },
+            }
+        },
+    })
+    g.cluster:start()
+end
+
+g.after_all = function()
+    g.cluster:stop()
+    fio.rmtree(g.cluster.datadir)
+end
+
+g.before_each(function() end)
+
+local function insert_customers(conn, id, count, timeout, report)
+    local customer = {id, box.NULL, 'David', 'Smith', 33, 'Los Angeles'}
+    local start = fiber.clock()
+
+    while (fiber.clock() - start) < timeout do
+        local ok, res, err = pcall(conn.call, conn, [[package.loaded.crud.insert]], {'customers', customer})
+        if not ok then
+            log.error('Insert error: %s', res)
+            table.insert(report.errors, res)
+        elseif err ~= nil then
+            errors.wrap(err)
+            log.error('Insert error: %s', err)
+            table.insert(report.errors, err)
+        else
+            report.count = report.count + 1
+        end
+        customer[1] = customer[1] + count
+    end
+end
+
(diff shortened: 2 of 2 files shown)


test/performance/select_perf_test.lua
```diff
@@ -0,0 +1,176 @@
...
+local fiber = require('fiber')
+local net_box = require('net.box')
+local log = require('log')
+
+local t = require('luatest')
+local g = t.group('perf')
+
+local helpers = require('test.helper')
+
+math.randomseed(os.time())
```
ligurio, line 11: I don't understand why we need seed initialization here if we don't use random generator in a test.

test/performance/select_perf_test.lua
```diff
@@ -0,0 +1,176 @@
...
+        local f = fiber.new(insert_customers, conn, id, fiber_count, timeout, report)
+        f:set_joinable(true)
+        table.insert(fibers, f)
+    end
+
+    for i = 1, fiber_count do
+        fibers[i]:join()
+    end
+
+    log.error('INSERT')
```
ligurio, line 132: Do we need `error` log level here? Why log.info is not enough?

test/performance/select_perf_test.lua
```diff
@@ -0,0 +1,176 @@
...
+        fibers[i]:join()
+    end
+
+    log.error('SELECT')
+    log.error('Fibers count - %d', fiber_count)
+    log.error('Connection count - %d', connection_count)
+    log.error('Timeout  - %f', timeout)
+    log.error('Requests - %d', report.count)
+    log.error('Errors - %s', #report.errors)
+    log.error('RPS - %f', report.count / timeout)
```
ligurio, line 175: The same. Seems 'info' level is enough here.

ligurio (changes requested):
Thanks for the patch!

typo in commit message: "containes" -> "containers"

test/performance/select_perf_test.lua
olegrok: As the highest log level it will be printed in console anyway.
"Info" could be lower than current log_level in some cases.

Currently it's not about errors and something similar. It's an approach when we don't lost this messages.

test/performance/select_perf_test.lua
```diff
@@ -0,0 +1,176 @@
...
+g.before_each(function() end)
+
+local function insert_customers(conn, id, count, timeout, report)
+    local customer = {id, box.NULL, 'David', 'Smith', 33, 'Los Angeles'}
+    local start = fiber.clock()
+
+    while (fiber.clock() - start) < timeout do
+        local ok, res, err = pcall(conn.call, conn, [[package.loaded.crud.insert]], {'customers', customer})
+        if not ok then
+            log.error('Insert error: %s', res)
```
ligurio, line 73: When error happen test prints messages like below:
```
Insert error: table: 0x1aa18998                                                                                                                               
Insert error: table: 0x1aa18c58                                                                                                                               
Insert error: table: 0x1aa18670                                                                                                                               
Insert error: table: 0x1aa18030                                                                                                                               
Insert error: table: 0x1aa182f0                                                                                                                               
Insert error: table: 0x1aa17d60
```
I suppose table address is not useful. It would be better to print error message to understand a reason of error.

ligurio (changes requested):

test/performance/select_perf_test.lua
```diff
@@ -0,0 +1,176 @@
...
+    end
+
+    log.error('SELECT')
+    log.error('Fibers count - %d', fiber_count)
+    log.error('Connection count - %d', connection_count)
+    log.error('Timeout  - %f', timeout)
+    log.error('Requests - %d', report.count)
+    log.error('Errors - %s', #report.errors)
+    log.error('RPS - %f', report.count / timeout)
+end
```
ligurio, line 176: What is criteria for passed test?
Should we fail test if `#report.errors ~= 0`?

ligurio (changes requested):

test/performance/select_perf_test.lua
olegrok: It's an ancient artifact. Removed.

test/performance/select_perf_test.lua
olegrok: Currently it should be better. Fixed.

test/performance/select_perf_test.lua
olegrok: I suppose no. It could be timeout errors. It's possible case under high load.

ligurio (changes requested):

test/performance/select_perf_test.lua
ligurio: It could be timeout errors or other errors. Raw idea: at the end of test print happened errors with a number of occurrence and sort them by a number of occurence.

test/performance/select_perf_test.lua
```diff
@@ -0,0 +1,177 @@
...
+    end)
+    if not ok then
+        table.insert(report.errors, err)
+        log.error(err)
+    end
+end
+
+g.test_insert = function()
+    local timeout = 30
+    local fiber_count = 600
```
ligurio, line 107: These parameters are always static. Do we need to calculate these parameters dynamically depending on RAM (for example)?

ligurio (changes requested):

ligurio:
I run performance test that you made in https://github.com/tarantool/crud/pull/196 and got following results:

**With proposed patch:**

INSERT (Fibers count - 600, Connection count - 10, Timeout  - 30.000000):
1. Requests - 744582, RPS - 24819.400000
2. Requests - 748020, RPS - 24934.000000                                 
3. Requests - 754282, RPS - 25142.733333                                                                                                                                                

SELECT (Fibers count - 200, Connection count - 10, Timeout  - 30.000000)
1. Requests - 104720, RPS - 3490.666667  
2. Requests - 104000, RPS - 3466.666667  
3. Requests - 106060, RPS - 3535.333333  

**Without proposed patch:**

INSERT (Fibers count - 600, Connection count - 10, Timeout  - 30.000000):
1. Requests - 681047, RPS - 22701.566667                                                                                                                                                 
2. Requests - 712037, RPS - 23734.566667                                                                                                                                                
3. Requests - 702531, RPS - 23417.700000                                                                                                                                                

SELECT (Fibers count - 200, Connection count - 10, Timeout  - 30.000000)
1. Requests - 97340, RPS - 3244.666667
2. Requests - 102120, RPS - 3404.000000 
3. Requests - 98680, RPS - 3289.333333

test/performance/select_perf_test.lua
```diff
@@ -0,0 +1,177 @@
...
+    for i = 1, fiber_count do
+        fibers[i]:join()
+    end
+
+    log.error('SELECT')
+    log.error('Fibers count - %d', fiber_count)
+    log.error('Connection count - %d', connection_count)
+    log.error('Timeout  - %f', timeout)
+    log.error('Requests - %d', report.count)
+    log.error('Errors - %s', #report.errors)
```
ligurio: `#report.errors` is a number, why specifier is for string?

ligurio (changes requested):

test/performance/select_perf_test.lua
olegrok: I don't know how we could calculate it. We don't consume huge amount of RAM and all transactions are processed in single thread.

test/performance/select_perf_test.lua
olegrok: It was a typo

ligurio (approved):

ligurio:
LGTM. Thanks!

test/entrypoint/srv_select.lua
```diff
@@ -13,6 +13,11 @@ package.preload['customers-storage'] = function()
     return {
         role_name = 'customers-storage',
         init = function()
+            rawset(_G, 'tarantool_tuning', function()
```
olegrok, line 16: cartridge.cfg accepts options as the second argument

PR #200, opened 2021-08-17
Fix opts table damaging by CRUD methods

### LINKED ISSUE #192: CRUD methods damage opts table
If make call like this: crud.upsert_object(space_name, tuple, operations, opts) then if try to call crud.get(space_name, primary_key, opts) with the same opts variable you will get an error: .rocks/share/tarantool/crud/get.lua:120: unexpected argument opts.add_space_schema_hash to get

AnaNek:
If we have `opts` variable and we pass it to some crud method
(`upsert_object` for example) and then use `opts` variable again
for crud method (`get` for example) we receive an error because
`*_object` method damages `opts` variable adding
`add_space_schema_hash` parameter to `opts` variable.

Closes #192


test/integration/simple_operations_test.lua
```diff
@@ -1007,3 +1007,139 @@ pgroup:add('test_partial_result_bad_input', function(g)
     t.assert_str_contains(err.err, 'Space format doesn\'t contain field named "lastname"')
 end)
 
+pgroup:add('test_opts_not_damaged', function(g)
```
ligurio, line 1010: I have reverted changes for CRUD and etst failed on asserts that checks opts. I have commented out these checks:
```lua
--- a/test/integration/simple_operations_test.lua
+++ b/test/integration/simple_operations_test.lua
@@ -1036,7 +1036,7 @@ pgroup:add('test_opts_not_damaged', function(g)
     ]], {insert_opts})
 
     t.assert_equals(err, nil)
-    t.assert_equals(new_insert_opts, insert_opts)
+    --t.assert_equals(new_insert_opts, insert_opts)
 
     -- upsert
     local upsert_opts = {timeout = 1, bucket_id = 907, fields = {'name', 'age'}}
@@ -1068,7 +1068,7 @@ pgroup:add('test_opts_not_damaged', function(g)
     ]], {upsert_opts})
 
     t.assert_equals(err, nil)
-    t.assert_equals(new_upsert_opts, upsert_opts)
+    --t.assert_equals(new_upsert_opts, upsert_opts)
 
     -- get
     local get_opts = {timeout = 1, bucket_id = 401, fields = {'name', 'age'}}
@@ -1128,7 +1128,7 @@ pgroup:add('test_opts_not_damaged', function(g)
     ]], {replace_opts})
 
     t.assert_equals(err, nil)
-    t.assert_equals(new_replace_opts, replace_opts)
+    --t.assert_equals(new_replace_opts, replace_opts)
 
     -- delete
     local delete_opts = {timeout = 1, bucket_id = 401, fields = {'name', 'age'}}
```
and test passed. Why?
I expected that test should fail with an error described in #192: "unexpected argument opts.add_space_schema_hash to get".

ligurio (changes requested):
Thanks for the patch!

test/integration/simple_operations_test.lua
```diff
@@ -1007,3 +1007,139 @@ pgroup:add('test_partial_result_bad_input', function(g)
...
+
+        return insert_opts, err
+    ]], {insert_opts})
+
+    t.assert_equals(err, nil)
+    t.assert_equals(new_insert_opts, insert_opts)
+
+    -- insert_object
+    local insert_opts = {timeout = 1, bucket_id = 477, fields = {'name', 'age'}}
+    local new_insert_opts, err = g.cluster.main_server.net_box:eval([[
```
ligurio, line 1028: Why do you use `eval()` instead of `call()`?

```lua
    -- insert_object
    local result, err = g.cluster.main_server.net_box:call(
            'crud.insert_object', {
                'customers',
                {id = 1, name = 'Fedor', age = 59},
                insert_opts
            }
    )
```

ligurio (changes requested):

test/integration/simple_operations_test.lua
AnaNek: This error will occur if damaged options (ex. `upsert_opts`) to the other method (ex. `get`) that does not accept `add_space_schema_hash` as an opts table field. In this test I don't pass opts table from one method to the other, I am just checking that opts table was not modified by each of CRUD method.

test/integration/simple_operations_test.lua
AnaNek: If I would use `call()` I could not catch modification of opts table

crud/insert.lua
```diff
@@ -147,7 +147,12 @@ function insert.object(space_name, obj, opts)
     opts = opts or {}
 
     -- insert can fail if router uses outdated schema to flatten object
-    opts.add_space_schema_hash = true
+    opts = {
+        timeout = opts.timeout,
```
ligurio, line 151: If then original `opts` is nil then `opts.timeout`, `opts.bucket_id`, `opts.fields` are equal to `nil` too and adding them to table is useless.
If the original `opts` is not `nil` then we reassign original {timeout, bucket_id, fields} values to the same key...

Didn't get the idea behind it.

ligurio (changes requested):

test/integration/simple_operations_test.lua
ligurio: I would reproduce the original problem in a test and then fix it in a code. Right now (and in the future) we cannot guarantee that problem has been fixed.

ligurio (changes requested):

crud/insert.lua
AnaNek: Before this fix original opts table was modified by adding a key `add_space_schema_hash`. To avoid this modification we need to copy the values of original opts table, that's why I use this construction `opts = {timeout = opts.timeout, ...}` (there is another method: table.copy(opts), but I thought it that `opts` is not huge or complex table to perform `table.copy()`)

crud/upsert.lua
```diff
@@ -160,7 +160,12 @@ function upsert.object(space_name, obj, user_operations, opts)
     opts = opts or {}
 
     -- upsert can fail if router uses outdated schema to flatten object
-    opts.add_space_schema_hash = true
+    opts = {
```
olegrok, line 163: I suggest to implement `make_options` function to avoid copy-paste.

test/integration/simple_operations_test.lua
```diff
@@ -1007,3 +1007,141 @@ pgroup:add('test_partial_result_bad_input', function(g)
...
+        local delete_opts = ...
+
+        local _, err = crud.delete('customers', 2, delete_opts)
+
+        return delete_opts, err
+    ]], {delete_opts})
+
+    t.assert_equals(err, nil)
+    t.assert_equals(new_delete_opts, delete_opts)
+end)
```
olegrok, line 1147: I suggest to add the same test for "select/pairs" functions. Probably this bug doesn't affect them but I believe we should cover them all.

olegrok (commented):
Thanks for your patch. See my two suggestions.

crud/insert.lua
```diff
@@ -147,7 +147,7 @@ function insert.object(space_name, obj, opts)
     opts = opts or {}
 
     -- insert can fail if router uses outdated schema to flatten object
-    opts.add_space_schema_hash = true
+    opts = utils.make_options(opts, true)
```
olegrok, line 150: Probably we don't need `opts = opts or {}`.

And make_options could be rewritten in following way:

```lua
if opts == nil then
    return {add_space_schema_hash  = add_space_schema_hash_value}
end

return {...}
```

test/integration/len_test.lua
```diff
@@ -62,3 +62,19 @@ pgroup:add('test_len_empty_space', function(g)
     t.assert_equals(err, nil)
     t.assert_equals(result, 0)
 end)
+
+pgroup:add('test_opts_not_damaged', function(g)
+    local len_opts = {timeout = 1}
+    local new_len_opts, err = g.cluster.main_server.net_box:eval([[
```
olegrok, line 68: After luatest update we could write `main_server:eval` without net_box (https://github.com/tarantool/luatest/commit/5d424209b42ad56bd15840f3d826567f10df0ffe)

crud/common/utils.lua
```diff
@@ -524,4 +524,17 @@ function utils.flatten_obj_reload(space_name, obj)
     return schema.wrap_func_reload(flatten_obj, space_name, obj)
 end
 
+function utils.make_options(opts, add_space_schema_hash_value)
```
mRrvz, line 527: It seems to me that the word ``create`` is more suitable here (not ``make``)

olegrok (approved):

crud/common/utils.lua
AnaNek: `make_option` name is often used for parsing options (in Python modules for [example](https://docs.python.org/3/library/optparse.html#populating-the-parser)). Thats why it seems to me that this name is suitable.

mRrvz (approved):

test/integration/simple_operations_test.lua
AnaNek: But the problem was in modification `opts` parameter inside `crud` methods and tests check that passed `opts` parameter and received `opts` parameter are the same.

ligurio (approved):

crud/common/utils.lua
Totktonada: A bit more bikeshedding: it looks specific for the `add_space_schema_hash` option, so it would be better either choose a specific name or rewrite the function to be common:

```lua
-- Merge two options map.
--
-- `a` and/or `b` can be `nil`.
--
-- If `a.foo` and `b.foo` exists, prefer `b.foo`.
function utils.merge_options(a, b)
    return fun.chain(a or {}, b or {}):tomap()
end
```

So the call will be the following:

```diff
-opts = utils.make_options(opts, true)
+opts = utils.merge_options(opts, {add_space_schema_hash = true})
```

Maybe a bit more Lua tables, but the call becomes more readable (now we see what is the true/false value).

AnaNek: Updated:

```diff
+-- Merge two options map.
+--
+-- `opts_a` and/or `opts_b` can be `nil`.
+--
+-- If `opts_a.foo` and `opts_b.foo` exists, prefer `opts_b.foo`.
+function utils.merge_options(opts_a, opts_b)
+    return fun.chain(opts_a or {}, opts_b or {}):tomap()
+end

```

PR #206, opened 2021-09-01
test: use appropriate asserts

ligurio:
luatest has a wide range of asserts [1] that checks actual and expected
values. Using these asserts instead of conditions in pure assert()
improves readability and provide more details in log when assert
triggered. This patch replaces asserts with conditions inside to
appropriate luatest's assert function.

1. https://github.com/tarantool/luatest#list-of-luatest-functions

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Closes #???


mRrvz (approved):
LGTM

PR #198, opened 2021-08-14
Update Tarantool versions in test matrix

ligurio:
Tarantool 2.8.1 [1] and Tarantool 2.9 are available [2], let's include
it to test matrix. Test
simple_operations[memtx].test_intermediate_nullable_fields_update has
failed on 2.8 [3] and to crash looks similar to a bug described in [4].
Testcase skipped on 2.8 to workaround it before next release.

Tarantool 2.4-6 has reached EOL, so these version has been removed from
test matrix.

1. https://github.com/tarantool/tarantool/releases/tag/2.8.1
2. https://github.com/tarantool/tarantool/releases/tag/2.9.0
3. https://github.com/tarantool/crud/pull/195/checks?check_run_id=3320469716
4. https://github.com/tarantool/tarantool/issues/6069

What has been done? Why? What problem is being solved?

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Closes #???

PR #210, opened 2021-09-03
Describe supported Tarantool versions with limitations

### LINKED ISSUE #217: Describe supported Tarantool versions with limitations

ligurio:

I didn't forget about

- [ ] ~~Tests~~
- [ ] Changelog (?)
- [ ] ~~Documentation~~

Closes #217

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,24 +58,26 @@ CRUD module before use:
 1. Differences from the built-in merger module - https://github.com/tarantool/tuple-merger#backward-and-forward-compatibility-guarantees
 2. Differences from the built-in key_def module - https://github.com/tarantool/tuple-keydef#differences-from-the-built-in-module
 
-For more information see the table below that describes Tarantool's versions
-and features for what CRUD module depends on.
-
 <details>
   <summary>Support of features used by CRUD in Tarantool versions</summary>
 
+---
+
+For more information see the table below that describes Tarantool's versions
+and features for what CRUD module depends on.
+
 | Branch | merger                  | tuple-merger | key_def      | tuple-keydef | UUID         | JSON paths  | JSON path updates | abs. nullable fields  |
 |--------|-------------------------|--------------|--------------|--------------|--------------|-------------|-------------------|-----------------------|
-| 1.10   | Unsupported             | 1.10.7-91    | Unsupported  | 1.10.7-85    | Unsupported  | Unsupported | Unsupported       | Partially<sup>1</sup>
-| 2.2    | Unsupported<sup>1</sup> | Unsupported  | All versions | Unsupported  | Unsupported  | Unsupported | Unsupported       | Partially<sup>1</sup>
-| 2.3    | 2.3.2-99<sup>1</sup>    | Unsupported  | All versions | Unsupported  | Unsupported  | Unsupported | 2.3.0-316         | Partially<sup>1</sup>
-| 2.5    | 2.5.0-275<sup>1</sup>   | 2.5.1-150    | All versions | 2.5.1-145    | All versions | Unsupported | All versions      | Partially<sup>1</sup>
-| 2.6    | All versions            | 2.6.0-193    | All versions | 2.6.0-188    | All versions | 2.6.2-99    | All versions      | Partially<sup>1</sup>
-| 2.7    | All versions            | All versions | All versions | All versions | All versions | 2.7.1-102   | All versions      | Partially<sup>1</sup>
+| 1.10   | Unsupported             | 1.10.7-91    | Unsupported  | 1.10.7-85    | Unsupported  | Unsupported | Unsupported       | Partially<sup>2</sup>
+| 2.2    | Unsupported<sup>1</sup> | Unsupported  | All versions | Unsupported  | Unsupported  | Unsupported | Unsupported       | Partially<sup>2</sup>
+| 2.3    | 2.3.2-99<sup>1</sup>    | Unsupported  | All versions | Unsupported  | Unsupported  | Unsupported | 2.3.0-316         | Partially<sup>2</sup>
+| 2.5    | 2.5.0-275<sup>1</sup>   | 2.5.1-150    | All versions | 2.5.1-145    | All versions | Unsupported | All versions      | Partially<sup>2</sup>
+| 2.6    | All versions            | 2.6.0-193    | All versions | 2.6.0-188    | All versions | 2.6.2-99    | All versions      | Partially<sup>2</sup>
+| 2.7    | All versions            | All versions | All versions | All versions | All versions | 2.7.1-102   | All versions      | Partially<sup>2</sup>
 | 2.8    | All versions            | All versions | All versions | All versions | All versions | 2.8.0-113   | All versions      | 2.8.0-307
 
-1. merger has been added in 2.2.0-268-gc7915eccd, however support was limited
-   due to [#4954](https://github.com/tarantool/tarantool/issues/4954).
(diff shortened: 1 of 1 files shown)


ligurio:
@Totktonada gentle reminder

README.md
```diff
@@ -8,6 +8,47 @@ The `CRUD` module allows to perform CRUD operations on the cluster.
...
+- 2.3: since 2.3.2-99-g487289ad2 with key_def and merger modules, without UUID
+  keys comparison and without Jsonpath indexes
+- 2.5: since 2.5.0-275-g9be50ae74 with key_def and merger modules and without
+  Jsonpath indexes
+- 2.6: since 2.6.2-99-gf92b9e21
+- 2.7: since 2.7.1-102-g84edc33fb
+- 2.8: since 2.8.0-113-gb3d4648d8
+
+1. Differences from the built-in merger module - https://github.com/tarantool/tuple-keydef#differences-from-the-built-in-module
+2. Differences from the built-in key_def module - https://github.com/tarantool/tuple-keydef#differences-from-the-built-in-module
```
Totktonada, line 34: The same link is written twice.

README.md
```diff
@@ -8,6 +8,47 @@ The `CRUD` module allows to perform CRUD operations on the cluster.
 It also provides the `crud-storage` role for
 [Tarantool Cartridge](https://github.com/tarantool/cartridge).
 
+# Supported Tarantool versions
```
Totktonada, line 11: The idea to describe supported versions appears when we discussed [this](https://github.com/tarantool/crud/pull/198) change. I had the question about crud.min() / crud.max(): on which tarantool versions the module has no support of those function. This information is not present here.

README.md
```diff
@@ -8,6 +8,47 @@ The `CRUD` module allows to perform CRUD operations on the cluster.
...
+1. Differences from the built-in merger module - https://github.com/tarantool/tuple-keydef#differences-from-the-built-in-module
+2. Differences from the built-in key_def module - https://github.com/tarantool/tuple-keydef#differences-from-the-built-in-module
+
+For more information see the table below that describes Tarantool's versions
+and features for what CRUD module depends on:
+
+| Branch | merger                           | tuple-merger         | key_def      | tuple-keydef         | UUID         | JSON paths            |
+|--------|----------------------------------|----------------------|--------------|----------------------|--------------|-----------------------|
+| 1.10   | Unsupported                      | 1.10.7-91-g9ee14eca5 | Unsupported  | 1.10.7-85-g840c13293 | Unsupported  | Unsupported           |
+| 2.2    | 2.2.0-268-gc7915eccd             | Unsupported          | All versions | Unsupported          | Unsupported  | Unsupported           |
```
Totktonada, line 42: The primary crud usage is from a connector, so I would shift accents: the built-in merger is working since 2.3.2-99-g487289ad2 (==2.4.1-112-g2e83c2f02, ==2.5.0-275-g9be50ae74), everything before is not a working merger. I mean, place 2.3.2-99 and so on to the table and mention that there was very limited merger support prior to those versions.

README.md
```diff
@@ -8,6 +8,47 @@ The `CRUD` module allows to perform CRUD operations on the cluster.
...
+CRUD depends on Tarantool features like key_def, merger, UUID modules and JSON
+paths support. We do our best to provide uniform support for all supported
+Tarantool version with help of compatibility layer, runtime feature detection
+etc. but anyway some Tarantool versions has limited support in CRUD. Please
+consult with a list below to choose a Tarantool version compatible with CRUD
+module before use:
+
+- 1.10: since 1.10.7-91-g9ee14eca5 with tuple-merger<sup>1</sup> and
+  tuple-keydef<sup>2</sup> modules, without UUID keys comparison and without
+  Jsonpath indexes
```
Totktonada, line 22: I don't see information about minimal tarantool version I can use with crud. It generally works without keydef and merger (built-in or external), correct? So, say, I can use it on 1.10.3 if I don't want JSON path indexes, UUID, decimal and sorting by a collation different than 'unicode' or 'unicode_ci'. Correct?

After PR #198 this rule splits to two:

* For all functions except min/max, <..as stated above..>.
* For min / max there is additional requirement <...>.

The compatibility guarantees were changed and I want to know how exactly. Any formal unambiguous form to present this information is okay.

README.md
```diff
@@ -8,6 +8,47 @@ The `CRUD` module allows to perform CRUD operations on the cluster.
...
+- 2.7: since 2.7.1-102-g84edc33fb
+- 2.8: since 2.8.0-113-gb3d4648d8
+
+1. Differences from the built-in merger module - https://github.com/tarantool/tuple-keydef#differences-from-the-built-in-module
+2. Differences from the built-in key_def module - https://github.com/tarantool/tuple-keydef#differences-from-the-built-in-module
+
+For more information see the table below that describes Tarantool's versions
+and features for what CRUD module depends on:
+
+| Branch | merger                           | tuple-merger         | key_def      | tuple-keydef         | UUID         | JSON paths            |
```
olegrok, line 39: Jsonpath updates is missed in this table. Absent nullable field update is also missed (https://github.com/tarantool/tarantool/issues/3378). However it's even partially supported for 1.10+ (https://github.com/tarantool/crud/commit/6c304839652be686e76495da297b5a14879efc41).

README.md
ligurio: Fixed.

README.md
ligurio: Just for context: commits that remove `crud.mix()`/`crud.max()` support in previous version:
- https://github.com/tarantool/crud/commit/5edf0274f28a2e607d0f34fa3423d00ce1d44f14
- https://github.com/tarantool/crud/commit/56247f27239e1f2e92701c9ad159e3977b512a79

ligurio: These commit reverted in https://github.com/tarantool/crud/pull/215

README.md
ligurio: Updated footnote.

README.md
ligurio: Added.

README.md
ligurio: The initial questions was "What is a minimal working version of Tarantool for CRUD?". I have prepared a table with information (for us) about Tarantool features used by CRUD in Tarantool versions and additionally a list with Tarantool versions starting from user can use CRUD.

filonenko-mikhail:
What is the reason to make so big information noise on the first Readme lines?

ligurio:
@filonenko-mikhail There are many different point of views how it looks in a README. Yours is only one of these. I think that a first question which should be answered here is "why and who does need such information in a readme?" and then it should be easier to answer to yours question.

CRUD depends on different Tarantool features  that were supported in different Tarantool versions. We decided to describe what CRUD features works on tarantool versions and how good this support is. Imagine you are tarantool's user and you want to start using crud on tarantool 1.10, but you dont' want to upgrade a major version. In such case you may go to our table and read that CRUD supports Tarantool since 1.10.7, it doesn't have uuid keys comparison support and doesn't have UUID support. Looks as a useful information for CRUD's users. This information available as an additional section in the README.

2. Due to a limited support of Tarantool's features by CRUD in different Tarantool's versions we need to understand how source code of CRUD maps to support of Tarantool versions. Take a look at compat layer in repository, it is full of conditions for different versions. The goal of added table is to help us to control compat layer without pain. I think that table may be interested to a small number of users so I have added it under spoiler.

filonenko-mikhail:
What is the plan to support such information up to date?
And if the plan is exists, why this pr still not in master?

PR #214, opened 2021-09-16
Bump luatest version

ligurio:
New version 0.5.5 brings:

- Repeat `_each` and `_test` hooks when `--repeat` is specified.
- Add group parametrization.

1. https://github.com/tarantool/luatest/commit/717cf1d03440f1134fbd8010775084878c1c3949

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -4,7 +4,7 @@
 set -e
 
 # Test dependencies:
-tarantoolctl rocks install luatest 0.5.4
+tarantoolctl rocks install luatest 0.5.5
 tarantoolctl rocks install luacheck 0.25.0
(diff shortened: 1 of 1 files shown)


deps.sh
```diff
@@ -4,7 +4,7 @@
 set -e
 
 # Test dependencies:
-tarantoolctl rocks install luatest 0.5.4
+tarantoolctl rocks install luatest 0.5.5
```
olegrok, line 7: If there is a group parametrization we probably should use it instead of pgroup.

ligurio: I know :) I will replace our own implementation with luatest's implementation as soon as I'll find a time for it.

olegrok (approved):

PR #218, opened 2021-09-27
Fix passing timeout option for select/pairs

### LINKED ISSUE #209: Select/pairs error with using timeout
If I pass timeout as an option for select or pairs I get an error:

Select

```lua
    local crud = require('crud')

    local res, err = crud.select('customers', nil, {timeout = 1})
        
    if err ~= nil then
        return err
    end

    return nil
```

Pairs

```lua
    local crud = require('crud')

    local results = {}
    for _, result in crud.pairs('customers', nil, {timeout = 1}) do
        table.insert(results, result)
    end
        
    return results
```

Error:

```
.../crud/select/compat/select.lua:21: unexpected argument opts.timeout to func
```

AnaNek:
Before this patch passing `timeout` option to `crud.select()`/`crud.pairs()`
had been leading to an error `unexpected argument opts.timeout to func`.
Function `build_select_iterator` accepts `timeout` as a part of `call_opts`
table option. Before this patch function `build_select_iterator` had been
receiving `timeout` as a separate parameter which had been leading to a failure.

Closes #209

### DIFF
--- a/test/integration/select_test.lua
+++ b/test/integration/select_test.lua
@@ -1508,3 +1508,33 @@ pgroup.test_jsonpath_index_field_pagination = function(g)
     local objects = crud.unflatten_rows(result.rows, result.metadata)
     t.assert_equals(objects, helpers.get_objects_by_idxs(cars, {2, 3}))
 end
+
+pgroup.test_select_timeout = function(g)
+    local customers = helpers.insert_objects(g, 'customers', {
+        {
+            id = 1, name = "Elizabeth", last_name = "Jackson",
+            age = 12, city = "New York",
+        }, {
+            id = 2, name = "Mary", last_name = "Brown",
+            age = 46, city = "Los Angeles",
+        }, {
+            id = 3, name = "David", last_name = "Smith",
+            age = 33, city = "Los Angeles",
+        }, {
+            id = 4, name = "William", last_name = "White",
+            age = 81, city = "Chicago",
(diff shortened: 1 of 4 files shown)


mRrvz:
Please write in the description (and commit message) why these changes are needed.

test/integration/select_test.lua
```diff
@@ -1508,3 +1508,41 @@ pgroup.test_jsonpath_index_field_pagination = function(g)
...
+        {4, 1161, "William", "White", 81, "Chicago"},
+    })
+    t.assert_equals(result.metadata, {
+        {name = 'id', type = 'unsigned'},
+        {name = 'bucket_id', type = 'unsigned'},
+        {name = 'name', type = 'string'},
+        {name = 'last_name', type = 'string'},
+        {name = 'age', type = 'number'},
+        {name = 'city', type = 'string'},
+    })
```
mRrvz, line 1547: Is this check really necessary? (metadata)

olegrok (approved):
Looks good after @mRrvz comments will be resolved.

test/integration/select_test.lua
AnaNek: Removed.

Totktonada (approved):
On a brief glance it looks logical.

mRrvz (approved):

ligurio (approved):

PR #221, opened 2021-10-13
Update select.lua

filonenko-mikhail:
What has been done? Why? What problem is being solved?

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Closes #???

PR #181, opened 2021-07-06
Use sharding keys to calculate bucket id

ligurio:
What has been done? Why? What problem is being solved?

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #166

PR #222, opened 2021-10-14
select/pairs: don't ignore provided bucket_id

### LINKED ISSUE #220: Select with bucket_id specified leads to map-reduce
See full repro (cartridge app, insert.lua, loader.lua) attached.

Long story short:
`crud.select('space_1', {{'==', 'secondary', 's_test_1'}}, { bucket_id = vshard.router.bucket_id_mpcrc32('s_test_1' )})`
leads to map-reduce. I can see it in box.stat of both shards, and it starts throwing errors if I monkeypatch _G._crud.select_on_storage on one of the shards.

crud 0.8.0

Totktonada:
The bug is simple: crud ignores provided bucket_id, when unable to determine it itself. For example, when no conditions are given or when given condition involves a secondary index, which is not entirely in the  primary index.

It leads to incorrect select/pairs result: tuples are collected from all replicasets, while should be collected from one storage pointed by bucket_id. Second, it involves all replicasets into the request processing (performs map-reduce) that may dramatically drop performance.

The idea of the fix is suggested by Michael Filonenko in PR #221.

As a side effect a set of simple testing utils for collecting select statistics from storages was added.

See details in the commit messages.

Fixes #220

### DIFF
--- a/crud/select/compat/select.lua
+++ b/crud/select/compat/select.lua
@@ -65,8 +65,40 @@ local function build_select_iterator(space_name, user_conditions, opts)
     -- set replicasets to select from
     local replicasets_to_select = replicasets
 
-    if plan.sharding_key ~= nil and opts.force_map_call ~= true then
+    -- Whether to call one storage replicaset or perform
+    -- map-reduce?
+    --
+    -- If map-reduce is requested explicitly, ignore provided
+    -- bucket_id and fetch data from all storage replicasets.
+    --
+    -- Otherwise:
+    --
+    -- 1. If particular replicaset is pointed by a caller (using
+    --    the bucket_id option[^1]), crud MUST fetch data only
+    --    from this storage replicaset: disregarding whether other
+    --    storages have tuples that fit given condition.
+    --
+    -- 2. If a replicaset may be deduced from conditions
+    --    (conditions -> sharding key -> bucket id -> replicaset),
+    --    fetch data only from the replicaset. It does not change
+    --    the result[^2], but significantly reduces network
+    --    pressure.
+    --
+    -- 3. Fallback to map-reduce otherwise.
+    --
+    -- [^1]: We can change meaning of this option in a future,
+    --       see gh-190. But now bucket_id points a storage
+    --       replicaset, not a virtual bucket.
+    --
+    -- [^2]: It is correct statement only if we'll turn a blind
+    --       eye to resharding. However, AFAIU, the optimization
+    --       does not make the result less consistent (sounds
+    --       weird, huh?).
+    local perform_map_reduce = opts.force_map_call == true or
+        (opts.bucket_id == nil and plan.sharding_key == nil)
+    if not perform_map_reduce then
         local bucket_id = sharding.key_get_bucket_id(plan.sharding_key, opts.bucket_id)
+        assert(bucket_id ~= nil)
 
         local err
         replicasets_to_select, err = common.get_replicasets_by_sharding_key(bucket_id)
--- a/crud/select/compat/select_old.lua
+++ b/crud/select/compat/select_old.lua
@@ -118,8 +118,13 @@ local function build_select_iterator(space_name, user_conditions, opts)
     -- set replicasets to select from
     local replicasets_to_select = replicasets
 
-    if plan.sharding_key ~= nil and opts.force_map_call ~= true then
+    -- See explanation of this logic in
+    -- crud/select/compat/select.lua.
+    local perform_map_reduce = opts.force_map_call == true or
+        (opts.bucket_id == nil and plan.sharding_key == nil)
+    if not perform_map_reduce then
         local bucket_id = sharding.key_get_bucket_id(plan.sharding_key, opts.bucket_id)
+        assert(bucket_id ~= nil)
 
         local err
         replicasets_to_select, err = common.get_replicasets_by_sharding_key(bucket_id)
--- a/test/integration/pairs_test.lua
+++ b/test/integration/pairs_test.lua
@@ -5,6 +5,7 @@ local t = require('luatest')
 local crud_utils = require('crud.common.utils')
 
 local helpers = require('test.helper')
+local storage_stat = require('test.helpers.storage_stat')
 
 local pgroup = t.group('pairs', {
     {engine = 'memtx'},
@@ -25,6 +26,13 @@ pgroup.before_all(function(g)
     g.cluster:start()
 
     g.space_format = g.cluster.servers[2].net_box.space.customers:format()
+
+    helpers.call_on_storages(g.cluster, function(server)
+        server.net_box:eval([[
+            local storage_stat = require('test.helpers.storage_stat')
+            storage_stat.init_on_storage()
+        ]])
+    end)
 end)
 
 pgroup.after_all(function(g) helpers.stop_cluster(g.cluster) end)
@@ -758,15 +766,19 @@ end
 pgroup.test_opts_not_damaged = function(g)
     local customers = helpers.insert_objects(g, 'customers', {
         {
+            -- bucket_id is 477, storage is s-2
             id = 1, name = "Elizabeth", last_name = "Jackson",
             age = 12, city = "Los Angeles",
         }, {
+            -- bucket_id is 401, storage is s-2
             id = 2, name = "Mary", last_name = "Brown",
             age = 46, city = "London",
         }, {
+            -- bucket_id is 2804, storage is s-1
             id = 3, name = "David", last_name = "Smith",
             age = 33, city = "Los Angeles",
         }, {
+            -- bucket_id is 1161, storage is s-2
             id = 4, name = "William", last_name = "White",
             age = 46, city = "Chicago",
         },
@@ -775,7 +787,6 @@ pgroup.test_opts_not_damaged = function(g)
     table.sort(customers, function(obj1, obj2) return obj1.id < obj2.id end)
 
     local expected_customers = {
-        {id = 3, name = "David", age = 33},
         {id = 4, name = "William", age = 46},
     }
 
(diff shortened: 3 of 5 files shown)


test/entrypoint/srv_select.lua
```diff
@@ -192,6 +192,34 @@ package.preload['customers-storage'] = function()
...
+            -- Here we count requests.
+            local storage_stat_table = {
+                select_requests = 0,
+            }
+
+            -- Wrap select_on_storage() function.
+            local select_on_storage_saved = _G._crud.select_on_storage
+            _G._crud.select_on_storage = function(...)
+                local requests = storage_stat_table.select_requests
+                storage_stat_table.select_requests = requests + 1
```
ligurio, line 211: Didn't get it: where is `requests` defined? May be `select_stat_table.select_requests`?

test/entrypoint/srv_select.lua
```diff
@@ -192,6 +192,34 @@ package.preload['customers-storage'] = function()
     }
 end
 
+package.preload['statistics-storage'] = function()
+    return {
+        role_name = 'statistics-storage',
+        dependencies = {'cartridge.roles.crud-storage'},
+        init = function()
```
ligurio, line 199: Could you define function in helpers to allow reuse in other entrypoints? I want to use it in PR #181, but my tests uses another entrypoint (`srv_ddl.lua`).

test/integration/select_test.lua
```diff
@@ -1581,3 +1583,75 @@ pgroup.test_opts_not_damaged = function(g)
...
+            -- bucket_id is 1161, storage is s-2
+            id = 4, name = 'William', last_name = 'White',
+            age = 81, city = 'Chicago',
+        },
+    })
+
+    table.sort(customers, function(obj1, obj2) return obj1.id < obj2.id end)
+
+    local storages = {'s1-master', 's1-replica', 's2-master', 's2-replica'}
+    local stat_a = helpers.collect_storage_stat(g.cluster, storages)
```
ligurio, line 1613: I would replace `stat_a` with `stat_b` to mean "stat before" and "stat after".

test/integration/select_test.lua
```diff
@@ -1581,3 +1583,75 @@ pgroup.test_opts_not_damaged = function(g)
...
+
+    -- Case: no conditions, just bucket id.
+    local result, err = g.cluster.main_server.net_box:call('crud.select', {
+        'customers',
+        nil,
+        {bucket_id = 2804, timeout = 1},
```
ligurio, line 1619: Why do we need timeout here?

test/helper.lua
```diff
@@ -230,4 +237,77 @@ function helpers.get_other_storage_bucket_id(cluster, bucket_id)
...
+--
+-- {
+--     ['s-1'] = {
+--         select_requests = 1,
+--     },
+--     ['s-2'] = {
+--         select_requests = 0,
+--     },
+-- }
+function helpers.collect_storage_stat(cluster, aliases)
```
ligurio, line 267: I would use `checks()` here.

test/helper.lua
```diff
@@ -230,4 +237,77 @@ function helpers.get_other_storage_bucket_id(cluster, bucket_id)
...
+    local res = {}
+
+    helpers.call_on_servers(cluster, aliases, function(server)
```
ligurio, line 270: I would use `checks()` here.

test/helper.lua
```diff
@@ -230,4 +237,77 @@ function helpers.get_other_storage_bucket_id(cluster, bucket_id)
...
+    end)
+
+    return res
+end
+
+-- Difference between 'a' and 'b' storage statistics.
+--
+-- The return value structure is the same as for
+-- helpers.collect_storage_stat().
+function helpers.diff_storage_stat(a, b)
```
ligurio, line 294: I would use `checks()` here.

test/helper.lua
```diff
@@ -230,4 +237,77 @@ function helpers.get_other_storage_bucket_id(cluster, bucket_id)
...
+        for key, val in pairs(storage_stat) do
+            local old = res[replicaset_alias][key] or 0
+            res[replicaset_alias][key] = old + val
+        end
+    end)
+
+    return res
+end
+
+-- Difference between 'a' and 'b' storage statistics.
```
ligurio, line 290: `a` and `b` are replicasets, not storages. If I got it right.

test/integration/pairs_test.lua
```diff
@@ -775,7 +781,6 @@ pgroup.test_opts_not_damaged = function(g)
     table.sort(customers, function(obj1, obj2) return obj1.id < obj2.id end)
 
     local expected_customers = {
-        {id = 3, name = "David", age = 33},
```
ligurio, line 778: It is not obvious why expected value is changed.

test/entrypoint/srv_select.lua
Totktonada: I like the idea, will look how to better do so.

test/integration/select_test.lua
Totktonada: I borrowed it from some of other test cases, because I guess it is a protection from hang in CI (where logs may not be accessible till end of the job).

test/integration/select_test.lua
```diff
@@ -1581,3 +1583,75 @@ pgroup.test_opts_not_damaged = function(g)
...
+            select_requests = 0,
+        },
+    })
+
+    -- Case: EQ on secondary index, which is not in the sharding
+    -- index (primary index in the case).
+    local result, err = g.cluster.main_server.net_box:call('crud.select', {
+        'customers',
+        {{'==', 'age', 81}},
+        {bucket_id = 1161, timeout = 1},
```
ligurio, line 1641: the same question about timout option

test/helper.lua
Totktonada: `a` and `b` are statistics objects, where each contains statistics for several replicasets (each replicaset entry may contain several metrics, but only 'select_requests' for now). I think I can name an object of this kind just 'storage statistics'.

The example (it is referenced just below in this comment) describes the return value structure and it should be enough to make it clear how to use the object. At least it looks more clear for me than the description I wrote here :)

test/integration/select_test.lua
Totktonada: Re-read your question. Nope, 'a' is just 'a', 'b' is just 'b' and so on: just three points. I can name it stat_1, stat_2 and stat_3 if you find it more preferable.

test/helper.lua
Totktonada: Now I'm going to drop the 'aliases' parameter, because we can determine storage nodes automatically by presence of the 'crud-storage' role. I can check that 'cluster' is a table, okay.

test/helper.lua
Totktonada: However, okay, if we'll touch the helper... I'll add a check here just in case.

crud/select/compat/select_old.lua
```diff
@@ -118,7 +118,8 @@ local function build_select_iterator(space_name, user_conditions, opts)
     -- set replicasets to select from
     local replicasets_to_select = replicasets
 
-    if plan.sharding_key ~= nil and opts.force_map_call ~= true then
+    local bucket_id_is_known = plan.sharding_key ~= nil or opts.bucket_id ~= nil
+    if bucket_id_is_known and opts.force_map_call ~= true then
```
ligurio, line 122: `bucket_id_is_known = plan.sharding_key ~= nil ...`
I think that `plan.sharding_key ~= nil` doesn't guarantee that bucket id is known. I would calculate a bucket_id and then: `bucket_id_is_known = bucket_id ~= nil or opts.bucket_id ~= nil`. What do you think?

```lua
    local bucket_id                                                                 
    if plan.sharding_key ~= nil then                                                
        bucket_id = sharding.key_get_bucket_id(plan.sharding_key, opts.bucket_id)
    end                                                                             
                                                                                    
    local bucket_id_is_known = bucket_id ~= nil or opts.bucket_id ~= nil            
    if bucket_id_is_known and opts.force_map_call ~= true then                      
        local err                                                                   
        replicasets_to_select, err = common.get_replicasets_by_sharding_key(bucket_id)
        if err ~= nil then                                                          
            return nil, err, true                                                   
        end                                                                      
    end 
```

test/helper.lua
```diff
@@ -179,6 +179,76 @@ function helpers.call_on_servers(cluster, aliases, func)
     end
 end
 
+-- Call given function for each server with the 'crud-storage'
```
ligurio, line 182: Minor: could you mention in commit message an issue #224?

crud/select/compat/select_old.lua
ligurio: > It is subtle difference: if we know a sharding key, we know a bucket id in fact (just by applying the sharding function).

What will happen if a function for bucket id calculation return not a number?

Totktonada: I stopped on the latest option now. Please, check it out.

Totktonada: (I didn't try to brute force you. Just a bit puzzled how to better structurize this code.)

crud/select/compat/select.lua
```diff
@@ -65,8 +65,40 @@ local function build_select_iterator(space_name, user_conditions, opts)
...
+    --
+    -- 2. If a replicaset may be deduced from conditions
+    --    (conditions -> sharding key -> bucket id -> replicaset),
+    --    fetch data only from the replicaset. It does not change
+    --    the result[^2], but significantly reduces network
+    --    pressure.
+    --
+    -- 3. Fallback to map-reduce otherwise.
+    --
+    -- [^1]: We can change meaning of this option is a future,
```
ligurio, line 89: 'in a future'

ligurio (approved):
Thanks! LGTM

crud/select/compat/select.lua
Totktonada: Thanks, fixed.

Issue #225, opened 2021-10-18
Create a test with comparison vshard.call() and CRUD performance


ligurio:
CRUD uses `vshard.call()` for executing CRUD operations on storage. It would be interesting to compare running, for example, SELECT and INSERT via `vshard.call()` and `crud:select()` and `crud:insert()`. Comparison will show us an overhead that CRUD adds to operations on storage.

The outcome of the task is a test with at least inserts and selects using `vshard.call()` and CRUD.

Totktonada:
NB: Consider the test case provided by Dmitry in #220. I measured PR #222 and PR #226 using it.

PR #226, opened 2021-10-19
merger: don't create key_def/merger for one source

Totktonada:
A merge source has the same API as merger itself, so there is no
difference in behaviour between a merger created from one source and
this source itself. However there is no overhead for creating key_def,
merger and passing tuples over the merger.

This optimization gives me 13% boost on the case from #220.

Follows up #220

----

There is no simple way to check it with a test, so I skipped it. Hopefully we'll see a difference after #225. I tested the patch manually for now.


ligurio:
We definitely need a performance regression test that allow us to avoid degradation in the future. If I get it right, we will address it in #225. Right?

ligurio (approved):
LGTM

PR #150, opened 2021-04-19
Select: jsonpath support

mRrvz:
Added jsonpath support for select. Part of #12

PR #227, opened 2021-10-19
Fix merger support recognition on tarantool 2.10+

Totktonada:
We cannot use just string comparison as before, because '2.10.0-<...>'
less than, say, '2.3' lexicographically. The module had to use the old
select implementation on tarantool 2.10 due to this problem.

Implemented more accurate checks, whether built-in and external
tuple-merger are supported. Say, it'll not enable the built-in merger on
tarantool 2.4.1, where the module has a known problem (see comments in
the code).

Fixed the problem of the same kind in a test case.

### DIFF
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -210,6 +210,35 @@ local function determine_enabled_features()
     -- since Tarantool 2.6.3 / 2.7.2 / 2.8.1
     enabled_tarantool_features.jsonpath_indexes = major >= 3 or (major >= 2 and ((minor >= 6 and patch >= 3)
         or (minor >= 7 and patch >= 2) or (minor >= 8 and patch >= 1) or minor >= 9))
+
+    -- The merger module was implemented in 2.2.1, see [1].
+    -- However it had the critical problem [2], which leads to
+    -- segfault at attempt to use the module from a fiber serving
+    -- iproto request. So we don't use it in versions before the
+    -- fix.
+    --
+    -- [1]: https://github.com/tarantool/tarantool/issues/3276
+    -- [2]: https://github.com/tarantool/tarantool/issues/4954
+    enabled_tarantool_features.builtin_merger =
+        (major == 2 and minor == 3 and patch >= 3) or
+        (major == 2 and minor == 4 and patch >= 2) or
+        (major == 2 and minor == 5 and patch >= 1) or
+        (major >= 2 and minor >= 6) or
+        (major >= 3)
+
+    -- The external merger module leans on a set of relatively
+    -- new APIs in tarantool. So it works only on tarantool
+    -- versions, which offer those APIs.
+    --
+    -- See README of the module:
+    -- https://github.com/tarantool/tuple-merger
+    enabled_tarantool_features.external_merger =
+        (major == 1 and minor == 10 and patch >= 8) or
+        (major == 2 and minor == 4 and patch >= 3) or
+        (major == 2 and minor == 5 and patch >= 2) or
+        (major == 2 and minor == 6 and patch >= 1) or
+        (major == 2 and minor >= 7) or
+        (major >= 3)
 end
 
 function utils.tarantool_supports_fieldpaths()
@@ -234,6 +263,22 @@ function utils.tarantool_supports_jsonpath_indexes()
     end
 
     return enabled_tarantool_features.jsonpath_indexes
+end
+
+function utils.tarantool_has_builtin_merger()
+    if enabled_tarantool_features.builtin_merger == nil then
+        determine_enabled_features()
+    end
+
+    return enabled_tarantool_features.builtin_merger
+end
+
+function utils.tarantool_supports_external_merger()
+    if enabled_tarantool_features.external_merger == nil then
+        determine_enabled_features()
+    end
+
+    return enabled_tarantool_features.external_merger
 end
 
 local function add_nullable_fields_recursive(operations, operations_map, space_format, tuple, id)
--- a/test/helper.lua
+++ b/test/helper.lua
@@ -300,4 +300,27 @@ function helpers.get_other_storage_bucket_id(cluster, bucket_id)
     ]], {bucket_id})
 end
 
+function helpers.tarantool_version_at_least(wanted_major, wanted_minor,
+        wanted_patch)
+    -- Borrowed from `determine_enabled_features()` from
+    -- crud/common/utils.lua.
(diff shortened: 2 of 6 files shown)


test/helper.lua
```diff
@@ -230,4 +230,46 @@ function helpers.get_other_storage_bucket_id(cluster, bucket_id)
     ]], {bucket_id})
 end
 
+local function parse_tarantool_version(component)
```
Totktonada, line 233: This code may be simpler if we're interested in just three components. See `crud/common/utils.lua`, `determine_enabled_features()` for example.

Totktonada: Simplified now.

crud/common/utils.lua
```diff
@@ -210,6 +210,35 @@ local function determine_enabled_features()
...
+    -- fix.
+    --
+    -- [1]: https://github.com/tarantool/tarantool/issues/3276
+    -- [2]: https://github.com/tarantool/tarantool/issues/4954
+    enabled_tarantool_features.builtin_merger =
+        (major == 2 and minor == 3 and patch >= 3) or
+        (major == 2 and minor == 4 and patch >= 2) or
+        (major == 2 and minor == 5 and patch >= 1) or
+        (major >= 2 and minor >= 6) or
+        (major >= 3)
```
olegrok, line 227: Some conditions above won't work for Tarantool 3.0. E.g. "enabled_tarantool_features.uuids" and "enabled_tarantool_features.fieldpaths".

olegrok (approved):
Thanks for your patch. I left several comments. If you don't want to fix version checks for "fieldpaths" and "uuids", please file an issue.

test/helper.lua
```diff
@@ -300,4 +300,27 @@ function helpers.get_other_storage_bucket_id(cluster, bucket_id)
     ]], {bucket_id})
 end
 
+function helpers.tarantool_version_at_least(wanted_major, wanted_minor,
```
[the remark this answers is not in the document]
Totktonada: Do you propose to cover testing code with tests?

crud/common/utils.lua
Totktonada: Do you want to reformat other conditions in the way I use here? I can do in a separate commit.

test/helper.lua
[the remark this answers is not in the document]
olegrok: Oh, I missed that it's a test helper. No, it's ok. My suggestion is not relevant.

crud/common/utils.lua
olegrok: Yes. I think other conditions should support next major Tarantool versions.
I suggest to do it in this patch in separate commit or even a bit later in separate PR.

Totktonada: Aside of obvious problems:

* `utils.tarantool_supports_fieldpaths()` and `utils.tarantool_supports_uuids()` are incorrect for tarantool-3* (tarantool-4* and so on).
* `utils.tarantool_supports_uuids()` and `utils.tarantool_supports_jsonpath_indexes()` are used only in tests.

There is one interesting:

```lua
-- since Tarantool 2.6.3 / 2.7.2 / 2.8.1
enabled_tarantool_features.jsonpath_indexes = <...>
```

However JSON path indexes were introduced in 2.1.2, see the following links:

https://github.com/tarantool/tarantool/issues/1012
https://github.com/tarantool/tarantool/commit/4273ec52e122d6d37c8deedf1bc10732a7e40c0e
https://github.com/tarantool/tarantool/commit/a754980d7feab110b8c82ee15ef13e080afa2882

I re-checked 2.6.3, 2.7.2 and 2.8.1 release notes, there is no 'json' word in them:

https://github.com/tarantool/tarantool/releases/2.6.3
https://github.com/tarantool/tarantool/releases/2.7.2
https://github.com/tarantool/tarantool/releases/2.8.1

This code was added in PR #158 and I don't see anything that would clarify why those versions are choosen. I looked at the issue as well: #12 and...

> Support JSON path
>
> Even Tarantool 1.10 supports value extraction via jsonpath. So we could patch our filter generator and add jsonpath's support of jsonpath conditions.

It is not about JSON path indexes? It is about filtering values using filters that use JSON path notation? How '1.10 supports value extraction via jsonpath'? Or it means that we can parse such notation using Lua without tarantool side support?

(What a hell...)

AFAIU from code and tests in PR #158 there was a problem with `crud.select()` with index name in a condition, when this index is constructed by a JSON path. The problem was resolved by using the `key_def`/`tuple.keydef` module to correctly extract the key and it will not work if the module is not present.

(Ugh... Why not describe things for humans in a changelog? That's my regular toothache: convince others that development is about skill to share knowledge rather than about coding.)

If my statements are correct, then we need the following property in *tests*: whether `<keydef object>:extract_key()` supports JSON paths. If so, run the tests. Otherwise a user will receive an error (I hope, but it is not tested). Ideally we should test all possible variants of 'keydef supports json path' and 'json path index is supported' properties.

It still doesn't correspond to 2.6.3/2.7.2/2.8.1 anyhow.

I don't know what to do here, got tired, will think later.

olegrok: > It is not about JSON path indexes? It is about filtering values using filters that use JSON path notation? How '1.10 supports value extraction via jsonpath'? Or it means that we can parse such notation using Lua without tarantool side support?

There were two parts of this task.
The first is json path in tuple. This feature was introduced in 1.10 (https://github.com/tarantool/tarantool/issues/1285).
The second is json path indexes (https://github.com/tarantool/tarantool/issues/1012) - 2.1.

> It still doesn't correspond to 2.6.3/2.7.2/2.8.1 anyhow.

It looks like a typo. Of course JSONPath indexes were introduced in 2.1.

Totktonada: I see now, there was PR #150. Thanks!

Totktonada: I'll prepare a patchset of two commits: fix of the checks itself and move two of them to test helpers. I would detach it from this change and open as a separate PR: it has more changes than I would comfortable include here, where I want to fix the problem with merger.

olegrok: No problem. This PR LGTM. Let's push it.

Totktonada: Nice, thanks!

Totktonada: Just changing JSON index support condition to 2.1.2 (and so enabling relevant tests) leads to fail in CI on tarantool 2.3-2.5 (maybe 2.2 as well). [Results](https://github.com/tarantool/crud/actions/runs/1466557064). I'll return here back later.

PR #239, opened 2021-11-24
Support custom sharding func

### LINKED ISSUE #237: Support custom sharding function
Let crud know about `_ddl_sharding_func` space (https://github.com/tarantool/ddl/issues/71) and use provided function to calculate a bucket id by a sharding key.

Part of #27.

AnaNek:
CRUD uses `strcrc32` as sharding function by default. 
The reason why using of `strcrc32` is undesirable is that
this sharding function is not consistent for cdata numbers.
In particular, it returns 3 different values for normal Lua
numbers like 123, for unsigned long long cdata
(like 123ULL, or ffi.cast('unsigned long long',
123)), and for signed long long cdata (like 123LL, or
ffi.cast('long long', 123)).

We should leave `strcrc32` function as sharding function
default to support existing applications using old version of
CRUD.

There is `_ddl_sharding_func` space as part of `ddl` API. 
This space contains sharding functions specified by user.
So user can specify more reliable sharding function
(like `mpcrc32`) using `ddl` API and crud will know about
that and will calculate bucket_id by specified sharding func.

Closes #237


ligurio (changes requested):

crud/common/sharding/sharding_func.lua
```diff
@@ -0,0 +1,88 @@
...
+        end
+
+        return sharding_func()
+    end
+
+    return nil, ShardingFuncError:new(
+            "Wrong sharding function type")
+end
+
+function sharding_func_module.extract_def_from_tuple(tuple)
```
ligurio, line 49: The function name confused me. 'tuple' in name is not a tuple with data, it is a tuple from `box.space._ddl_sharding_func` where we store sharding metadata (sharding keys and functions).
I would rename the function to something like `extract_function_def`.

ligurio (changes requested):

crud/common/sharding/sharding_func.lua
```diff
@@ -0,0 +1,88 @@
...
+            return nil
+        end
+
+        sharding_func = rawget(sharding_func, chunk)
+    end
+
+    return sharding_func
+end
+
+local function as_callable_object(sharding_func_def)
```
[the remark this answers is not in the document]
Totktonada: But we have no strict dependency of crud on ddl.

CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+* CRUD operations calculates bucket id automatically using sharding
+  key specified with DDL schema or in `_ddl_sharding_key` space.
+  NOTE: CRUD methods delete(), get() and update() requires that sharding key
+  must be a part of primary key.
+* CRUD operations calculates bucket id automatically using sharding
```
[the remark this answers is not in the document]
AnaNek: Updated:

```
### Added

* CRUD operations calculates bucket id automatically using sharding
  key specified with DDL schema or in `_ddl_sharding_key` space.
  NOTE: CRUD methods delete(), get() and update() requires that sharding key
  must be a part of primary key.
* Support bucket id calculating using sharding func specified in
  DDL schema or in `_ddl_sharding_func` space.
```

crud/common/sharding/sharding_func.lua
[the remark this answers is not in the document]
AnaNek: Updated.

crud/common/sharding/sharding_func.lua
```diff
@@ -0,0 +1,88 @@
...
+    if type(sharding_func_def) == 'string' then
+        local sharding_func = get_function_from_G(sharding_func_def)
+        if sharding_func == nil then
+            return nil, ShardingFuncError:new(
+                    "No sharding func name (%s) in _G", sharding_func_def)
+        end
+
+        return sharding_func
```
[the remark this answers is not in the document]
AnaNek: Check `is_callable` was added

crud/common/sharding/sharding_func.lua
```diff
@@ -0,0 +1,88 @@
+local errors = require('errors')
+
+local dev_checks = require('crud.common.dev_checks')
+
+local ShardingFuncError = errors.new_class('ShardingFuncError',  {capture_stack = false})
+
+local sharding_func_module = {}
+
+local function get_function_from_G(func_name)
```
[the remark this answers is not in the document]
AnaNek: Now this function is equal with function in ddl

crud/common/sharding/sharding_func.lua
AnaNek: Renamed:

```lua
function sharding_func_module.extract_function_def(tuple)
    if not tuple then
        return nil
    end
...
end
```

crud/common/sharding/sharding.lua
```diff
@@ -4,15 +4,24 @@ local errors = require('errors')
...
 
 local utils = require('crud.common.utils')
-local sharding_key_module = require('crud.common.sharding_key')
+local sharding_metadata_module = require('crud.common.sharding.sharding_metadata')
 
 local sharding = {}
 
-function sharding.key_get_bucket_id(key, specified_bucket_id)
+function sharding.key_get_bucket_id(space_name, key, specified_bucket_id)
     if specified_bucket_id ~= nil then
```
ligurio, line 12: Add checks() here.

ligurio (changes requested):
Thanks for changes!
This patchset contains:
- you own changes with implementation of sharding function support
- copy-pasted code from DDL
- my own code for implementation of secondary sharding keys support (because you are moving module `sharding_key.lua` to another directory)

I don't want to spend time on review my own code and code copy-pasted from DDL. Could you split all patchset for at least three commits:
- moving `sharding_key.lua` to a separate dir, commit marked as a part of issue #239. Git will show file as a renamed and nothing to review here.
- commit with copy-pasted functions from DDL (like `is_called` and other with their tests)
- your own changes with sharding function support

if you will do it you will simplify review a lot

crud/common/sharding/sharding.lua
AnaNek: Done:

```lua
function sharding.key_get_bucket_id(space_name, key, specified_bucket_id)
    dev_checks('string', '?', '?number')

    if specified_bucket_id ~= nil then
        return specified_bucket_id
    end
...
end
```

README.md
```diff
@@ -53,11 +53,63 @@ crud.unflatten_rows(res.rows, res.metadata)
 **Notes:**
 
 * A space should have a format.
-* By default, `bucket_id` is computed as `vshard.router.bucket_id_strcrc32(key)`,
-  where `key` is the primary key value.
-  Custom bucket ID can be specified as `opts.bucket_id` for each operation.
-  For operations that accepts tuple/object bucket ID can be specified as
-  tuple/object field as well as `opts.bucket_id` value.
+
```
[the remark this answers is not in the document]
AnaNek: I have, description in README:

```diff
diff --git a/README.md b/README.md
index c62a541..2b64058 100644
--- a/README.md
+++ b/README.md
@@ -77,6 +77,11 @@ documentation). As soon as sharding key for a certain space is available in
 automatically. Note that CRUD methods `delete()`, `get()` and `update()`
 requires that sharding key must be a part of primary key.
 
+You can specify sharding function to calculate bucket_id with
+sharding func definition as a part of [DDL
+schema](https://github.com/tarantool/ddl#input-data-format)
+or insert manually to the space `_ddl_sharding_func`.
+
 Table below describe what operations supports custom sharding key:
```

ligurio:
- Commit messages must follow to our [guideline](https://www.tarantool.io/en/doc/latest/dev_guide/developer_guidelines/#how-to-write-a-commit-message).
- Commit "Replace common sharding methods in new folder `sharding`" renames "crud/common/sharding.lua" to "crud/common/sharding/sharding.lua". What is a point to do it? Could we choose a better name for module?

ligurio (changes requested):

ligurio:
> I don't want to spend time on review my own code and code copy-pasted from DDL. Could you split all patchset for at least three commits:
> - moving sharding_key.lua to a separate dir, commit marked as a part of issue #239. Git will show file as a renamed and nothing to review here.
> - commit with copy-pasted functions from DDL (like is_called and other with their tests)
your own changes with sharding function support

It is still valid.

README.md
```diff
@@ -77,6 +77,24 @@ documentation). As soon as sharding key for a certain space is available in
...
+this sharding function is not consistent for cdata numbers.
+In particular, it returns 3 different values for normal Lua
+numbers like 123, for unsigned long long cdata
+(like 123ULL, or ffi.cast('unsigned long long',
+123)), and for signed long long cdata (like 123LL, or
+ffi.cast('long long', 123)).
+
+We should leave `strcrc32` function as sharding function
+default to support existing applications using old version of
+CRUD.
```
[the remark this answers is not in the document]
AnaNek: Updated, thanks.

README.md
```diff
@@ -77,6 +77,24 @@ documentation). As soon as sharding key for a certain space is available in
...
+or insert manually to the space `_ddl_sharding_func`.
+
+CRUD uses `strcrc32` as sharding function by default.
+The reason why using of `strcrc32` is undesirable is that
+this sharding function is not consistent for cdata numbers.
+In particular, it returns 3 different values for normal Lua
+numbers like 123, for unsigned long long cdata
+(like 123ULL, or ffi.cast('unsigned long long',
+123)), and for signed long long cdata (like 123LL, or
+ffi.cast('long long', 123)).
```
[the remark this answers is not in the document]
AnaNek: Updated

crud/common/sharding/sharding.lua
AnaNek: Fixed:

```lua
function sharding.key_get_bucket_id(space_name, key, specified_bucket_id)
    dev_checks('string', '?', '?number|cdata')
    ...
end
```

crud/common/sharding/sharding_func.lua
```diff
@@ -93,6 +93,31 @@ function sharding_func_module.extract_function_def(tuple)
...
+                                                          space_name)
+            if err ~= nil then
+                if specified_space_name == space_name then
+                    result_err = err
+                else
+                    log.warn(err)
+                end
+            end
+
+            cache.sharding_func_map[space_name] = sharding_func
```
Totktonada, line 114: Let's consider the case, when the function is defined by its name (strictly speaking by a path in the dot notation) in `_ddl_sharding_func`. If we'll update the function itself (the callable object, e.g. `_G.my_sharding_func = function(key) <...> end`), the old callable will stuck in the cache.

We're solving the same problem in https://github.com/tarantool/ddl/issues/82 (consider 6th bullet).

That's lowers performance of the bucket id calculation in 1.5 times for strcrc32 alike function according to numbers in https://github.com/tarantool/ddl/pull/85. But it is the safe default behaviour.

I suggest to look at the cache strcture in the linked pull request: here we store either parsed dot notation (`{'vshard', 'router', 'mpcrc32'}` for example) or callable (when a function is set as a body).

crud/common/sharding/sharding_key.lua
```diff
@@ -106,12 +109,18 @@ function sharding_key_module.construct_as_index_obj_cache(metadata_map)
                                                                    metadata.space_format,
                                                                    metadata.sharding_key_def)
             if err ~= nil then
-                return err
+                if specified_space_name == space_name then
+                    result_err = err
+                else
+                    log.warn(err)
+                end
```
Totktonada, line 116: We can issue a log entry in the former case as well (`log.warn()` for the latter case and `log.error()` for the former one).

A person who handles an error from the library and a person who looking into production logs are different ones usually.

crud/common/sharding/sharding_metadata.lua
```diff
@@ -38,25 +39,39 @@ local function locked(f)
...
+        end
+    end
+
+    if sharding_func_space ~= nil then
+        for _, tuple in sharding_func_space:pairs() do
+            local space_name = tuple[SPACE_NAME_FIELDNO]
+            local sharding_func_def = sharding_func.extract_function_def(tuple)
+            metadata_map[space_name] = metadata_map[space_name] or {}
+            metadata_map[space_name].sharding_func_def = sharding_func_def
+        end
```
Totktonada, line 74: This is the only usage of the `extract_function_def()` function and I would descrease intermodule coupling and include its code right here (as it is done above for sharding keys).

Also I would use field numbers instead of names for consistency with other code and for potentical ability to reuse this code later for tuples from net.box on 1.10, see https://github.com/tarantool/tarantool/commit/97e8c658f842bd9a623b1673f6ae752082502114.

crud/common/sharding/sharding_func.lua
Totktonada: I would make exception for `vshard.router.bucket_id_strcrc32()`, `vshard.router.bucket_id()` and `vshard.router.bucket_id_mpcrc32()` by default (they are known as stable) and allow users to manage those exceptions.

Since most usages are about built-in vshard bucket id functions, your current solution looks good enough. Because of that and because it is simpler, I retire my comment (but vote for highlighting the limitation in readme).

Totktonada: Filed related issue regarding tarantool/ddl: https://github.com/tarantool/ddl/issues/89.

Totktonada:
Several general points:

1. Copying and removal in another commit: a bit hard to glance on the changes. I would prefer a commit, which doing just moves of functions. I also see that internal API changes are mixed with this functions moving. I propose to split it into different commits.
2. In the second commit calls like `sharding_key_module.extract_from_pk()` (where the module name states that it's about sharding key) becomes `sharding_metadata_module.extract_from_pk()`. Maybe okay as a transient state, but please mention it as expected temporary state in the commit message.
3. Please, expand the scope of the issue regarding sharding keys reload and file another issue. Now we don't track this.
4. `crud/common/sharding/sharding.lua` would look better in `crud/common/sharding/init.lua` or `crud/common/sharding.lua`.

We also discussed several points with Nastya and elaborated the following agreements:

* There is the point from me that it would be better to get rid of space name argument in the cache filling logic, because 1) it simplifies intermodule def-use dependencies; 2) we likely will get rid of this, if we'll implement some kind of background metainfo reloading. However it will require to track errors in the cache and I don't want to work on it within this PR. So we agreed to leave it as is.
* There is the point that we will not fill sharding functions cache at error in constructing cache for sharding keys. Here we also agreed to leave it as is for now to don't stuck on massive rework.

----

I'll look at the following points:

* File/functions moving without real changes.
* Sharding function cache / restore logic (more or less the same as in tarantool/ddl).
* Necessary internal API and structures rework.
* Criteria, when to call fetch and when acquire data from the cache.
* The new fetch logic and API for acquire a sharding function.
* Places, where we change change bucket_id acquiring logic (handling an error at least).
* What is covered by tests.

If I can look at a point more of less independently from others, it is ideal. Some are tightly dependent, but I hope that at least code moving and API reworks will be in different commits.

test/helper.lua
```diff
@@ -323,13 +323,59 @@ function helpers.tarantool_version_at_least(wanted_major, wanted_minor,
...
+function helpers.update_sharding_key_cache(cluster, space_name)
     return cluster.main_server.net_box:eval([[
-        local sharding_key = require('crud.common.sharding_key')
+        local sharding_metadata = require('crud.common.sharding.sharding_metadata')
 
         local space_name = ...
-        return sharding_key.update_cache(space_name)
+        return sharding_metadata.update_sharding_key_cache(space_name)
     ]], {space_name})
 end
```
Totktonada, line 333: `require('crud.common.sharding_key').update_cache()` should work (verbatim). At least one user rely on it currently.

crud/common/sharding/sharding_key.lua
AnaNek: Done.

crud/common/sharding/sharding_metadata.lua
AnaNek: Updated.

crud.lua
```diff
@@ -12,7 +12,7 @@ local select = require('crud.select')
 local truncate = require('crud.truncate')
 local len = require('crud.len')
 local borders = require('crud.borders')
-local sharding_key = require('crud.common.sharding_key')
+local sharding_metadata = require('crud.common.sharding.sharding_metadata')
 local utils = require('crud.common.utils')
```
ligurio, line 17: I have reverted all commits except first one in a patch series and tests failed:

```
sergeyb@pony:~/sources/MRG/crud$ git describe                                                                                           
0.10.0-1-gde58f4e                                                                                                                       
sergeyb@pony:~/sources/MRG/crud$ .rocks/bin/luatest -v test/                                                                            
"tuple.keydef" module is used                                                                                                           
"tuple.merger" module is used                                                                                                           /home/sergeyb/sources/MRG/crud/crud.lua:15: module 'crud.common.sharding_key' not found:No LuaRocks module found for crud.common.shardin
g_key                                                                                                                                   
        no field package.preload['crud.common.sharding_key']                                                                            
        no file '/home/sergeyb/sources/MRG/crud/crud/common/sharding_key.lua'
        no file '/home/sergeyb/sources/MRG/crud/crud/common/sharding_key/init.lua'
```

Probably the same for other commits.

ligurio (changes requested):

test/helper.lua
AnaNek: Updated.

crud.lua
AnaNek: Fixed, thanks, now tests are running, but tests for sharding keys are failing because fetching functions have been moved to sharding metadata module and dependencies between sharding metadata module and sharding key module have be resolved in the next commits.

AnaNek: Fixed, now all tests are passed:

```lua
function sharding_key_module.extract_from_pk(space_name, primary_index_parts, primary_key, timeout)
     dev_checks('string', 'table', '?', '?number')

     local sharding_key_as_index_obj, err = sharding_key_module.fetch_on_router(space_name, timeout)
     local sharding_key_as_index_obj, err = require(
                                            'crud.common.sharding.sharding_metadata'
                                            ).fetch_on_router(space_name, timeout)
     ...
end
```

```lua
local _fetch_on_router = locked(function(timeout)
     ...
     for space_name, metadata in pairs(metadata_map) do
         local sharding_key_as_index_obj, err = require(
                                                'crud.common.sharding.sharding_key'
                                                ).internal.as_index_object(space_name,
                                                                           metadata.space_format,
                                                                           metadata.sharding_key_def)
     ...
     end
     ...
end
```

Totktonada (commented):
While I'm looking on the changes, several minor comments about the 'Separate common sharding metadata methods from sharding key methods' commit.

README.md
```diff
@@ -106,6 +125,13 @@ Current limitations for using custom sharding key:
 - `primary_index_fieldno_map` is not cached, see
   [#243](https://github.com/tarantool/crud/issues/243).
 
+Current limitations for using custom sharding functions:
+
+- It's not possible to update sharding functions automatically when schema is
+  updated on storages, see
+  [#212](https://github.com/tarantool/crud/issues/212). However it is possible
+  to do it manually with `require('crud.common.sharding_func').update_cache()`.
```
Totktonada, line 133: Sadly, we didin't think on it before (my bad).

We should leave `require('crud.common.sharding_key').update_cache()` working and we should not add more hacky intructions to our customers to make things 'just work'. IOW, `require('crud.common.sharding_key').update_cache()` should renew all caches.

I hope we'll fix it soon and it'll be no-op.

Totktonada (commented):
One comment on 'Add support of custom sharding func for crud methods'.

crud.lua
ligurio: For me tests ddl_sharding_func_cache, ddl_sharding_func and ddl_sharding_key broken in commit "Add support of custom sharding func for crud methods".

```
94) ddl_sharding_key.engine:"vinyl".test_update_cache_with_incorrect_key                                                           
...rocks/share/tarantool/cartridge/test-helpers/cluster.lua:177: "localhost:13304": /home/sergeyb/sources/MRG/crud/test/entrypoint/
srv_ddl.lua:175: spaces["customers_G_func"]: redundant key "sharding_func"                                                         
stack traceback:   

93) ddl_sharding_key.engine:"vinyl".test_update_cache                                                                              
...rocks/share/tarantool/cartridge/test-helpers/cluster.lua:177: "localhost:13304": /home/sergeyb/sources/MRG/crud/test/entrypoint/
srv_ddl.lua:175: spaces["customers_G_func"]: redundant key "sharding_func"                                                         
stack traceback:   
```

AnaNek: For this commit CI works fine, I suppose it may happen because of `ddl` version. In latest version of `ddl` suppuration of sharding func was added, but in previous versions `sharding_func` field was not allowed in schema and this had led to an [error](https://github.com/tarantool/ddl/blob/master/ddl/check.lua#L1016). Please, correct me if I am wrong.

crud/common/sharding/sharding_metadata.lua
```diff
@@ -81,21 +82,13 @@ local _fetch_on_router = locked(function(timeout)
...
-        local sharding_key_as_index_obj, err = require(
-                                               'crud.common.sharding.sharding_key'
-                                               ).internal.as_index_object(space_name,
-                                                                          metadata.space_format,
-                                                                          metadata.sharding_key_def)
-        if err ~= nil then
-            return err
-        end
-        cache.sharding_key_as_index_obj_map[space_name] = sharding_key_as_index_obj
+    local err = sharding_key.construct_as_index_obj_cache(metadata_map, space_name)
```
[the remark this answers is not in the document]
AnaNek: I deleted this change from 3rd commit and put it to 4rd commit.

crud/common/sharding/sharding_metadata.lua
```diff
@@ -108,37 +101,38 @@ end)
...
     end
 
-    local timeout = timeout or const.FETCH_SHARDING_KEY_TIMEOUT
-    local err = _fetch_on_router(timeout)
+    local timeout = timeout or const.FETCH_SHARDING_METADATA_TIMEOUT
+    local err = _fetch_on_router(timeout, space_name, metadata_map_name)
     if err ~= nil then
-        if cache.sharding_key_as_index_obj_map ~= nil then
-            return cache.sharding_key_as_index_obj_map[space_name]
-        end
```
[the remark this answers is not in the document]
AnaNek: I deleted this change from 3rd commit and put it to 4rd commit.

crud/common/sharding_key.lua
```diff
@@ -0,0 +1,9 @@
+local sharding_metadata_module = require('crud.common.sharding.sharding_metadata')
+
+local sharding_key_cache = {}
+
+function sharding_key_cache.update_cache(space_name)
+    return sharding_metadata_module.update_sharding_key_cache(space_name)
+end
```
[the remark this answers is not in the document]
AnaNek: I added a comment about this problem. For sharding key:

```diff
+ -- This method is exported here because
+ -- we already have customers using old API
+ -- for updating sharding key cache in their
+ -- projects like `require('crud.common.sharding_key').update_cache()`
function sharding_key_cache.update_cache(space_name)
    return sharding_metadata_module.update_sharding_key_cache(space_name)
end
```

For sharding func:

```diff
+ -- This method is exported here because
+ -- we already have customers using old API
+ -- for updating sharding key cache in their
+ -- projects like `require('crud.common.sharding_key').update_cache()`
+ -- This method provides similar behavior for
+ -- sharding function cache.
function sharding_func_cache.update_cache(space_name)
    return sharding_metadata_module.update_sharding_func_cache(space_name)
end
```

README.md
AnaNek: `require('crud.common.sharding_key').update_cache()` updates both caches, you can see it in `_fetch_on_router` function:

```lua
local _fetch_on_router = locked(function(timeout, space_name, metadata_map_name)
    dev_checks('number', 'string', 'string')

    if cache[metadata_map_name] ~= nil then
        return
    end

    local metadata_map, err = call.any(FETCH_FUNC_NAME, {}, {
        timeout = timeout
    })
    if err ~= nil then
        return err
    end
    if metadata_map == nil then
        cache[cache.SHARDING_KEY_MAP_NAME] = {}
        cache[cache.SHARDING_FUNC_MAP_NAME] = {}
        return
    end

    local err = sharding_key.construct_as_index_obj_cache(metadata_map, space_name)
    if err ~= nil then
        return err
    end

    local err = sharding_func.construct_as_callable_obj_cache(metadata_map, space_name)
    if err ~= nil then
        return err
    end
end)
```

But it returns only sharding key for specified space, I highlighted it in README:

```diff
  to do it manually with `require('crud.common.sharding_key').update_cache()`
+ (this function updates both caches: sharding key cache and sharding function
+ cache, but returned value is sharding key from cache).
```

Similarly for sharding func:

```diff
  to do it manually with `require('crud.common.sharding_func').update_cache()`
+ (this function updates both caches: sharding key cache and sharding function
+ cache, but returned value is sharding function from cache).
```

crud.lua
ligurio: Seems the problem was in ddl version, thanks!

crud/common/sharding_key.lua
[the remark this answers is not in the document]
Totktonada: Thanks!

ligurio (approved):
LGTM

PR #242, opened 2021-11-24
Extract sharding key from conditions

### LINKED ISSUE #213: Fix plan in case of select operation with custom sharding key
Cases:

- pk with fields id and 'name', select with condition by field name with operator `==` triggered map-reduce if there is no a separate index by filed 'name' and name 'name'
- pk = {id1, id3},  sk1={id1, id2}, sk2={id3, id4}. When condition uses sk1 and sk2, then we can recover sharding key too (pair of id1, id3)

rough patch:

```lua
diff --git a/crud/select/plan.lua b/crud/select/plan.lua
index 528ac54..7990fb4 100644
--- a/crud/select/plan.lua
+++ b/crud/select/plan.lua
@@ -49,6 +49,34 @@ local function get_index_for_condition(space_indexes, space_format, condition)
     end
 end
 
+-- Check that if fields that included in sharding key and conditions
+-- have iterator equal to box.index.EQ or box.index.REQ.
+local function extract_sharding_key_from_conditions(conditions, ddl_sharding_key)
+    if ddl_sharding_key == nil then
+        return nil
+    end
+
+    local conditions_map = {}
+    for _, condition in ipairs(conditions) do
+        conditions_map[condition.operand] = {
+            iter = condition.operator,
+            values = condition.values,
+        }
+    end
+
+    local sharding_key_values = {}
+    for _, field_name in ipairs(ddl_sharding_key) do
+        local condition = conditions_map[field_name]
+        if condition ~= nil and condition.iter == '==' then
+            table.insert(sharding_key_values, condition.values)
+        end
+    end
+
+        local inspect = require('inspect')
+        print(inspect.inspect(sharding_key_values))
+    return sharding_key_values
+end
+
 local function extract_sharding_key_from_scan_value(scan_value, scan_index, sharding_index)
     if #scan_value < #sharding_index.parts then
         return nil
@@ -241,6 +269,9 @@ function select_plan.new(space, conditions, opts)
     if scan_value ~= nil and (scan_iter == box.index.EQ or scan_iter == box.index.REQ) then
         sharding_key = extract_sharding_key_from_scan_value(scan_value, scan_index, sharding_index)
     end
+    if sharding_key == nil then
+        sharding_key = extract_sharding_key_from_conditions(conditions, ddl_sharding_key)
+    end
 
     if sharding_key ~= nil and opts.force_map_call ~= true then
         total_tuples_count = 1
```

Part of #166

DifferentialOrange:
PR #181 introduced support of DDL sharding keys. But if sharding key hasn't got a separate index in schema, select with equal conditions for all required sharding key fields still led to map-reduce instead of a single storage call. This patch introduces impoved support of sharding keys extraction and fixes the issue.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,8 +101,6 @@ Current limitations for using custom sharding key:
   updated on storages, see
   [#212](https://github.com/tarantool/crud/issues/212). However it is possible
   to do it manually with `require('crud.sharding_key').update_cache()`.
-- CRUD select may lead map reduce in some cases, see
-  [#213](https://github.com/tarantool/crud/issues/213).
 - No support of JSON path for sharding key, see
   [#219](https://github.com/tarantool/crud/issues/219).
 - `primary_index_fieldno_map` is not cached, see
--- a/crud/select/plan.lua
+++ b/crud/select/plan.lua
@@ -48,41 +48,81 @@ local function get_index_for_condition(space_indexes, space_format, condition)
     end
 end
 
-local function extract_sharding_key_from_scan_value(scan_value, scan_index, sharding_index)
-    if #scan_value < #sharding_index.parts then
+local function extract_sharding_key_from_conditions(conditions, sharding_index, space_indexes, fieldno_map)
+    dev_checks('table', 'table', 'table', 'table')
+
+    -- If name is both valid index name and field name,
+    -- it is interpreted as index name.
+    local filled_fields = {}
+    for _, condition in ipairs(conditions) do
+        if condition.operator ~= compare_conditions.operators.EQ then
+            goto continue
+        end
+
+        local index = space_indexes[condition.operand]
+        if index ~= nil then
+            for i, part in ipairs(index.parts) do
+            -- Consider the following case:
+            --     index_0: {'foo', 'bar'},
+            --     index_1: {'baz', 'bar'},
+            --     conditions: {{'==', 'index_0', {1, 2}}, {'==', 'index_1', {3, nil}}}.
+            -- To check that nil parts will not overwrite already filled_fields,
+            -- we verify that filled_fields[part.fieldno] is empty. If there are
+            -- more than one non-null different value in conditions with equal operator,
+            -- request is already in conflict and it doesn't matter what sharding key we
+            -- will return.
+                if filled_fields[part.fieldno] == nil then
+                    filled_fields[part.fieldno] = condition.values[i]
+                end
+            end
+
+            goto continue
+        end
+
+        local fieldno = fieldno_map[condition.operand]
+        if fieldno == nil then
+            goto continue
+        end
+        filled_fields[fieldno] = condition.values[1]
+
+        ::continue::
+    end
+
+    local sharding_key = {}
+    for i, v in ipairs(sharding_index.parts) do
+        if filled_fields[v.fieldno] == nil then
+            return nil
+        end
+
+        sharding_key[i] = filled_fields[v.fieldno]
+    end
+
+    return sharding_key
+end
+
+local function get_sharding_key_from_scan_value(scan_value, scan_index, scan_iter, sharding_index)
+    dev_checks('?', 'table', 'number', 'table')
+
+    if scan_value == nil then
+        return nil
+    end
+
+    if scan_iter ~= box.index.EQ and scan_iter ~= box.index.REQ then
         return nil
     end
 
     if scan_index.id == sharding_index.id then
+        if type(scan_value) ~= 'table' then
+            return scan_value
+        end
+
+        for i, _ in ipairs(sharding_index.parts) do
+            if scan_value[i] == nil then return nil end
+        end
         return scan_value
     end
 
-    local scan_value_fields_values = {}
-    for i, scan_index_part in ipairs(scan_index.parts) do
-        scan_value_fields_values[scan_index_part.fieldno] = scan_value[i]
-    end
-
-    -- check that sharding key is included in the scan index fields
-    local sharding_key = {}
-    for _, sharding_key_part in ipairs(sharding_index.parts) do
-        local fieldno = sharding_key_part.fieldno
-
-        -- sharding key isn't included in scan key
-        if scan_value_fields_values[fieldno] == nil then
-            return nil
-        end
-
-        local field_value = scan_value_fields_values[fieldno]
-
-        -- sharding key contains nil values
-        if field_value == nil then
-            return nil
-        end
-
-        table.insert(sharding_key, field_value)
-    end
-
-    return sharding_key
+    return nil
 end
 
 -- We need to construct after_tuple by field_names
@@ -90,7 +130,9 @@ end
 -- and these fields are ordered by field_names + primary key + scan key
 -- this order can be differ from order in space format
 -- so we need to cast after_tuple to space format for scrolling tuples on storage
-local function construct_after_tuple_by_fields(space_format, field_names, tuple)
+local function construct_after_tuple_by_fields(fieldno_map, field_names, tuple)
+    dev_checks('table', '?table', '?table|cdata')
+
     if tuple == nil then
         return nil
     end
@@ -99,15 +141,10 @@ local function construct_after_tuple_by_fields(space_format, field_names, tuple)
         return tuple
     end
 
-    local positions = {}
     local transformed_tuple = {}
 
-    for i, field in ipairs(space_format) do
-        positions[field.name] = i
-    end
-
     for i, field_name in ipairs(field_names) do
-        local fieldno = positions[field_name]
+        local fieldno = fieldno_map[field_name]
         if fieldno == nil then
             return nil, FilterFieldsError:new(
                     'Space format doesn\'t contain field named %q', field_name
@@ -145,6 +182,8 @@ function select_plan.new(space, conditions, opts)
     local scan_value
     local scan_condition_num
 
+    local fieldno_map = utils.get_format_fieldno_map(space_format)
+
     -- search index to iterate over
     for i, condition in ipairs(conditions) do
         scan_index = get_index_for_condition(space_indexes, space_format, condition)
@@ -177,7 +216,7 @@ function select_plan.new(space, conditions, opts)
     -- handle opts.first
     local total_tuples_count
     local scan_after_tuple, err = construct_after_tuple_by_fields(
-            space_format, field_names, opts.after_tuple
+        fieldno_map, field_names, opts.after_tuple
     )
     if err ~= nil then
         return nil, err
@@ -230,9 +269,11 @@ function select_plan.new(space, conditions, opts)
     local sharding_index = opts.sharding_key_as_index_obj or primary_index
 
     -- get sharding key value
-    local sharding_key
-    if scan_value ~= nil and (scan_iter == box.index.EQ or scan_iter == box.index.REQ) then
-        sharding_key = extract_sharding_key_from_scan_value(scan_value, scan_index, sharding_index)
+    local sharding_key = get_sharding_key_from_scan_value(scan_value, scan_index, scan_iter, sharding_index)
+
+    if sharding_key == nil then
+        sharding_key = extract_sharding_key_from_conditions(conditions, sharding_index,
+                                                            space_indexes, fieldno_map)
     end
 
     local plan = {
@@ -250,5 +291,10 @@ function select_plan.new(space, conditions, opts)
 
     return plan
 end
+
+select_plan.internal = {
+    get_sharding_key_from_scan_value = get_sharding_key_from_scan_value,
+    extract_sharding_key_from_conditions = extract_sharding_key_from_conditions
+}
 
 return select_plan
--- a/test/integration/ddl_sharding_key_test.lua
+++ b/test/integration/ddl_sharding_key_test.lua
@@ -51,6 +51,8 @@ pgroup.before_each(function(g)
     helpers.truncate_space_on_cluster(g.cluster, 'customers_name_key_non_uniq_index')
     helpers.truncate_space_on_cluster(g.cluster, 'customers_secondary_idx_name_key')
     helpers.truncate_space_on_cluster(g.cluster, 'customers_age_key')
+    helpers.truncate_space_on_cluster(g.cluster, 'customers_name_age_key_different_indexes')
+    helpers.truncate_space_on_cluster(g.cluster, 'customers_name_age_key_three_fields_index')
 end)
 
 pgroup.test_insert_object = function(g)
@@ -279,13 +281,7 @@ pgroup.test_select = function(g)
     t.assert_equals(result.rows[1], tuple)
 end
 
--- TODO: After enabling support of sharding keys that are not equal to primary
--- keys, we should handle it differently: it is not enough to look just on scan
--- value, we should traverse all conditions. Now missed cases lead to
--- map-reduce. Will be resolved in #213.
-pgroup.test_select_wont_lead_map_reduce = function(g)
-    local space_name = 'customers_name_key_uniq_index'
-
+local prepare_data_name_sharding_key = function(g, space_name)
     local conn_s1 = g.cluster:server('s1-master').net_box
     local conn_s2 = g.cluster:server('s2-master').net_box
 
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -53,11 +53,58 @@ crud.unflatten_rows(res.rows, res.metadata)
...
+| `truncate()`                     | No (not required)          |
+| `len()`                          | No (not required)          |
+
+Current limitations for using custom sharding key:
+
+- It's not possible to update sharding keys automatically when schema is
+updated on storages, see [#212](https://github.com/tarantool/crud/issues/212).
+However it is possible to do it manually with
+`require('crud.sharding_key').update_sharding_keys_cache()`.
+- CRUD select may lead map reduce in some cases, see
```
ligurio, line 104: Please remove this.

ligurio (changes requested):

README.md
DifferentialOrange: Fixed

crud/select/plan.lua
```diff
@@ -235,6 +277,17 @@ function select_plan.new(space, conditions, opts)
         sharding_key = extract_sharding_key_from_scan_value(scan_value, scan_index, sharding_index)
     end
 
+    if sharding_key == nil then
+        -- Ignore possible errors to preserve old behavior
+        -- since here it affects only extracting sharding_key from
+        -- conditions and it should not be critical to select call.
+        local ddl_sharding_key = sharding_key_module.fetch_on_router(space_name)
```
Totktonada, line 284: Let's use `opts.sharding_key_as_index_obj`.

BTW, I vote for keeping naming of this object the same. Since PR #181 introduces `sharding_key_as_index_obj`, let's continue.

crud/select/plan.lua
```diff
@@ -48,6 +49,46 @@ local function get_index_for_condition(space_indexes, space_format, condition)
...
+    for _, condition in ipairs(conditions) do
+        if condition.operator ~= compare_conditions.operators.EQ then
+            goto continue
+        end
+
+        local index = space_indexes[condition.operand]
+        if index ~= nil then
+            for i, part in ipairs(index.parts) do
+                filled_fields[part.fieldno] = condition.values[i]
+            end
```
Totktonada, line 69: Please, remind me, whether we support partial keys here?

crud/select/plan.lua
```diff
@@ -48,6 +49,46 @@ local function get_index_for_condition(space_indexes, space_format, condition)
...
+local function extract_sharding_key_from_conditions(conditions, ddl_sharding_key, space_indexes, space_field_positions)
+    if ddl_sharding_key == nil then
+        return nil
+    end
+
+    -- If name is both valid index name and field name,
+    -- it is interpreted as index name.
+    local filled_fields = {}
+    for _, condition in ipairs(conditions) do
+        if condition.operator ~= compare_conditions.operators.EQ then
```
Totktonada, line 61: How about REQ?

crud/select/plan.lua
```diff
@@ -235,6 +277,17 @@ function select_plan.new(space, conditions, opts)
...
 
+    if sharding_key == nil then
+        -- Ignore possible errors to preserve old behavior
+        -- since here it affects only extracting sharding_key from
+        -- conditions and it should not be critical to select call.
+        local ddl_sharding_key = sharding_key_module.fetch_on_router(space_name)
+
+        sharding_key = extract_sharding_key_from_conditions(
+            conditions, ddl_sharding_key, space_indexes, space_field_positions
+        )
```
Totktonada, line 288: Does not it supersedes `extract_sharding_key_from_scan_value()`? I mean, handles all its cases and even more.

crud/select/plan.lua
```diff
@@ -48,6 +49,46 @@ local function get_index_for_condition(space_indexes, space_format, condition)
     end
 end
 
+local function extract_sharding_key_from_conditions(conditions, ddl_sharding_key, space_indexes, space_field_positions)
```
ligurio, line 52: During discussion of implementation of #181 we decided to split sharding key for two terms:
- sharding key - is tuple fields used for bucket id calculation.
- sharding key definition - field names of tuple that used for sharding key. In DDL schema and in `box.space.ddl_sharding_key` user defines sharding key definition.

Seems in `extract_sharding_key_from_conditions()` you are passing sharding key defintion and I would rename second argument to sharding_key_def or something like that.

ligurio (changes requested):

crud/select/plan.lua
Totktonada: It is what you call `sharding_key_as_index_obj` (it is a form of sharding key definition, sure) in PR #181, so I would name it in the same way. See https://github.com/tarantool/crud/pull/242#discussion_r756929428.

crud/select/plan.lua
DifferentialOrange: In terms of `compare_conditions`, there is no such term as `REQ`:
https://github.com/tarantool/crud/blob/fbe4de01f5ae4620a2e5945f24df2fa33307c770/crud/compare/conditions.lua#L10-L16

DifferentialOrange:
I think I will not do any changes until new changes in #181 will stabilize

crud/select/plan.lua
Totktonada: Okay, got it.

crud/select/plan.lua
```diff
@@ -145,6 +181,12 @@ function select_plan.new(space, conditions, opts)
     local scan_value
     local scan_condition_num
 
+    local space_field_positions = {}
+
+    for i, field in ipairs(space_format) do
+        space_field_positions[field.name] = i
+    end
```
Totktonada, line 188: Looks the same as `utils.get_format_fieldno_map()`.

crud/select/plan.lua
DifferentialOrange: Reworked to use with existing values, keep naming in new version.

crud/select/plan.lua
DifferentialOrange: As you know, I had some troubles with partial keys (https://github.com/tarantool/tarantool/issues/6655). I added a test case for partial index with last missing field.

crud/select/plan.lua
DifferentialOrange: Yeah, you were right. New reworked version changed a behavior in plan unit test. I think it is as expected and an improvement rather than drawback, but maybe it should be double-checked. Using `extract_sharding_key_from_conditions` may introduce minor performance drop for a case when sharding value is exactly scan value since all cases treated equally and `scan_value` is not used here.

crud/select/plan.lua
DifferentialOrange: Thanks, used the util

crud/select/plan.lua
DifferentialOrange: Used `sharding_key_as_index_obj`

crud/select/plan.lua
Totktonada: That's two checks: whether the key is not partial and whether scan index is sharding index. We can add it back as fast path.

DifferentialOrange: Added it

crud/select/plan.lua
```diff
@@ -48,49 +48,75 @@ local function get_index_for_condition(space_indexes, space_format, condition)
     end
 end
 
-local function extract_sharding_key_from_scan_value(scan_value, scan_index, sharding_index)
-    if #scan_value < #sharding_index.parts then
+local function extract_sharding_key_from_conditions(conditions, sharding_index, space_indexes, fieldno_map)
```
ligurio, line 51: Add `dev_checks()` in `extract_sharding_key_from_conditions()`.

crud/select/plan.lua
```diff
@@ -48,49 +48,75 @@ local function get_index_for_condition(space_indexes, space_format, condition)
...
         end
 
-        table.insert(sharding_key, field_value)
+        sharding_key[i] = filled_fields[v.fieldno]
     end
 
     return sharding_key
 end
 
+local function get_sharding_key_from_scan_value(scan_value, scan_index, scan_iter, sharding_index)
```
ligurio, line 94: Add `dev_checks()` here.

crud/select/plan.lua
ligurio, line 51: I would add unit tests for `extract_sharding_key_from_conditions()`. I see, integration tests covers function in general, but unit tests are simple to understand and more fast in execution than integration tests. At least case `sharding_index == nil` is not covered in integration tests.

crud/select/plan.lua
```diff
@@ -48,49 +48,75 @@ local function get_index_for_condition(space_indexes, space_format, condition)
...
+    return nil
+end
+
 -- We need to construct after_tuple by field_names
 -- because if `fields` option is specified we have after_tuple with partial fields
 -- and these fields are ordered by field_names + primary key + scan key
 -- this order can be differ from order in space format
 -- so we need to cast after_tuple to space format for scrolling tuples on storage
-local function construct_after_tuple_by_fields(space_format, field_names, tuple)
+local function construct_after_tuple_by_fields(fieldno_map, field_names, tuple)
```
ligurio, line 119: Add `dev_checks()` here.

ligurio, line 119: I would add unit tests for function `construct_after_tuple_by_fields()`. At least, case `with error is not covered.

crud/select/plan.lua
ligurio, line 94: I would add unit tests for function `get_sharding_key_from_scan_value()`. At least, case when `scan_value == nil` is not covered.

test/integration/ddl_sharding_key_test.lua
```diff
@@ -301,12 +297,85 @@ pgroup.test_select_wont_lead_map_reduce = function(g)
...
+        t.assert_not_equals(result, nil)
+        t.assert_equals(#result.rows, 1)
+
+        local stat_b = storage_stat.collect(g.cluster)
+
+        -- Check a number of select() requests made by CRUD on cluster's storages
+        -- after calling select() on a router. Make sure only a single storage has
+        -- a single select() request. Otherwise we lead to map-reduce.
+        local stats = storage_stat.diff(stat_b, stat_a)
+        t.assert_equals(storage_stat.total(stats), 1, 'Select request was not a map reduce')
```
ligurio, line 367: Assert will be wrong if one storage will get two select() requests.

ligurio (changes requested):

crud/select/plan.lua
Totktonada: Let's consider the example:

tuple field names: `{'foo', 'bar', 'baz'}`
index_0: `{'foo', 'bar'}`
index_1: `{'baz', 'bar'}`

conditions: `{{'==', 'index_0', {1, 2}}, {'==', 'index_1', {3, nil}}}`

After processing of the first condition (EQ by index_0), `filled_fields` will be `{1, 2, nil}`. After processing of the second condition (partial EQ by index_1), it will be `{1, nil, 3}`. However it is know that the only tuple we can receive is `{1, 2, 3}`.

Correct?

test/integration/ddl_sharding_key_test.lua
Totktonada: Two selects from one storage is unexpected result too, AFAIS. Are you about the assertion comment (third argument)?

crud/select/plan.lua
Totktonada: Nice.

Totktonada (approved):
Looks generally okay for me.

Please, handle my comment about partial EQ and Sergey's comments regarding testing.

crud/select/plan.lua
DifferentialOrange: Since `plan.new` now uses `fieldno_map` it would be weird to extract it multiple times inside different functions, thus I adapted `construct_after_tuple_by_fields` to use it. But this function does not related in any way to this PR and I don't understand why it should be covered with unit tests as part of this PR.

crud/select/plan.lua
DifferentialOrange: Added

crud/select/plan.lua
DifferentialOrange: Added

crud/select/plan.lua
DifferentialOrange: Added tests, `sharding_key == nil` is not supported anymore

crud/select/plan.lua
DifferentialOrange: Added

crud/select/plan.lua
DifferentialOrange: Added tests, reworked after more thorough inspect

test/integration/ddl_sharding_key_test.lua
DifferentialOrange: I'm also don't get what's the deal

crud/select/plan.lua
DifferentialOrange: Thank you for notice. Fixed it, added several test cases.

Totktonada: Well, the check against `nil` should work. Since it was not obvious neither for me, nor for you, it looks as the check that should have a comment in the code.

DifferentialOrange: Added

Totktonada: Nice.

ligurio (changes requested):

test/unit/select_plan_test.lua
```diff
@@ -339,3 +339,201 @@ g.test_table_count = function()
...
+    local space_indexes = box.space.customers.index
+    local space_format = box.space.customers:format()
+    local fieldno_map = utils.get_format_fieldno_map(space_format)
+
+    -- Sharded by id (primary index).
+    local sharding_index = { parts = {{ fieldno = 1 }} }
+    local conditions = compare_conditions.parse({{ '==', 'id', 2 }})
+    local sharding_key = select_plan._internal.extract_sharding_key_from_conditions(conditions, sharding_index,
+                                                                                    space_indexes, fieldno_map)
+    t.assert_equals(sharding_key, { 2 }, "Extracted sharding key from equal condition for pk index")
```
ligurio, line 353: If this assert() will be broken then all cases below will be blocked by it.
It is better to parameterize testcase like you did it in `test/integration/ddl_sharding_key_test.lua`.

ligurio (changes requested):

crud/select/plan.lua
```diff
@@ -48,49 +50,92 @@ local function get_index_for_condition(space_indexes, space_format, condition)
...
+    dev_checks('?table', '?table', '?table|cdata')
```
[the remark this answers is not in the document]
DifferentialOrange: Fixed

crud/select/plan.lua
```diff
@@ -12,7 +12,9 @@ if has_keydef then
     keydef_lib = compat.require('tuple.keydef', 'key_def')
 end
 
-local select_plan = {}
+local select_plan = {
+    _internal = {},
```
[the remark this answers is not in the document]
DifferentialOrange: Changed

test/unit/select_plan_test.lua
DifferentialOrange: Fair, separated test cases

test/unit/select_plan_test.lua
```diff
@@ -339,3 +339,203 @@ g.test_table_count = function()
...
+        sharding_key = nil,
+    },
+}
+
+for name, case in pairs(get_sharding_key_from_scan_value_cases) do
+    g[('test_%s'):format(name)] = function()
+        local scan_index = box.space.customers.index[case.scan_index]
+        local sharding_index = box.space.customers.index[case.sharding_index]
+
+        local sharding_key = select_plan.internal.get_sharding_key_from_scan_value(case.scan_value, scan_index,
```
ligurio, line 405: Personally for me long names looks awful. I would make a shortcut for it.
```lua
local get_key = select_plan.internal.get_sharding_key_from_scan_value
local sharding_key = get_key(case.scan_value, scan_index,
                             case.scan_iter, sharding_index)
t.assert_equals(sharding_key, case.sharding_key)
```

Feel free to ignore a comment and close.

test/unit/select_plan_test.lua
```diff
@@ -339,3 +339,203 @@ g.test_table_count = function()
...
+        sharding_key = nil,
+    },
+    extract_table_sharding_key_from_partial_conditions = {
+        -- Input values.
+        sharding_index = { parts = {{ fieldno = 1 }, { fieldno = 4 }} },
+        conditions = {{ '==', 'full_name', { nil,  'Doe' } }, { '==', 'id', 1 }},
+        -- Expected values.
+        sharding_key = { 1, 'Doe' },
+    },
+    extract_table_sharding_key_from_partial_conditions_with_nil_and_non_nil_for_same_value_in_different_conditions = {
```
ligurio, line 496: Gosha, length of variable is exceeded of line's length limit:

```
$ echo "extract_table_sharding_key_from_partial_conditions_with_nil_and_non_nil_for_same_value_in_different_conditions" | wc -c
111
```


https://www.tarantool.io/en/doc/latest/dev_guide/c_style_guide/
> The limit on the length of lines is 80 columns and this is a strongly preferred limit. As for comments, the same limit of 80 columns is applied.

test/unit/select_plan_test.lua
```diff
@@ -339,3 +339,203 @@ g.test_table_count = function()
...
+}
+
+for name, case in pairs(extract_sharding_key_from_conditions_cases) do
+    g[('test_%s'):format(name)] = function()
+        local space_indexes = box.space.customers.index
+        local space_format = box.space.customers:format()
+        local fieldno_map = utils.get_format_fieldno_map(space_format)
+        local conditions = compare_conditions.parse(case.conditions)
+
+        local sharding_key = select_plan.internal.extract_sharding_key_from_conditions(conditions, case.sharding_index,
```
ligurio, line 512: May be to create a shortcut for a function?

test/unit/select_plan_test.lua
```diff
@@ -339,3 +339,203 @@ g.test_table_count = function()
...
+    local space_indexes = box.space.customers.index
+    local space_format = box.space.customers:format()
+    local fieldno_map = utils.get_format_fieldno_map(space_format)
+
+    local sharding_index = { parts = {{ fieldno = 6 }} }
+
+    local conditions = compare_conditions.parse({{ '==', 'city', { 1, 'New York' } }})
+    local sharding_key = select_plan.internal.extract_sharding_key_from_conditions(conditions, sharding_index,
+                                                                                   space_indexes, fieldno_map)
+    t.assert_equals(sharding_key, { 'New York' }, "Extracted sharding key from index in case of name collision")
```
ligurio, line 540: Does it look better or not?

```lua
--- a/test/unit/select_plan_test.lua
+++ b/test/unit/select_plan_test.lua
@@ -535,7 +535,9 @@ g.test_extract_sharding_key_from_conditions_for_index_and_field_with_same_name =
     local sharding_index = { parts = {{ fieldno = 6 }} }
 
     local conditions = compare_conditions.parse({{ '==', 'city', { 1, 'New York' } }})
-    local sharding_key = select_plan.internal.extract_sharding_key_from_conditions(conditions, sharding_index,
-                                                                                   space_indexes, fieldno_map)
-    t.assert_equals(sharding_key, { 'New York' }, "Extracted sharding key from index in case of name collision")
+    local extract_sharding_key = select_plan.internal.extract_sharding_key_from_conditions
+    local sharding_key = extract_sharding_key(conditions, sharding_index,
+                                              space_indexes, fieldno_map)
+    t.assert_equals(sharding_key, { 'New York' },
+        "Extracted sharding key from index in case of name collision")
 end
```

crud/select/plan.lua
```diff
@@ -230,9 +267,11 @@ function select_plan.new(space, conditions, opts)
...
 
     -- get sharding key value
-    local sharding_key
-    if scan_value ~= nil and (scan_iter == box.index.EQ or scan_iter == box.index.REQ) then
-        sharding_key = extract_sharding_key_from_scan_value(scan_value, scan_index, sharding_index)
+    local sharding_key = get_sharding_key_from_scan_value(scan_value, scan_index, scan_iter, sharding_index)
+
+    if sharding_key == nil then
+        sharding_key = extract_sharding_key_from_conditions(conditions, sharding_index,
+                                                                                  space_indexes, fieldno_map)
```
ligurio, line 274: broken indentation

ligurio (changes requested):

test/unit/select_plan_test.lua
DifferentialOrange: Well, the config in `.luacheckrc` declares that `max_line_length = 120`. Moreover, the rest of `crud` code do not satisfy it: 12 lines in `sharding_key.lua`, 18 kines in `plan.lua` before this PR, etc.

test/unit/select_plan_test.lua
DifferentialOrange: Used a shortcut

test/unit/select_plan_test.lua
DifferentialOrange: Shortened it

DifferentialOrange: And why are you referring C style guide for Lua code?

test/unit/select_plan_test.lua
DifferentialOrange: Used a shortcut

test/unit/select_plan_test.lua
DifferentialOrange: Used a shortcut and extra line break

crud/select/plan.lua
DifferentialOrange: Fixed

test/unit/select_plan_test.lua
ligurio: Thanks!

ligurio (approved):
LGTM

PR #244, opened 2021-11-29
Add statistics for CRUD operations on router

DifferentialOrange:
Add statistics module for collecting metrics of CRUD operations on
router. Wrap all CRUD operation calls in the statistics collector.
Statistics must be enabled manually with `crud.cfg`. They can be
disabled, restarted or re-enabled later.

`crud.stats()` returns
```yml
---
- spaces:
    my_space:
      insert:
        ok:
          latency: 0.002
          count: 19800
          time: 39.6
        error:
          latency: 0.000001
          count: 4
          time: 0.000004
      select:
        ok:
          latency: 0.032
          count: 43100
          time: 1379.2
        error:
          latency: 0.000001
          count: 2
          time: 0.000002
        details:
          map_reduces: 48
          tuples_fetched: 105000
          tuples_lookup: 2380000
---
```
`spaces` section contains statistics for each observed space.
If operation has never been called for a space, the corresponding
field will be empty. If no requests has been called for a
space, it will not be represented. Space data is based on
client requests rather than storages schema, so requests
for non-existing spaces are also collected.

This patch introduces `crud.cfg`. `crud.cfg` is a tool to set module
configuration. It is similar to Tarantool `box.cfg`, although we don't
need to call it to bootstrap the module -- it is used only to change
configuration. `crud.cfg` is a callable table. To change configuration,
call it: `crud.cfg{ stats = true }`. You can check table contents as
with ordinary table, but do not change them directly -- use call
instead. Table contents is immutable and use proxy approach
(see [1, 2]). Iterating through `crud.cfg` with pairs is not supported
yet, refer to tarantool/crud#265.

Possible statistics operation labels are
`insert` (for `insert` and `insert_object` calls),
`get`, `replace`  (for `replace` and `replace_object` calls), `update`,
`upsert` (for `upsert` and `upsert_object` calls), `delete`,
`select` (for `select` and `pairs` calls), `truncate`, `len`, `count`
and `borders` (for `min` and `max` calls).

Each operation section consists of different collectors
for success calls and error (both error throw and `nil, err`)
returns. `count` is total requests count since instance start
or stats restart. `latency` is average time of requests execution,
`time` is the total time of requests execution.

Since `pairs` request behavior differs from any other crud request, its
statistics collection also has specific behavior. Statistics (`select`
section) are updated after `pairs` cycle is finished: you
either have iterated through all records or an error was thrown.
If your pairs cycle was interrupted with `break`, statistics will
be collected when pairs objects are cleaned up with Lua garbage
[...]

PR #251, opened 2021-12-15
Add performance tests to compare crud and vshard

### LINKED ISSUE #225: Create a test with comparison vshard.call() and CRUD performance

DifferentialOrange:
This patch adds new cases for performance tests: select for equal
conditions for primary and secondary indexes (including known bucket_id
case from #220), and adds corresponding vshard test cases to compare
performance. Comparison may be not exactly precise since vshard test
functions use naive mergers, but should at least estimate basic
differences.

Test run on HP ProBook 440 G7 i7/16Gb/256SSD shows that CRUD module is
3-6 times slower than vshard calls for prepared functions for select and
1.6 times slower for insert.

#### SUCCESS REQUESTS PER SECOND
(The higher the better)

|                                   |       vshard | crud (stats disabled) | crud (stats disabled, known bucket_id) |
| --------------------------------- | ------------ | --------------------- | -------------------------------------- |
|                      select by pk |     81824.72 |              23936.52 |                               23992.12 |
|        select gt by pk (limit 10) |     15201.35 |               4305.91 |                                        |
| select eq by secondary (limit 10) |     14783.89 |               4835.79 |                                        |
|   select eq by sharding secondary |     79791.30 |              13689.91 |                               18376.29 |
|                            insert |     84113.60 |              51847.18 |                                        |

#### AVERAGE CALL TIME
(The lower the better)

|                                   |       vshard | crud (stats disabled) | crud (stats disabled, known bucket_id) |
| --------------------------------- | ------------ | --------------------- | -------------------------------------- |
|                      select by pk |     2.444 ms |              8.355 ms |                               8.335 ms |
|        select gt by pk (limit 10) |    13.156 ms |             46.446 ms |                                        |
| select eq by secondary (limit 10) |    13.528 ms |             41.357 ms |                                        |
|   select eq by sharding secondary |     2.506 ms |             14.604 ms |                              10.881 ms |
|                            insert |     7.132 ms |             11.571 ms |                                        |

#### MAX CALL TIME
(The lower the better)

|                                   |       vshard | crud (stats disabled) | crud (stats disabled, known bucket_id) |
| --------------------------------- | ------------ | --------------------- | -------------------------------------- |
|                      select by pk |    17.277 ms |             37.793 ms |                              36.457 ms |
|        select gt by pk (limit 10) |    41.015 ms |            103.195 ms |                                        |
| select eq by secondary (limit 10) |    43.169 ms |             98.459 ms |                                        |
|   select eq by sharding secondary |    15.934 ms |             71.610 ms |                              58.771 ms |
|                            insert |    66.039 ms |             72.328 ms |                                        |

### CI Runner

See https://github.com/tarantool/crud/runs/5421561722?check_suite_focus=true

Several minor fixes also was introduced.

Fix replicaset alias in test cluster. Disable checks for perf tests.
Return tuples from pairs perf case.

Fix running separate tests. After this patch, developer can run luatest
with -p flag for perf tests to get results only for requested tests.
Example:
```bash
PERF_MODE_ON=true ./.rocks/bin/luatest -c -p perf.test_crud_pairs_gt_with_stats_disabled
```

- [x] Tests
- Changelog (I don't think it should bother users)
- Documentation (There's nothing to document)

Closes #225

### DIFF
--- a/test/performance/perf_test.lua
+++ b/test/performance/perf_test.lua
@@ -9,6 +9,7 @@ local t = require('luatest')
 local g = t.group('perf')
 
 local helpers = require('test.helper')
+helpers.disable_dev_checks()
 
 
 local id = 0
@@ -55,7 +56,7 @@ g.before_all(function(g)
             },
             {
                 uuid = helpers.uuid('d'),
-                alias = 's-2',
+                alias = 's-3',
                 roles = { 'customers-storage', 'crud-storage' },
                 servers = {
                     { instance_uuid = helpers.uuid('d', 1), alias = 's3-master' },
@@ -132,7 +133,10 @@ local function visualize_section(total_report, name, comment, section, params)
         local row_str = '| ' .. normalize(row, params.row_header_width) .. ' |'
 
         for _, column in ipairs(params.columns) do
-            local report = total_report[row][column]
+            local report = nil
+            if total_report[row] ~= nil then
+                report = total_report[row][column]
+            end
 
             local report_str
             if report ~= nil then
@@ -201,6 +205,31 @@ local function visualize_report(report)
         'The lower the better', 'average_time', params)
     report_str = report_str .. visualize_section(report, 'MAX CALL TIME',
         'The lower the better', 'max_time', params)
+
+    -- Expect every row to have the same run parameters.
+    local comment_str = ''
+    for _, row in ipairs(params.rows) do
+        local row_report = report[row]
+        if row_report == nil then
+            goto continue
+        end
+
+        local row_report_sample = select(2, next(row_report))
+        if row_report_sample == nil then
+            goto continue
+        end
+
+        local row_comment = ('%q was planned for %d seconds with %d connections and %d fibers total.\n'):format(
+            row,
+            row_report_sample.params.timeout,
+            row_report_sample.params.connection_count,
+            row_report_sample.params.fiber_count)
+        comment_str = comment_str .. row_comment
+
+        ::continue::
+    end
+
+    report_str = report_str .. comment_str
 
     log.info(report_str)
 end
@@ -393,6 +422,8 @@ local cases = {
                     for _, tuple in require('crud').pairs(...) do
                         table.insert(t, tuple)
                     end
+
+                    return t
                 end
             ]])
             select_prepare(g)
@@ -413,6 +444,8 @@ local cases = {
                     for _, tuple in require('crud.select').pairs(...) do
                         table.insert(t, tuple)
                     end
+
+                    return t
                 end
             ]])
             select_prepare(g)
@@ -512,6 +545,8 @@ for name, case in pairs(cases) do
                 success_rps = ('%.2f'):format(report.count / run_time),
                 max_time = ('%.3f ms'):format(report.max_time * 1e3),
             }
+
+            report.params = params
 
             local total_count = report.count + #report.errors
             if total_count > 0 then
--- a/test/performance/perf_test.lua
+++ b/test/performance/perf_test.lua
@@ -102,15 +102,19 @@ local row_name = {
     insert = 'insert',
     select_pk = 'select by pk',
     select_gt_pk = 'select gt by pk (limit 10)',
+    select_secondary_eq = 'select eq by secondary (limit 10)',
+    select_secondary_sharded = 'select eq by sharding secondary',
     pairs_gt = 'pairs gt by pk (limit 100)',
 }
 
 local column_name = {
-    without_stats_wrapper = 'without stats wrapper',
-    stats_disabled = 'stats disabled',
-    local_stats = 'local stats',
-    metrics_stats = 'metrics stats (no quantiles)',
-    metrics_quantile_stats = 'metrics stats (with quantiles)',
+    vshard = 'vshard',
+    without_stats_wrapper = 'crud (without stats wrapper)',
+    stats_disabled = 'crud (stats disabled)',
+    bucket_id = 'crud (stats disabled, known bucket_id)',
+    local_stats = 'crud (local stats)',
+    metrics_stats = 'crud (metrics stats, no quantiles)',
+    metrics_quantile_stats = 'crud (metrics stats, with quantiles)',
 }
 
 local function visualize_section(total_report, name, comment, section, params)
@@ -142,7 +146,7 @@ local function visualize_section(total_report, name, comment, section, params)
             if report ~= nil then
                 report_str = report.str[section]
             else
-                report_str = 'unknown'
+                report_str = ''
             end
 
             row_str = row_str .. ' ' .. normalize(report_str, params.col_width[column]) .. ' |'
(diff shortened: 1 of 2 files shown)


DifferentialOrange:
PR is ready for review, but since it depends on #244 improvements, I will not request any reviews for now (until #244 is merged).

test/performance/perf_test.lua
```diff
@@ -101,15 +101,19 @@ local row_name = {
     insert = 'insert',
```
Totktonada, line 101: Regarding line 58: `alias = 's-2'` is used twice.

Totktonada, line 101: It seems, `dev_checks()` is not disabled for performance tests.

test/performance/perf_test.lua
```diff
@@ -385,6 +685,115 @@ local cases = {
...
+        params = vshard_select_params_sharded_by_secondary,
+        matrix = { [''] = { column_name = column_name.vshard } },
+        integration_params = integration_params,
+        perf_params = select_perf,
+        row_name = row_name.select_secondary_sharded,
+    },
+
     crud_pairs_gt = {
         prepare = function(g)
             g.router:eval([[
```
Totktonada, line 799: `crud_pairs_*()` do not return any result.

test/performance/perf_test.lua
Totktonada, line 101: There is no way to run particular test and look at partial report:

```shell
$ PERF_MODE_ON= luatest -v perf.test_crud_pairs_gt_with_stats_disabled
```

test/performance/perf_test.lua
```diff
@@ -234,6 +251,200 @@ local select_prepare = function(g)
...
+                for _, tuples in pairs(storage_response) do
+                    if tuples[i] ~= nil then
+                        response[ind] = tuples[i]
+                    end
+                end
+
+                if ind == opts.limit then
+                    break
+                end
+            end
```
Totktonada, line 374: `idx` is always zero? The naive merger does not sort the tuples?

Totktonada: Maybe it worth to add reference results for test cases and check the results on a warm up (several first request, where we don't count timings).

test/performance/perf_test.lua
```diff
@@ -314,6 +568,16 @@ local pairs_perf = {
 }
```
Totktonada, line 568: It is not obvious from the final report that the testing is performed using different parameters (fiber/connection count) for different rows. I would show the parameters somewhere.

test/performance/perf_test.lua
```diff
@@ -234,6 +251,200 @@ local select_prepare = function(g)
...
+            rawset(_G, '_vshard_select_storage', _vshard_select_storage)
+
+
+            local function _vshard_select_customer_by_name_and_age_storage(key)
+                local space = box.space.customers_name_age_key_different_indexes
+                local index = space.index.age
+
+                for _, tuple in index:pairs(key[2]) do
+                    if tuple.name == key[1] then
+                        return tuple
```
Totktonada, line 436: Since it is named as 'select', I gues it would be more intuitive to return a table of tuples even when there is only one.

test/performance/perf_test.lua
```diff
@@ -234,6 +251,200 @@ local select_prepare = function(g)
...
+    ]])
+
+    for _, server in ipairs(g.cluster.servers) do
+        server.net_box:eval([[
+            local function _vshard_insert_storage(space_name, tuple, bucket_id)
+                local space = box.space[space_name]
+                assert(space ~= nil)
+
+                assert(space.index.bucket_id ~= nil)
+                tuple[space.index.bucket_id.parts[1].fieldno] = bucket_id
```
Totktonada, line 402: crud sets bucket id in a tuple on router. Maybe the comparison will be a bit more fair, if we'll do this on router for the vshard case too.

Totktonada (approved):
LGTM.

Quite interesting results, thanks!

I left some minor comments. Feel free to fix or go ahead.

test/performance/perf_test.lua
DifferentialOrange: > `idx` is always zero?

I don't get the question. I take top records from each instance response until I reach expected count.

> The naive merger does not sort the tuples?

Yeah. Honestly, I have no idea how to work with merger and it seemed to me that introducing complicated tuples sorting here is not that easy and will require many additional lines of code.

Totktonada: > I don't get the question. I take top records from each instance response until I reach expected count.

The response will always contain one tuple (in `response[0]`). It does not look as expected result.

> Yeah. Honestly, I have no idea how to work with merger and it seemed to me that introducing complicated tuples sorting here is not that easy and will require many additional lines of code.

The `table.sort()` accepts the comparator function. It is easy to implement, when you know a field number.

test/performance/perf_test.lua
DifferentialOrange: Fixed in separate commit

DifferentialOrange: Well, it is really crucial and I had completely forgot about it. Thank you for your comment. Now results seems more optimistic

DifferentialOrange: Fixed. Also it should be noted that `-c` is required since `-v` do not captures printout.

test/performance/perf_test.lua
DifferentialOrange: Added

test/performance/perf_test.lua
DifferentialOrange: Fixed

test/performance/perf_test.lua
DifferentialOrange: Changed

test/performance/perf_test.lua
DifferentialOrange: > The response will always contain one tuple (in `response[0]`). It does not look as expected result.

Oh, now I get it. My bad.

DifferentialOrange: > The `table.sort()` accepts the comparator function. It is easy to implement, when you know a field number.

Ok, I have thought that something more complicated is required.

DifferentialOrange: > `idx` is always zero? The naive merger does not sort the tuples?

Resolved

DifferentialOrange: > Maybe it worth to add reference results for test cases and check the results on a warm up (several first request, where we don't count timings).

It sounds like a relevant comment, but also as one that requires refactoring and additional work. I think we skip this one for now

test/performance/perf_test.lua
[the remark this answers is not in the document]
DifferentialOrange: Yeah, current structure is a mess, but I think it still better that the one before and we should refactor it if someone will need it for their work.

Totktonada: That's nice: should be quite useful for debugging performance problems.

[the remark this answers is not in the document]
Totktonada: Sure.

Issue #234, opened 2021-11-11
Optimize crud in the single replicaset select case


Totktonada:
Say, when bucket_id is already known. I was equipped by the test case from #220 and saw significant difference between pure vshard call and crud's select even after PR #222 and PR #226. It was around 2 times, but I have no precise numbers under hands now. The bottleneck is on router.

Brief `jit.p` profiling does not show low hanging fruits anymore. Selective enabling / disabling of crud's code shows that, say, considerable amount of time is spent on creating select plan. Surprisingly, [pcall()](https://github.com/tarantool/crud/blob/5e8c91e39bd608804e0ab85989a537dfd63ec7bf/crud/select/compat/select.lua#L288-L290) also leads to a visible slowdown. Checks was in the `jit.p` profile (around 5%, but it is by the profile, not by RPS).

Anyway, my observations were quite brief: let's look into this more systematically. Carefully review involved code and evaluate time spent in different parts of the code. Next, look what we can do to speed up our case.

The goal is to shrink crud / vshard RPS ratio to ~1.5.

Let's start from #225.

DifferentialOrange:
This ticket is became something like "Select optimization" epic due to stuff like "The goal is to shrink crud / vshard RPS ratio to ~1.5". We need to reestimate it if we decide to continue working on it.

PR #252, opened 2021-12-17
Optimize select with known bucket_id

DifferentialOrange:
Merge after #251.

After this patch, select and pairs requests will no longer fetch
sharding key info and extract sharding key info if bucket_id specified.
Since calls with specified bucket_id already ignore sharding key
values, behavior will not change. Other crud operations already have
this optimization.

Based on test runs on HP ProBook 440 G7 i7/16Gb, performance had
increased by 6-7%.

I didn't forget about

- Tests (see #251)
- [x] Changelog
- Documentation (don't think there is something to describe)

Part of #234

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Integrate CRUD statistics with [`metrics`](https://github.com/tarantool/metrics) (#224).
 
 ### Changed
+* Optimize select with known bucket_id (#234).
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


DifferentialOrange:
PR is ready for review, but since it depends on #244 and #251 improvements, I will not request any reviews for now (until #244 and #251 is merged).

Totktonada:
The implemented optimization looks okay, but I had idea to find more headrooms in the scope of the issue. Are you plan to do so?

DifferentialOrange:
> The implemented optimization looks okay, but I had idea to find more headrooms in the scope of the issue. Are you plan to do so?

I thought the plan was only to solve this one and investigate code later as part of some another ticket/epic.

Totktonada (approved):
LGTM.

Issue #212, opened 2021-09-13
Make automated invalidation of caches on router on schema reload or ddl sharding keys update


ligurio:
Imagine user have a cluster with running CRUD module. He wants to update a schema or update sharding keys  in `_ddl_sharding_key` space. In such case we should update cache with sharding keys on router.
Now we have a mechanism to follow schema changes on storage (see option `add_space_schema_hash` in CRUD operations and function `schema.get_space_schema_hash()`).

Patch for this can look like this:

```lua
diff --git a/crud/borders.lua b/crud/borders.lua
index ac77cef..b702d09 100644
--- a/crud/borders.lua
+++ b/crud/borders.lua
@@ -108,9 +108,6 @@ local function call_get_border_on_router(border_name, space_name, index_name, op
         local storage_result = storage_result[1]
         if storage_result.err ~= nil then
             local need_reload = schema.result_needs_reload(space, storage_result)
-            if need_reload == true then
-                sharding.schema_reload_actions(space_name, storage_result.ddl_sharding_key)
-            end
             return nil, BorderError:new("Failed to get %s: %s", border_name, storage_result.err), need_reload
         end
 
diff --git a/crud/common/schema.lua b/crud/common/schema.lua
index 93afc01..bbfae66 100644
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -214,7 +214,6 @@ function schema.wrap_func_result(space, func, args, opts)
     else
         result.res = filter_tuple_fields(func_res, opts.field_names)
     end
-    result.ddl_sharding_key = schema.fetch_ddl_sharding_key(box, space.name)
 
     return result
 end
diff --git a/crud/insert.lua b/crud/insert.lua
index 3b9fe1f..9d4332c 100644
--- a/crud/insert.lua
+++ b/crud/insert.lua
@@ -84,9 +84,6 @@ local function call_insert_on_router(space_name, tuple, opts)
 
     if storage_result.err ~= nil then
         local need_reload = schema.result_needs_reload(space, storage_result)
-        if need_reload == true then
-            sharding.schema_reload_actions(space_name, storage_result.ddl_sharding_key)
-        end
         return nil, InsertError:new("Failed to insert: %s", storage_result.err), need_reload
     end
 
diff --git a/crud/replace.lua b/crud/replace.lua
index f52f331..26d5721 100644
--- a/crud/replace.lua
+++ b/crud/replace.lua
[...]

Issue #166, opened 2021-06-03
Support calculating bucket_id based on ddl.sharding_key


sharonovd:
Crud/ddl/migrations is a way to go for our customers. But project after a project I observe the following clutches:

```lua
local function get_object_by_pk(part1, part2)    
    local bucket_id = vshard.router.bucket_id_mpcrc32(part2)    
    local settings, err_settings = crud.get('object', {part1, part2}, {bucket_id = bucket_id})   
    if err_settings ~= nil then return nil, err_settings    end    
    return settings
end

local function upsert_object(object)    
    local id = settings["id"]    
    local bucket_id = vshard.router.bucket_id_mpcrc32(id)
    local res, err = crud.replace_object("settings_object", settings, {bucket_id = bucket_id})    
    if err ~= nil then        return nil, err    end    
end
```

But this info is already present in ddl! (`ddl.register_sharding_key('object', {"id"})`)
This shoould work out-of-the-box

artur-barsegyan:
## TODO:
- [ ]  Поддержка `sharding_key` из CRUD (описано выше)
- [ ]  Поддержка кастомных шард функций в DDL 
- [ ]  DDL API
    - [ ]  `ddl.bucket_id()` 
    - [ ]  `ddl.*.set_sharding_opts`
    
Подробнее о каждом из пунктом описано [в Notion.](https://www.notion.so/tarantool/DDL-1811eb0c2efa4168b7aa79a93c519450)

artur-barsegyan:
As part of the pull request https://github.com/tarantool/crud/pull/181, I saw that there was a commit with a change to the default hash function.

I wrote an issue with details about that problem: #185. We should implement it in the current scope of work. Without that we can't deliver:
- CRUD + DDL feature for our paid customers
- The new Getting started with CRUD on-board

ligurio:
@Totktonada, generally, you are right. Additionally `select()` was required sharding key support too.

To sum up

| CRUD method      | Added sharding key support |
| ---------------- | ----------- |
| `get()`                       | Yes |
| `insert()` / `insert_object()` | Yes |
| `delete()`        | Yes |
| `replace()` / `replace_object()` | Yes |
| `upsert()` / `upsert_object()` | Yes |
| `select()` / `pairs()` | Yes |
| `update()`       | Yes |
| `upsert()` / `upsert_object()` | Yes |
| `replace()` / `replace_object()` | Yes | 
| `min()` / `max()` | **No** (not required) |
[...]

PR #256, opened 2022-01-27
readme: fix update_cache() recipe

Totktonada:
When a sharding key is updated on storages a user should call this
function on routers to re-fetch the new sharding keys. I would highlight
that even if we add a space that was never seen before, we should call
the function on routers. Otherwise crud will assume that the new space
has the sharding key equal to the primary key and will calculate
`bucket_id` incorrectly.

Follows up #166
Related to #212
Related to TNT-462


Totktonada:
That's trivial, but important documentation fix, so I'll proceed.

PR #260, opened 2022-02-09
Add check for incorrect call mode

### LINKED ISSUE #259: Add check for incorrect call mode
In README highlights that `mode` parameter for CRUD
methods should be string with `read` or `write`
value but if user mistypes and sets invalid `mode`
value like `wriite` `call` returns no error
and `call` falls through `callrw`.

AnaNek:
In README highlights that `mode` parameter for CRUD
methods should be string with `read` or `write`
value but if user mistypes and sets invalid `mode`
value like `wriite` `call` returns no error
and `call` method is called with read mode.

Closes #259

### DIFF
--- a/README.md
+++ b/README.md
@@ -586,6 +586,21 @@ crud.len('customers')
 ...
 ```
 
+### Call options for crud methods
+
+Combinations of `mode`, `prefer_replica` and `balance` options lead to:
+
+* `mode` == `write` - method performed on master with vshard call `callrw`
+* `mode` == `read`
+  * not prefer_replica, not balance -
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -586,6 +586,19 @@ crud.len('customers')
 ...
 ```
 
+Combinations of `mode`, `prefer_replica` and `balance` options lead to:
```
Totktonada, line 589: It is in the 'Len' section.

Totktonada (approved):
LGTM.

README.md
AnaNek: I added a title for this section:

```diff
+ ### Call options for crud methods

  Combinations of `mode`, `prefer_replica` and `balance` options lead to:
```

PR #262, opened 2022-02-09
Refactor calls strategies test with parameterization

### LINKED ISSUE #261: Refactor calls strategies test with parameterization
Without parameterization we check CRUD methods
in calls strategies test with different values
of `prefer_replica`, `balance` and `mode`,
a broken testcase can block test cases that
follows it. With parameterization it will not happen.

AnaNek:
Without parameterization we check CRUD methods
in calls strategies test with different values
of `prefer_replica`, `balance` and `mode`,
a broken testcase can block test cases that
follows it. With parameterization it will not happen.

Closes #261

### DIFF
--- a/test/integration/read_calls_strategies_test.lua
+++ b/test/integration/read_calls_strategies_test.lua
@@ -1,11 +1,34 @@
 local fio = require('fio')
 
 local t = require('luatest')
-local g = t.group('read_calls_strategies')
+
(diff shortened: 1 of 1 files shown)


ligurio (approved):
LGTM

Totktonada (approved):
LGTM.

Issue #253, opened 2021-12-23
design document: how database schema information is renewed


Totktonada:
It is not easy to deduce design decisions and reasons from the code, so I would like to see a document that would answer the following questions:

* When and how the schema is reloaded.
* Where it leans on built-in net.box schema reload and where it doesn't.
* Which alternatives were considered and why they were declined?
* The same questions about sharding key and sharding function information reload.

----

See the following discussion: https://github.com/tarantool/crud/pull/111.

See `==` conditions in `schema.batching_result_needs_reload()` and `sharding.batching_result_needs_sharding_reload()` from PR #232.

----

NB: net.box space objects can update in background, correct? So we can flatten an object using an old schema, receive the new space hash from the storage and compare it with *the new* schema on the router (because it was updated).

PR #268, opened 2022-03-24
Differential orange/gh 212 reload sharding keys

DifferentialOrange:

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Closes #???


DifferentialOrange:
Current version performance drops are 13%

DifferentialOrange:
Current solution blocked by https://github.com/tarantool/crud/issues/271

DifferentialOrange:
TODO: cleanup https://github.com/tarantool/crud/tree/DifferentialOrange/gh-212-reload-sharding-keys-v2

crud/insert.lua
```diff
@@ -58,6 +58,8 @@ local function call_insert_on_router(space_name, tuple, opts)
         return nil, InsertError:new("Space %q doesn't exist", space_name), true
     end
 
+    local tuple = table.deepcopy(original_tuple)
```
Totktonada, line 73: Raw thoughts on possible optimizations:

- Skip copying on `crud.insert_object()`.
- Pass the bucket_id separately to storage and insert it into the tuple on the storage (take care to router/storage rolling code update).

I don't want to change it within this pull request, but rather to think around later.

test/entrypoint/srv_ddl_reload.lua
```diff
@@ -0,0 +1,198 @@
...
+                    return
+                end
+
+                local new_schema = table.deepcopy(schema)
+                new_schema.spaces['customers_new'] = table.deepcopy(customers_schema)
+                new_schema.spaces['customers_new'].sharding_func = {
+                    body = [[
+                        function(key)
+                            local vshard = require('vshard')
+                            return vshard.router.bucket_id_mpcrc32(key) % 3000 + 1
```
Totktonada, line 187: Nit: It is a bit strange to apply `x % 3000 + 1` on `bucket_id`, which is already a hash by modulo of `bucket_count` plus one, where `bucket_count` is set below in `cartridge.cfg()` and is 3000.

test/integration/ddl_sharding_info_reload_test.lua
```diff
@@ -0,0 +1,118 @@
...
+        -- Assert it is sharded based on updated ddl info.
+        local conn_s1 = g.cluster:server('s1-master').net_box
+        local result = conn_s1.space['customers_new']:get(1)
+        t.assert_equals(result, case.result.s1)
+
+        local conn_s2 = g.cluster:server('s2-master').net_box
+        local result = conn_s2.space['customers_new']:get(1)
+        t.assert_equals(result, case.result.s2)
+    end
+end
```
Totktonada, line 196: This test imposes relaxed guarantees comparing to what is stated in the commit message. The commit message says:

> Add tests to validate that if space was added at runtime, its sharding
info is fetched to router.

But look at the code (from `master`):

https://github.com/tarantool/crud/blob/d1b1c1aa15388fde282f4ee142333ca1c3907ad4/crud/common/sharding/sharding_metadata.lua#L135-L138

If we have *some* fetched metainfo, we'll use it. If the space is not present here, we'll use the default `strcrc32()` on the primary key.

What is going on here:

1. Before a first request we have no schema, so it'll be fetched. Here the test already incorrect (according to its definition in the commit message): we should perform some request to trigger schema fetch, then add a new space, then issue a request against it.
2. Next, we going to the next test case. The space name in it is the same as in the first one, so we'll just use old cached info. The test cases should use different space names.

Maybe I misunderstood a goal of the commit 'ddl: test new space sharding info is fetched' (I'm in progress of reading the patchset and didn't look at the following patches) and here we want to add a stub to show what is working before next commits? If so, it is not clear from the description of the commit.

crud/common/sharding/storage_metadata_cache.lua
```diff
@@ -0,0 +1,89 @@
...
+    -- It is possible that ddl space was dropped and created again,
+    -- so removing non-existing trigger will cause fail;
+    -- thus we use pcall.
+    pcall(space.on_replace, space, nil, trigger_stash[section])
+
+    trigger_stash[section] = space:on_replace(
+        function(_, new)
+            return update_hash_func(new)
+        end
+    )
```
Totktonada, line 55: Shouldn't it be no-op for the `'applier'` [session type](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_session/type/)?

crud/common/sharding/storage_metadata_cache.lua
```diff
@@ -0,0 +1,89 @@
...
+
+    for _, tuple in space:pairs() do
+        local space_name = tuple[utils.SPACE_NAME_FIELDNO]
+        if cache_data[section][space_name] == nil then
+            update_hash_func(tuple)
+        end
+    end
```
Totktonada, line 62: Duplicate entries? As I [see](https://github.com/tarantool/ddl/blob/f63b291738efeff3262432aab7948075a88cced2/ddl.lua#L95-L111) in tarantool/ddl the `_ddl_sharding_key` and `_ddl_sharding_func` spaces have primary (and so unique) indexes on the space name column.

crud/common/sharding/storage_metadata_cache.lua
```diff
@@ -0,0 +1,89 @@
...
+end
+
+local function get_sharding_hash(space_name, section)
+    if box.space[ddl_space[section]] == nil then
+        return nil
+    end
+
+    if cache_data[section] == nil then
+        init_cache(section)
+    end
```
Totktonada, line 72: Here we lean on assumption that the `_ddl_sharding_*` spaces will not be re-created from scratch. It is okay (at least for now), but I would write the assumption explicitly in a comment in the code.

NB: It could be solved to some extent if we'll store space id for which the cache is created.

How nice it would be to have https://github.com/tarantool/tarantool/discussions/6800, right?

test/integration/ddl_sharding_info_reload_test.lua
DifferentialOrange: Thank you for noticing! In final commit reordering session I messed some things up, so let's clarify them and fix tests.

> Before a first request we have no schema, so it'll be fetched. Here the test already incorrect (according to its definition in the commit message): we should perform some request to trigger schema fetch, then add a new space, then issue a request against it.

I perform a request for non-existing in `before_each` hook, so schema must be fetched before each test case. What I forgot is that to use outdated ddl schema we must have some non-null ddl schema before space is created. I've added a non-empty ddl schema fetch and tests are failing now. I reworked them to be similar to "migration" tests: check if we get hash mismatch error in "ddl: fail on sharding info mismatch" commit and check that info is reloaded in "ddl: reload and retry on sharding info mismatch" commit.

> Next, we going to the next test case. The space name in it is the same as in the first one, so we'll just use old cached info. The test cases should use different space names.

Again, since we get an explicit "space doesn't exist" error in before_test hook, space schema must be updated. Or am I missing something?

crud/common/sharding/storage_metadata_cache.lua
DifferentialOrange: To be honest, I'm not sure how exactly `pairs` work and how it behaves if data is updated in the middle of the cycle with trigger.

If `cache_data[section][space_name]` is not nil, it is already filled with trigger. If we _may get_ outdated info from `pairs` it would lead to inconsistency. On the other hand, this approach ensures that we always have up-to-date info disregarding how pairs works.

crud/insert.lua
DifferentialOrange: Second thought seems useful, thank you! I'll file a ticket for the future

DifferentialOrange: Refer to #274

test/entrypoint/srv_ddl_reload.lua
DifferentialOrange: Fixed, thanks!

crud/common/sharding/storage_metadata_cache.lua
DifferentialOrange: > Here we lean on assumption that the _ddl_sharding_* spaces will not be re-created from scratch.

Yes, if one would mess up with ddl spaces directly (we already have a test doing something like this), it may break. I'll add a comment.

DifferentialOrange: Oh, there's a one safety wrapper already
image

DifferentialOrange: Added a comment

crud/common/sharding/storage_metadata_cache.lua
DifferentialOrange: Cache may be useful for RO-instances too, so I think it shouldn't

DifferentialOrange:
I have responded to all current comments with comment, fix or both.

crud/common/sharding/storage_metadata_cache.lua
Totktonada: You're right.

crud/common/sharding/storage_metadata_cache.lua
Totktonada: AFAIU we can't see a tuple with the same primary key during iteration by the primary index, at least since https://github.com/tarantool/tarantool/issues/1796.

Anyway, we don't do any DML here (trigger is a runtime object), spaces are in memtx, so we'll not yeilds during the iteration.

If we would do some DDL, we should run the loop in a transaction (if a transaction is not opened by a calling code). Here we don't need it, I guess.

I'm okay with any approach here, but, please, add your assumptions into a comment in the code.

crud/common/sharding/storage_metadata_cache.lua
Totktonada: > Oh, there's a one safety wrapper already

The code on the picture will not be executed at all if we already have caches. We'll not invoke the `init_cache()` function. So I expect that the `cache_data[section] == nil` condition will be commented with assumptions under which it implements expected logic.

> Added a comment

I don't see a comment to this code block in the 'ddl: store sharding info hashes on storage' commit.

test/integration/ddl_sharding_info_reload_test.lua
Totktonada: > Again, since we get an explicit "space doesn't exist" error in before_test hook, space schema must be updated. Or am I missing something?

Not in current `master` if 'schema' assumes data from `_ddl_sharding_*` spaces. But since the test is now goes after your changes, it is irrelevant.

crud/common/sharding/storage_metadata_cache.lua
Totktonada: > I don't see a comment to this code block in the 'ddl: store sharding info hashes on storage' commit.

Now I see it.

Totktonada (approved):
Thanks for the patchset!

I have no objections (aside of several nits above). Feel free to merge.

I think that we should revisit our implementation of schema reloading in a future and make it less coupled with other logic. So I assigned #253 to you as the first step into this direction.

crud/common/sharding/storage_metadata_cache.lua
DifferentialOrange: Added a comment

crud/common/sharding_key.lua
```diff
@@ -1,13 +0,0 @@
...
-
--- This method is exported here because
--- we already have customers using old API
--- for updating sharding key cache in their
--- projects like `require('crud.common.sharding_key').update_cache()`
-function sharding_key_cache.update_cache(space_name)
-    return sharding_metadata_module.update_sharding_key_cache(space_name)
-end
-
-return sharding_key_cache
```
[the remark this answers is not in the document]
DifferentialOrange: Restored files, added deprecation log message and changelog entry.

PR #277, opened 2022-04-29
Log empty or nil select calls

### LINKED ISSUE #276: Write in log when see `select(nil)`
Extracted from https://github.com/tarantool/tarantool/issues/6539

After internal discussion, we decided:
- Warn the user about potentially dangerous code
- With logging level - CRIT
- For all cases:
     - `crud:select()`
     - `crud:select(nil)`
     - `crud:select{}`
- With ratelimit (it's worth raising it higher than 0.5 sec for this particular case)
- Add the option fullscan=true to suppress the message in the log

oleg-jukovec:
Empty or nil select calls on user spaces tend to be dangerous.
   
A critical log entry containing the current stack traceback is created
upon empty or nil select calls — a user can explicitly request a full
scan though by passing fullscan = true to select's options table
argument in which a case a log entry will not be created.
    
Tarantool has a similar implementation[1].
    
1. https://github.com/tarantool/tarantool/pull/7064
    
Closes #276

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 * Optimize `crud.select()` without conditions and with `after`.
+* Behaviour of potentially long `select` calls: a critical log entry containing
+  the current stack traceback is created upon such function calls — an user can
+  explicitly request a full scan through by passing `fullscan=true` to `select`
+  options table argument in which a case a log entry will not be created (#276).
 
 ### Fixed
 * `crud.select()` if a condition is '<=' and it's value < `after`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 * Optimize `crud.select()` without conditions and with `after`.
-* Behaviour of potentially long `select` calls: a critical log entry containing
-  the current stack traceback is created upon such function calls — an user can
-  explicitly request a full scan through by passing `fullscan=true` to `select`
-  options table argument in which a case a log entry will not be created (#276).
+* Behaviour of potentially long `select` and `count` calls: a critical log entry
+  containing the current stack traceback is created upon such function calls —
+  an user can explicitly request a full scan through by passing `fullscan=true`
+  to `select` or `count` options table argument in which a case a log entry will
+  not be created (#276).
 
 ### Fixed
 * `crud.select()` if a condition is '<=' and it's value < `after`.
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ It can be used to convert received tuples to objects via `crud.unflatten_rows` f
 For example:
 
 ```lua
-res, err = crud.select('customers')
+res, err = crud.select('customers', nil, {first = 2})
 res
 ---
 - metadata:
@@ -450,6 +450,8 @@ where:
      if full primary key equal condition is specified
   * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
   * `fields` (`?table`) - field names for getting only a subset of fields
+  * `fullscan` (`?boolean`) - if `true` then a critical log entry will be skipped
+    on potentially long `select`, see [avoiding full scan](doc/select.md#avoiding-full-scan).
   * `mode` (`?string`, `read` or `write`) - if `write` is specified then `select` is
     performed on master
   * `prefer_replica` (`?boolean`) - if `true` then the preferred target is one of
@@ -472,7 +474,7 @@ Each condition is a table `{operator, field-identifier, value}`:
 **Example:**
 
 ```lua
-crud.select('customers', {{'<=', 'age', 35}})
+crud.select('customers', {{'<=', 'age', 35}}, {first = 10})
 ---
 - metadata:
   - {'name': 'id', 'type': 'unsigned'}
@@ -496,8 +498,10 @@ See more examples of select queries [here.](https://github.com/tarantool/crud/bl
 ### Pairs
 
 You can iterate across a distributed space using the `crud.pairs` function.
-Its arguments are the same as [`crud.select`](#select) arguments,
-but negative `first` values aren't allowed.
+Its arguments are the same as [`crud.select`](#select) arguments except
+`fullscan` (it does not exist because `crud.pairs` does not generate a critical
+log entry on potentially long requests) and negative `first` values aren't
+allowed.
 User could pass use_tomap flag (false by default) to iterate over flat tuples or objects.
 
 **Example:**
@@ -597,7 +601,7 @@ Returns true or nil with error.
 **Example:**
 
 ```lua
-#crud.select('customers', {{'<=', 'age', 35}})
+#crud.select('customers', {{'<=', 'age', 35}}, {first = 10})
 ---
 - 1
 ...
@@ -605,7 +609,7 @@ crud.truncate('customers', {timeout = 2})
 ---
 - true
 ...
-#crud.select('customers', {{'<=', 'age', 35}})
+#crud.select('customers', {{'<=', 'age', 35}}, {first = 10})
 ---
 - 0
 ...
@@ -633,7 +637,7 @@ Returns number or nil with error.
 Using `memtx`:
 
 ```lua
-#crud.select('customers')
+#crud.select('customers', nil, {fullscan = true})
 ---
 - 5
 ...
--- a/README.md
+++ b/README.md
@@ -690,6 +690,8 @@ where:
   * `force_map_call` (`?boolean`) - if `true`
     then the map call is performed without any optimizations even,
     default value is `false`
+  * `fullscan` (`?boolean`) - if `true` then a critical log entry will be skipped
+    on potentially long `count`, see [avoiding full scan](doc/select.md#avoiding-full-scan).
   * `mode` (`?string`, `read` or `write`) - if `write` is specified then `count` is
     performed on master, default value is `read`
   * `prefer_replica` (`?boolean`) - if `true` then the preferred target is one of
@@ -699,9 +701,9 @@ where:
     default value is `false`
 
 ```lua
-crud.count('customers', {{'<=', 'age', 35}})
+crud.count('customers', {{'==', 'age', 35}})
 ---
-- 5
+- 1
 ...
 ```
 
--- a/crud/count.lua
+++ b/crud/count.lua
@@ -10,6 +10,7 @@ local sharding = require('crud.common.sharding')
 local filters = require('crud.compare.filters')
 local count_plan = require('crud.compare.plan')
 local dev_checks = require('crud.common.dev_checks')
+local ratelimit = require('crud.ratelimit')
 local schema = require('crud.common.schema')
 local sharding_metadata_module = require('crud.common.sharding.sharding_metadata')
 
@@ -90,6 +91,22 @@ function count.init()
     _G._crud.count_on_storage = count_on_storage
 end
 
+local check_count_safety_rl = ratelimit.new()
+local function check_count_safety(space_name, plan, opts)
+    if opts.fullscan == true then
+        return
+    end
+
+    local iter = plan.tarantool_iter
+    if iter == box.index.EQ or iter == box.index.REQ then
+        return
+    end
+
+    local rl = check_count_safety_rl
+    local traceback = debug.traceback()
+    rl:log_crit("Potentially long count from space '%s'\n %s", space_name, traceback)
+end
+
 -- returns result, err, need_reload
 -- need_reload indicates if reloading schema could help
 -- see crud.common.schema.wrap_func_reload()
@@ -98,6 +115,7 @@ local function call_count_on_router(space_name, user_conditions, opts)
         timeout = '?number',
         bucket_id = '?number|cdata',
         force_map_call = '?boolean',
+        fullscan = '?boolean',
         yield_every = '?number',
         prefer_replica = '?boolean',
         balance = '?boolean',
@@ -147,6 +165,7 @@ local function call_count_on_router(space_name, user_conditions, opts)
     if err ~= nil then
         return nil, CountError:new("Failed to plan count: %s", err), const.NEED_SCHEMA_RELOAD
     end
+    check_count_safety(space_name, plan, opts)
 
     -- set replicasets to count from
     local replicasets_to_count = replicasets
@@ -297,6 +316,7 @@ function count.call(space_name, user_conditions, opts)
         timeout = '?number',
         bucket_id = '?number|cdata',
         force_map_call = '?boolean',
+        fullscan = '?boolean',
         yield_every = '?number',
         prefer_replica = '?boolean',
         balance = '?boolean',
--- a/crud/ratelimit.lua
+++ b/crud/ratelimit.lua
@@ -1,0 +1,125 @@
+-- Mostly it's a copy-paste from tarantool/tarantool log.lua:
+-- https://github.com/tarantool/tarantool/blob/29654ffe3638e5a218dd32f1788830ff05c1c05c/src/lua/log.lua
+--
+-- We have three reasons for the copy-paste:
+-- 1. Tarantool has not log.crit() (a function for logging with CRIT level).
+-- 2. Only new versions of Tarantool have Ratelimit type.
+-- 3. We want own copy of Ratelimit in case the implementation in Tarantool
+-- changes. Less pain between Tarantool versions.
+local ffi = require('ffi')
+
+local S_CRIT = ffi.C.S_CRIT
+local S_WARN = ffi.C.S_WARN
+
+local function say(level, fmt, ...)
+    if ffi.C.log_level < level then
+        -- don't waste cycles on debug.getinfo()
+        return
+    end
+    local type_fmt = type(fmt)
+    local format = "%s"
+    if select('#', ...) ~= 0 then
+        local stat
+        stat, fmt = pcall(string.format, fmt, ...)
+        if not stat then
+            error(fmt, 3)
+        end
+    elseif type_fmt == 'table' then
+        -- An implementation in tarantool/tarantool supports encoding a table in
+        -- JSON, but it requires more dependencies from FFI. So we just deleted
+        -- it because we don't need such encoding in the module.
+        error("table not supported", 3)
+    elseif type_fmt ~= 'string' then
+        fmt = tostring(fmt)
+    end
+
+    local debug = require('debug')
+    local frame = debug.getinfo(3, "Sl")
+    local line, file = 0, 'eval'
+    if type(frame) == 'table' then
+        line = frame.currentline or 0
+        file = frame.short_src or frame.src or 'eval'
+    end
+
+    ffi.C._say(level, file, line, nil, format, fmt)
+end
+
+local ratelimit_enabled = true
+
+local function ratelimit_enable()
+    ratelimit_enabled = true
+end
+
+local function ratelimit_disable()
+    ratelimit_enabled = false
+end
+
+local Ratelimit = {
+    interval = 60,
+    burst = 10,
+    emitted = 0,
+    suppressed = 0,
+    start = 0,
+}
+
+local function ratelimit_new(object)
+    return Ratelimit:new(object)
+end
+
+function Ratelimit:new(object)
+    object = object or {}
+    setmetatable(object, self)
+    self.__index = self
+    return object
+end
+
+function Ratelimit:check()
+    if not ratelimit_enabled then
+        return 0, true
+    end
+
+    local clock = require('clock')
+    local now = clock.monotonic()
+    local saved_suppressed = 0
+    if now > self.start + self.interval then
+        saved_suppressed = self.suppressed
+        self.suppressed = 0
+        self.emitted = 0
+        self.start = now
+    end
+
+    if self.emitted < self.burst then
+        self.emitted = self.emitted + 1
+        return saved_suppressed, true
+    end
+    self.suppressed = self.suppressed + 1
+    return saved_suppressed, false
+end
+
+function Ratelimit:log_check(lvl)
+    local suppressed, ok = self:check()
+    if lvl >= S_WARN and suppressed > 0 then
+        say(S_WARN, '%d messages suppressed due to rate limiting', suppressed)
+    end
+    return ok
+end
+
+function Ratelimit:log(lvl, fmt, ...)
+    if self:log_check(lvl) then
+        say(lvl, fmt, ...)
+    end
+end
+
+local function log_ratelimited_closure(lvl)
+    return function(self, fmt, ...)
+        self:log(lvl, fmt, ...)
+    end
+end
+
+Ratelimit.log_crit = log_ratelimited_closure(S_CRIT)
+
+return {
+    new = ratelimit_new,
+    enable = ratelimit_enable,
+    disable = ratelimit_disable,
+}
--- a/crud/select/compat/common.lua
+++ b/crud/select/compat/common.lua
@@ -1,6 +1,28 @@
+local ratelimit = require('crud.ratelimit')
+local check_select_safety_rl = ratelimit.new()
+
 local common = {}
 
 common.SELECT_FUNC_NAME = '_crud.select_on_storage'
 common.DEFAULT_BATCH_SIZE = 100
+
+common.check_select_safety = function(space_name, plan, opts)
+    if opts.fullscan == true then
+        return
+    end
+
+    if opts.first ~= nil and math.abs(opts.first) <= 1000 then
+        return
+    end
+
+    local iter = plan.tarantool_iter
+    if iter == box.index.EQ or iter == box.index.REQ then
+        return
+    end
+
+    local rl = check_select_safety_rl
+    local traceback = debug.traceback()
+    rl:log_crit("Potentially long select from space '%s'\n %s", space_name, traceback)
+end
 
 return common
--- a/crud/select/compat/select.lua
+++ b/crud/select/compat/select.lua
@@ -180,6 +180,7 @@ local function build_select_iterator(space_name, user_conditions, opts)
     return {
         tuples_limit = tuples_limit,
         merger = merger,
+        plan = plan,
         space_format = filtered_space_format,
     }
 end
@@ -252,14 +253,16 @@ local function select_module_call_xc(space_name, user_conditions, opts)
     checks('string', '?table', {
         after = '?table|cdata',
         first = '?number',
-        timeout = '?number',
         batch_size = '?number',
         bucket_id = '?number|cdata',
         force_map_call = '?boolean',
         fields = '?table',
+        fullscan = '?boolean',
+
+        mode = '?vshard_call_mode',
         prefer_replica = '?boolean',
         balance = '?boolean',
-        mode = '?vshard_call_mode',
+        timeout = '?number',
     })
 
     opts = opts or {}
@@ -292,6 +295,7 @@ local function select_module_call_xc(space_name, user_conditions, opts)
     if err ~= nil then
         return nil, err
     end
+    common.check_select_safety(space_name, iter.plan, opts)
 
     local tuples = {}
 
--- a/doc/select.md
+++ b/doc/select.md
@@ -24,7 +24,7 @@
 
 ## Filtering
 
-``CRUD`` allows to filter tuples by conditions. Each condition can use field name (or number) or index name. The first condition that uses index name is used to iterate over space. If there is no conditions that match index names, full scan is performed. Other conditions are used as additional filters. Search condition for the indexed field must be placed first to avoid a full scan.
+``CRUD`` allows to filter tuples by conditions. Each condition can use field name (or number) or index name. The first condition that uses index name is used to iterate over space. If there is no conditions that match index names, full scan is performed. Other conditions are used as additional filters. Search condition for the indexed field must be placed first to avoid a full scan. In additional, don't forget to limit amount of results with ``first`` parameter. This will help to avoid too long selects in production.
 
 **Note:** If you specify sharding key or ``bucket_id`` select will be performed on single node. Otherwise Map-Reduce over all nodes will be occurred.
 
@@ -107,7 +107,7 @@ We have an ``age_index`` index. Example below gets a list of ``customers`` over 
 **Example:**
 
 ```lua
-crud.select('developers', {{'>=', 'age_index', 30}})
+crud.select('developers', {{'>=', 'age_index', 30}}, {first = 10})
 ---
 - metadata:
   - {'name': 'id', 'type': 'unsigned'}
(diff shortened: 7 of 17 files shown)


crud/ratelimit.lua
```diff
@@ -0,0 +1,99 @@
...
+    end
+    local type_fmt = type(fmt)
+    local format = "%s"
+    if select('#', ...) ~= 0 then
+        local stat
+        stat, fmt = pcall(string.format, fmt, ...)
+        if not stat then
+            error(fmt, 3)
+        end
+    elseif type_fmt ~= 'string' then
```
Totktonada, line 25: How about commenting differences from tarantool's implementation? We don't use table as a log entry in the module -- it is completely ok to don't implement conversion to json here -- but I would leave a comment regarding removed parts of the implementation to decrease possible confusion.

crud/select/compat/common.lua
```diff
@@ -1,6 +1,20 @@
+local ratelimit = require'crud.ratelimit'
```
Totktonada, line 1: Nit: Please, use `local bar = require('foo.bar')` form (with parentheses). See https://www.tarantool.io/en/doc/latest/dev_guide/lua_style_guide/#idioms-and-patterns.

crud/select/compat/select.lua
```diff
@@ -193,6 +193,7 @@ function select_module.pairs(space_name, user_conditions, opts)
         bucket_id = '?number|cdata',
         force_map_call = '?boolean',
         fields = '?table',
+        fullscan = '?boolean',
```
Totktonada, line 196: NB: I asked the product team whether we want to issue the warning for `crud.pairs()` (not only `crud.select()`). I'll return back here with an update tomorrow.

crud/select/compat/select.lua
```diff
@@ -252,15 +253,18 @@ local function select_module_call_xc(space_name, user_conditions, opts)
     checks('string', '?table', {
         after = '?table|cdata',
         first = '?number',
-        timeout = '?number',
         batch_size = '?number',
         bucket_id = '?number|cdata',
         force_map_call = '?boolean',
         fields = '?table',
+        fullscan = '?boolean',
```
Totktonada, line 260: Please, add the new option into README.

crud/select/compat/select.lua
oleg-jukovec: No pairs mentioned after this comment.
https://github.com/tarantool/tarantool/issues/6539#issuecomment-1073753815

test/unit/select_nil_test.lua
```diff
@@ -0,0 +1,54 @@
...
+    end
+end
+
+g.test_no_warnings = function()
+    local opts = {
+        {uc = nil, opts = {fullscan = true}},
+        {uc = {}, opts = {fullscan = true}},
+        {uc = {0}, opts = nil},
+        {uc = {0}, opts = {}},
+        {uc = {0}, opts = {fullscan = false}},
```
Totktonada, line 40: I would prefer to test it with a correct condition just in case. Otherwise we lean on implementation details (order of the checks) more than it is necessary for a simple test.

test/unit/select_nil_test.lua
```diff
@@ -0,0 +1,54 @@
...
+
+    for _, o in ipairs(opts) do
+        local capture = luatest_capture:new()
+
+        capture:wrap(true, function()
+            pcall(crud.select, any_space, o.uc, o.opts)
+        end)
+
+        local captured = capture:flush()
+        t.assert_equals(captured.stderr, "")
```
Totktonada, line 51: Nit: I would prefer a bit more specific check: whether given pattern is not present in the output. I'm afraid of the test flakiness due to work of other fibers (say, spawned by other tests) or due to audit log in the tarantool enterprise.

Now we don't yield before the check and audit log does not say anything about Lua calls, but it is better to make things less fragile when it is simple to do.

crud/ratelimit.lua
```diff
@@ -0,0 +1,99 @@
...
+        fmt = tostring(fmt)
+    else
+        return
+    end
```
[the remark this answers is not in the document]
oleg-jukovec: Yes, I got confused in Lua. Fixed.

crud/ratelimit.lua
oleg-jukovec: Great idea, done.

crud/select/compat/common.lua
oleg-jukovec: Thanks, I'll remember that.

crud/select/compat/select.lua
oleg-jukovec: Done + I have updated examples from the file.

test/unit/select_nil_test.lua
oleg-jukovec: Thank you, I rewrote the test into an integration test.

test/unit/select_nil_test.lua
oleg-jukovec: Thank you. Updated, now I try to find a substring.

crud/select/compat/common.lua
```diff
@@ -1,6 +1,20 @@
...
 
+common.check_select_args = function(space_name, user_conditions, opts)
+    local rl = check_select_args_rl
+    local uc = user_conditions
+
+    if (type(uc) == 'nil' or (type(uc) == 'table' and next(uc) == nil)) and
+       (opts == nil or not opts.fullscan) then
+        rl:log_crit("empty or nil `select` call on user space=%s\n %s",
+                    space_name, debug.traceback())
+    end
```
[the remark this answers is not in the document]
Totktonada: @NeraverinTarantool said that the product team decided to leave it as is. While it is obvious what to do with `limit = 1`, things get less obvious for `limit = 10000`. There is no good universal threshold.

crud/select/compat/select.lua
Totktonada: @NeraverinTarantool said that the product team decided to don't touch `crud.pairs()`.

It seems, this line (and the similar one in `select_old.lua`) should be removed.

README.md
```diff
@@ -25,7 +25,7 @@ It can be used to convert received tuples to objects via `crud.unflatten_rows` f
 For example:
 
 ```lua
-res, err = crud.select('customers')
+res, err = crud.select('customers', nil, {fullscan = true})
```
Totktonada, line 28: As part of this activity we had intention to get rid of unlimited selects in the documentation. I would consider crud examples as part of this activity. It is not nice to suggest `fullscan = true` in the first example.

README.md
```diff
@@ -395,6 +395,8 @@ where:
      if full primary key equal condition is specified
   * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
   * `fields` (`?table`) - field names for getting only a subset of fields
+  * `fullscan` (`?boolean`) - if `true` then critical log entry will be skipped
+    on empty or nil `select`
```
Totktonada, line 399: `crud.pairs()` description has the following phrase:

> Its arguments are the same as `crud.select` arguments,
> but negative `first` values aren't allowed.

It should be corrected.

oleg-jukovec: Updated.
```patch
--- a/README.md
+++ b/README.md
@@ -441,8 +443,9 @@ See more examples of select queries [here.](https://github.com/tarantool/crud/bl
 ### Pairs
 
 You can iterate across a distributed space using the `crud.pairs` function.
-Its arguments are the same as [`crud.select`](#select) arguments,
-but negative `first` values aren't allowed.
+Its arguments are the same as [`crud.select`](#select) arguments except
+`fullscan`(it does not exist because `crud.pairs` does not generate a critical
+log entry on empty or nil conditions), but negative `first` values aren't allowed.
 User could pass use_tomap flag (false by default) to iterate over flat tuples or objects.
 
 **Example:**
 ```

File: README.md (hunk 9 above)
Comment (oleg-jukovec) [reply README.md]: Sure. It has been replaced with a condition.

Similar actions were done in the doc/select.md. There is an unlimited select, but I think it is logical from the context:
```patch
--- a/doc/select.md
+++ b/doc/select.md
@@ -60,11 +60,11 @@ Let's check ``developers`` space content to make other examples more clear. Just
 **Example:**
 
 ```lua
-crud.select('developers', nil, { first = 6 })
+crud.select('developers', nil, { first = 6, fullscan = true })
```
Is it better to replace it with a condition?

crud/select/compat/select.lua
oleg-jukovec: The change has been reverted.

crud/select/compat/common.lua
```diff
@@ -1,6 +1,20 @@
...
 
+common.check_select_args = function(space_name, user_conditions, opts)
+    local rl = check_select_args_rl
+    local uc = user_conditions
+
+    if (uc == nil or (type(uc) == 'table' and next(uc) == nil)) and
+       (opts == nil or not opts.fullscan) then
+        rl:log_crit("empty or nil `select` call on user space=%s\n %s",
+                    space_name, debug.traceback())
+    end
```
Totktonada, line 17: Yet another design flaw: we can set a condition by a non-indexed field (or a field that is not the first one in any index) and get full scan without the warning. Asked products managers...

README.md
Totktonada: Before I read the resulting documentation, I had thought that the proposed approach may be helpful in some cases despite many design flaws. Now I see that it confuses more than helps.

We have forward and backward pagination examples. All select calls are limited, but they need either `fullscan = true` or `{'>', 'id', 0}` condition (which is obviously just to get rid of the warning, because it is evaluated to true for all tuples). It has no sense.

But first example from the 'Avoiding full scan` section, which performs full scan, will not give the warning. It has no sense too.

Ugh. I reached product managers once again. I don't know, how to review the implementation, if the proposal itself has so many flaws.

Please, don't take it as criticism of your work. It is mine fault: ideally I should spot all those points before asking a developer to start working on the implementation.

test/integration/select_test.lua
```diff
@@ -77,6 +78,54 @@ pgroup.test_not_valid_value_type = function(g)
     t.assert_str_contains(err.err, "Supplied key type of part 0 does not match index part type: expected unsigned")
 end
 
+pgroup.test_select_nil = function(g)
+    local space = 'customers'
+    local crit_log = "C> empty or nil `select` call on user space=" .. space
+    local valid_condition = {'>=', 'id', 2}
+    local tests = {
```
DifferentialOrange, line 85: Such tests will fail at first invalid assert. I prefer taking out parametrized data and creating separate cases for each entry (like this: https://github.com/tarantool/crud/blob/20023a9a6faab1976997534e11c3b71f24d9061f/test/unit/stats_test.lua#L93-L209), so each failed case will fail and each successful case will succeed.

test/integration/select_test.lua
```diff
@@ -77,6 +78,54 @@ pgroup.test_not_valid_value_type = function(g)
...
+            t.assert_equals(err, nil)
+        end
+
+        -- We have a delay here. This hack helps to wait for the end of the output.
+        -- It shouldn't take much time.
+        g.cluster:server('router').net_box:eval([[
+            require('log').error("end of test_select_nil case")
+        ]])
+        local captured = ""
+        while not string.find(captured, "end of test_select_nil case", 1, true) do
```
DifferentialOrange, line 116: Oh, that hack is really cool! We definitely should use it in similar cases when needed.

test/integration/select_test.lua
```diff
@@ -77,6 +78,54 @@ pgroup.test_not_valid_value_type = function(g)
...
+    local space = 'customers'
+    local crit_log = "C> empty or nil `select` call on user space=" .. space
+    local valid_condition = {'>=', 'id', 2}
+    local tests = {
+        {crit = true, err = false, uc = nil, opts = nil},
+        {crit = true, err = false, uc = nil, opts = {}},
+        {crit = true, err = false, uc = nil, opts = {fullscan = false}},
+        {crit = true, err = false, uc = {}, opts = nil},
+        {crit = true, err = false, uc = {}, opts = {}},
+        {crit = true, err = false, uc = {}, opts = {fullscan = false}},
```
DifferentialOrange, line 91: Why are there no `crit = true, error = true` cases? It is hard to implement?

test/integration/select_test.lua
DifferentialOrange, line 116: But when we log first `end of test_select_nil case` string, this `while` loop will be useless for next ones, since there is a single capture for all of them. One more reason to separate test cases.

DifferentialOrange (commented):
Thank you for your patch! (And sorry for the long response.) I have left several comments.

The implementation seems legitimate, but there are some active discussions in tarantool/tarantool about processing `select(nil)` with `limit` now. Seems like it's better to wait for a newer version of `select(nil)` approach in core tarantool and add corresponding fixes here or discuss it with product managers.

test/integration/select_test.lua
oleg-jukovec: capture:flush() does not return all captured data for the entire time, but a chunk of data between capture:flush() calls. The code works as expected (waits only for one case).

DifferentialOrange: Oh, ok. Thank for the clarification.

README.md
oleg-jukovec: requirements have been updated:
https://github.com/tarantool/tarantool/issues/7129

crud/select/compat/common.lua
oleg-jukovec: requirements have been updated:
https://github.com/tarantool/tarantool/issues/7129

oleg-jukovec:
The pull request has been updated according to the new requirements: https://github.com/tarantool/tarantool/issues/7129

test/integration/select_test.lua
oleg-jukovec: The tests have been rewritten.

test/integration/select_test.lua
oleg-jukovec: I extracted it to a separate function.

test/integration/select_test.lua
oleg-jukovec: Great idea, done.

crud/ratelimit.lua
```diff
@@ -0,0 +1,102 @@
+-- Mostly it's a copy-paste from tarantool/src/lua/log.lua. We have three
```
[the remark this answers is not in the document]
oleg-jukovec: Done.

crud/ratelimit.lua
```diff
@@ -0,0 +1,102 @@
...
+    local type_fmt = type(fmt)
+    local format = "%s"
+    if select('#', ...) ~= 0 then
+        local stat
+        stat, fmt = pcall(string.format, fmt, ...)
+        if not stat then
+            error(fmt, 3)
+        end
+    elseif type_fmt == 'table' then
+        -- An implementation in Tarantool supports encoding a table in JSON,
```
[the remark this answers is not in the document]
oleg-jukovec: No problems, replaced.

README.md
```diff
@@ -441,8 +443,9 @@ See more examples of select queries [here.](https://github.com/tarantool/crud/bl
 ### Pairs
 
 You can iterate across a distributed space using the `crud.pairs` function.
-Its arguments are the same as [`crud.select`](#select) arguments,
-but negative `first` values aren't allowed.
+Its arguments are the same as [`crud.select`](#select) arguments except
+`fullscan`(it does not exist because `crud.pairs` does not generate a critical
+log entry on empty or nil conditions), but negative `first` values aren't allowed.
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, replaced.

crud/select/compat/common.lua
```diff
@@ -1,6 +1,32 @@
...
+            if it ~= nil and type(it) == 'string' and (it == '=' or it == '==') then
+                return
+            end
+        end
+    end
+
+    local rl = check_select_safety_rl
+    local traceback = debug.traceback()
+    rl:log_crit("Potentially long select from space '%s'\n %s", space_name, traceback)
+end
```
Totktonada, line 30: > a full scan will be performed, since non-indexed field is placed first in search conditions

https://github.com/tarantool/crud/blob/52b2b4a78832ee8c97a8b946cb1764fcf59f0755/doc/select.md#avoiding-full-scan

I think it would be better to use query plan information (see `crud/compare/plan.lua`) to check, whether full scan will be performed. We need database schema information to actually say, whether particular select query will fall into the linear space scan. The query planning already uses the schema to create the plan.

crud/select/compat/common.lua
Totktonada: It is still relevant. Tarantool itself has pretty simple query language, but crud's one a bit more powerful. It requires more analysis to actually say, whether the query will fall into the linear space scan. I propose to consider using the query plan information.

crud/ratelimit.lua
```diff
@@ -0,0 +1,125 @@
...
+    self.__index = self
+    return object
+end
+
+function Ratelimit:check()
+    if not ratelimit_enabled then
+        return 0, true
+    end
+
+    local clock = require('clock')
```
DifferentialOrange, line 81: We make all requires on top-level (and I don't see any reason for in-function require here). Even if it's copypasted from core, I think we should move this line.

crud/select/compat/common.lua
```diff
@@ -1,6 +1,32 @@
+local ratelimit = require('crud.ratelimit')
+local check_select_safety_rl = ratelimit.new()
```
DifferentialOrange, line 2: For now we don't set any limits on warnings count, aren't we?

test/integration/count_test.lua
```diff
@@ -26,6 +27,9 @@ pgroup.before_all(function(g)
     g.cluster:server('router').net_box:eval([[
         require('crud').cfg{ stats = true }
     ]])
+    g.cluster:server('router').net_box:eval([[
+        require('crud.ratelimit').disable()
```
DifferentialOrange, line 31: What do we need it for?

test/integration/select_test.lua
```diff
@@ -77,6 +81,111 @@ pgroup.test_not_valid_value_type = function(g)
...
+        has_crit = false,
+        user_conditions = nil,
+        opts = {fullscan = true},
+    },
+    limit = {
+        has_crit = false,
+        user_conditions = nil,
+        opts = {first = 123},
+    },
+    min_limit = {
```
DifferentialOrange, line 105: Well, it's more like max negative limit

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 ### Changed
+* Behaviour of potentially long `select` and `count` calls (`>`, `>=`, `<`, `<=`
+  and `first` > 1000): a critical log entry containing the current stack
+  traceback is created upon such function calls — an user can explicitly request
+  a full scan though by passing fullscan=true to select's options table argument
```
DifferentialOrange, line 16:
```suggestion
  a full scan though by passing `fullscan=true` to `select`'s options table argument
```

crud/select/compat/common.lua
```diff
@@ -1,6 +1,32 @@
...
+            local it = v[1]
+            if it ~= nil and type(it) == 'string' and (it == '=' or it == '==') then
+                return
+            end
+        end
+    end
+
+    local rl = check_select_safety_rl
+    local traceback = debug.traceback()
+    rl:log_crit("Potentially long select from space '%s'\n %s", space_name, traceback)
```
DifferentialOrange, line 29: This message is a bit non-informative: it is hard to understand what should we do to remove potentially long selects from the application. Even though I don't have any alternatives for the message, I think we should document warning conditions in README so developer may understand what really "potentially long select" is.

README.md
```diff
@@ -397,6 +397,8 @@ where:
      if full primary key equal condition is specified
   * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
   * `fields` (`?table`) - field names for getting only a subset of fields
+  * `fullscan` (`?boolean`) - if `true` then a critical log entry will be skipped
+    on potentially long `select` (`>`, `>=`, `<`, `<=` and `first` > 1000)
```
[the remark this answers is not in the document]
oleg-jukovec: ```offset=N``` just iterates over an index N times. ```after``` acts in other way: iterates over an index with GE or LE from the tuple. So we don't know "offset" of ```after```.

But yes, firs t!= limit, it can fetch a lot more tuples. But it seems to me that if we start to care about it, it will be too difficult to understand for an user. first <= 1000 is very simple and obvious condition.

crud/select/compat/common.lua
oleg-jukovec: Default values are used (as in the Tarantool implementation).

crud/select/compat/common.lua
oleg-jukovec: The message is copy-paste from the Tarantool implementation:
https://github.com/tarantool/tarantool/pull/7131#pullrequestreview-972344813

I will add a note to README.

crud/ratelimit.lua
oleg-jukovec: There is something to optimize here. For example, debug.traceback() is called every time, not just when it used by the ratelimit. In general, I don't like the implementation. But I prefer copy-paste in this case: it will be very easy to update the logic after update in core.

Especially that the change is about code style, not about a critical defect.

test/integration/count_test.lua
oleg-jukovec: It disable the ratelimit for tests:

https://github.com/tarantool/tarantool/pull/7131#pullrequestreview-972405738
https://github.com/tarantool/tarantool/pull/7131#pullrequestreview-973719435

crud/ratelimit.lua
DifferentialOrange: It's not about code style: it has performance overhead. But a small one, to be honest. (But an easily fixable one.)

Well, it's ok to leave it as is due to copypaste.

README.md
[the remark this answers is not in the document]
DifferentialOrange: Yeah, I understand that it's really difficult to implement. Just decided to clarify this case explicitly.

crud/select/compat/common.lua
DifferentialOrange: Ok, thanks

crud/select/compat/common.lua
DifferentialOrange: Oh, ok. Sorry, I have only seen the previous version 
image

README.md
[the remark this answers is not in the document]
Totktonada: > `after` acts in other way: iterates over an index with GE or LE from the tuple.

If it is so in all cases (I don't know), we should not count it. We should count only tuples, which are linearly traversed (it is why the `offset` value is counted).

BTW, it would be good to leave some internal documentation regarding the query planner fields. Their meaning is not obvious.

Say, what the executor will do, when both 'scan value' and 'after scan value' are present? I guess, it'll use 'after scan value' with GT/LT and transform 'scan value' to a post-filter, but I'm not completely sure.

Whether the 'after scan value' always corresponds to the index determined by the first select condition? What if the first condition is not backed by any index? Will the 'after' condition be transformed to a post-filter (rather than GT/LT on an index) in the case?

Now it is hard to think about different cases, to be honest.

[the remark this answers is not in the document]
Totktonada: I think it worth to write a developer documentation on the topic. I filed https://github.com/tarantool/crud/issues/292.

crud/select/compat/common.lua
```diff
@@ -1,6 +1,32 @@
...
+    end
+
+    if opts.first ~= nil and math.abs(opts.first) <= 1000 then
+        return
+    end
+
+    if user_conditions ~= nil then
+        for _, v in ipairs(user_conditions) do
+            local it = v[1]
+            if it ~= nil and type(it) == 'string' and (it == '=' or it == '==') then
```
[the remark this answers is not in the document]
oleg-jukovec: Outdated.

crud/count.lua
```diff
@@ -90,6 +91,26 @@ function count.init()
...
+local check_count_safety_rl = ratelimit.new()
+local function check_count_safety(space_name, user_conditions, opts)
+    if opts.fullscan ~= nil and opts.fullscan == true then
+        return
+    end
+
+    if user_conditions ~= nil then
+        for _, v in ipairs(user_conditions) do
+            local it = v[1]
+            if it ~= nil and type(it) == 'string' and (it == '=' or it == '==') then
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you. Outdated.

README.md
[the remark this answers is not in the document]
oleg-jukovec: I described the algorithm in #292.

In short, we can consider `after` as a regular full scan. Sometimes we can additionally skip some tuples in `after` case (if a key from condition == a key from after), but I think it's not critical. In any case, there is no way to calculate amount of the tuples in advance.

crud/select/compat/common.lua
oleg-jukovec: Updated.

crud/select/compat/common.lua
```diff
@@ -1,6 +1,32 @@
+local ratelimit = require('crud.ratelimit')
+local check_select_safety_rl = ratelimit.new()
+
 local common = {}
 
 common.SELECT_FUNC_NAME = '_crud.select_on_storage'
 common.DEFAULT_BATCH_SIZE = 100
 
+common.check_select_safety = function(space_name, user_conditions, opts)
+    if opts.fullscan ~= nil and opts.fullscan == true then
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, fixed.

crud/count.lua
```diff
@@ -90,6 +91,26 @@ function count.init()
     _G._crud.count_on_storage = count_on_storage
 end
 
+local check_count_safety_rl = ratelimit.new()
+local function check_count_safety(space_name, user_conditions, opts)
+    if opts.fullscan ~= nil and opts.fullscan == true then
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, fixed too.

test/integration/select_test.lua
oleg-jukovec: No problems. I've renamed it to `max_negative_limit`.

CHANGELOG.md
oleg-jukovec: Thank you, fixed.

README.md
```diff
@@ -662,6 +666,8 @@ where:
   * `force_map_call` (`?boolean`) - if `true`
     then the map call is performed without any optimizations even,
     default value is `false`
+  * `fullscan` (`?boolean`) - if `true` then a critical log entry will be skipped
+    on potentially long `select`, see [avoiding full scan](https://github.com/tarantool/crud/blob/master/doc/select.md#avoiding-full-scan).
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, fixed! I need to be more careful with copy-paste.

README.md
```diff
@@ -426,6 +426,8 @@ where:
      if full primary key equal condition is specified
   * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
   * `fields` (`?table`) - field names for getting only a subset of fields
+  * `fullscan` (`?boolean`) - if `true` then a critical log entry will be skipped
+    on potentially long `count`, see [avoiding full scan](https://github.com/tarantool/crud/blob/master/doc/select.md#avoiding-full-scan).
```
AnaNek, line 430: In my opinion, you mixed up the place of correction, this description was correct ('...on potentially `select`...'), the correction had to be done in `count` section which is below.

README.md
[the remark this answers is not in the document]
AnaNek: https://github.com/tarantool/crud/pull/277#discussion_r881423309

README.md
oleg-jukovec: Yes, I'm sorry. I hope it is correct now.

AnaNek: It's correct now, thank you!

oleg-jukovec:
I will delete a doc entry:
```
**Note:**

Pagination with parameter ``after`` leads to a full scan even with the condition '=' or '==' on the index fields.
...
from doc/select.md after #295 (or I'll do it there, it depends on the merging order).

**Review by DifferentialOrange — APPROVED** [review]
I think we should merge https://github.com/tarantool/crud/pull/295 first and then deal with this PR.

Are there any plans to add more critical log conditions or we decided to stop for now?

**Comment by oleg-jukovec** [comment]
> Are there any plans to add more critical log conditions or we decided to stop for now?

I think that we decided to stop with current conditions.

File: README.md (hunk 25 above)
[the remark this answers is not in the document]
Comment (Totktonada) [reply README.md]: It looks as something that we'll completely resolve in PR #295. Closing the discussion so.

File: README.md [hunk 32]
```diff
@@ -671,9 +677,9 @@ where:
     default value is `false`
 
 ```lua
-crud.count('customers', {{'<=', 'age', 35}})
+crud.count('customers', {{'==', 'age', 35}})
```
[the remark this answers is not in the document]
Totktonada: A limit (`first`) for `crud.count()` has a bit different meaning: it is not how much data we want to see, but rather limiting load on the service. Soft thresholds for them possibly should be different.

I'm okay to leave it as is and maybe return back after users' feedback. Resolving the discussion so.

doc/select.md
```diff
@@ -202,11 +202,62 @@ crud.select('developers', {{'==', 'surname', "Adams"}})
...
+  - {'name': 'bucket_id', 'type': 'unsigned'}
+  - {'name': 'name', 'type': 'string'}
+  - {'name': 'surname', 'type': 'string'}
+  - {'name': 'age', 'type': 'number'}
+  rows:
+  - [3, 9661, 'Pavel', 'Adams', 27]
+...
+```
+
+You can avoid the full scan with '=' or '==' condition on an index fields.
```
Totktonada, line 236: Nit: an index fields -> index fields (plural).

Also the order of conditions is important. I would highlight it.

Totktonada (approved):

test/helper.lua
```diff
@@ -509,4 +509,17 @@ function helpers.assert_timeout_error(value, message)
     error(err, 2)
 end
 
+function helpers.fflush_cluster_stdout(cluster, capture)
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you for noticing.

I've renamed it:
fflush_cluster_stdout -> fflush_main_server_stdout
and now it calls eval from `main_server` not from `router` because in integration tests we run queries on the `main_server`.

doc/select.md
Totktonada: As I understood, the following sentence (it is removed from the select.md document) is just incorrect:

> In this case, a full scan will be performed, since non-indexed field is placed first in search conditions. Example below shows how you can avoid a full scan.

I can add a condition on a non-indexed field as the first condition and it will not change anything about the query plan (we'll not go to the linear scanning). Correct?

oleg-jukovec: Yes, non-index conditions don't change anything. Only a first condition on index fields matters.

AnaNek (approved):

oleg-jukovec:
A note `first leads to a  full scan` has been removed from `doc/select.md`. The pull request should be merged after #295 .

oleg-jukovec:
The test `run-tests-ce (1.10.6, false) (push) ` failed because `tarantool.io` has expired certificate.

PR #279, opened 2022-05-06
doc: add a script to ease experimenting with crud

Totktonada:
Basic usage:

```shell
$ ./doc/playground.lua
tarantool> crud.select('developers', nil, {first = 6})
tarantool> crud.<something else>(<...>)
```

What the script is doing, step by step:

* Configure the database to don't leave WAL and snapshot files.
* Configure and bootstrap vshard storage and router on the single
  instance.
* Create a space and fill it with data.
* Start a console.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ It also provides the `crud-storage` and `crud-router` roles for
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
+- [Quickstart](#quickstart)
 - [API](#api)
   - [Insert](#insert)
   - [Get](#get)
@@ -36,6 +37,29 @@ It also provides the `crud-storage` and `crud-router` roles for
 - [License](#license)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Quickstart
+
+First, [install Tarantool](https://www.tarantool.io/en/download).
+
+Now you have the following options how to learn crud API and use it in a
+project:
+
+* Play with crud on a testing dataset on a single instance:
+
+  ```shell
+  $ git clone https://github.com/tarantool/crud.git
+  $ cd crud
+  $ tarantoolctl rocks make
+  $ ./doc/playground.lua
+  tarantool> crud.select('customers', {{'<=', 'age', 35}}, {first = 10})
+  tarantool> crud.select('developers', nil, {first = 6})
+  ```
+* Add crud into dependencies of a Cartridge application and add crud roles into
+  dependencies of your roles (see [Cartridge roles](#cartridge-roles) section).
+* Add crud into dependencies of your application (rockspec, RPM spec -- depends
+  of your choice) and call crud initialization code from storage and router
+  code (see [API](#api) section).
 
 ## API
 
--- a/doc/playground.lua
+++ b/doc/playground.lua
@@ -1,0 +1,150 @@
+#!/usr/bin/env tarantool
+
+-- How to run:
+--
+-- $ ./doc/playground.lua
+--
+-- Or
+--
+-- $ KEEP_DATA=1 ./doc/playground.lua
+--
+-- What to do next:
+--
+-- Choose an example from README.md, doc/select.md or doc/pairs.md
+-- and run. For example:
+--
+-- tarantool> crud.select('customers', {{'<=', 'age', 35}}, {first = 10})
+-- tarantool> crud.select('developers', nil, {first = 6})
+
+local fio = require('fio')
+local console = require('console')
+local vshard = require('vshard')
+local crud = require('crud')
+
+-- Trick to don't leave *.snap, *.xlog files. See
+-- test/tuple_keydef.test.lua in the tuple-keydef module.
+if os.getenv('KEEP_DATA') ~= nil then
+    box.cfg()
+else
+    local tempdir = fio.tempdir()
+    box.cfg({
+        memtx_dir = tempdir,
+        wal_mode = 'none',
+    })
+    fio.rmtree(tempdir)
+end
+
+-- Setup vshard.
+_G.vshard = vshard
+box.once('guest', function()
+    box.schema.user.grant('guest', 'super')
+end)
+local uri = 'guest@localhost:3301'
+local cfg = {
+    bucket_count = 3000,
+    sharding = {
+        [box.info().cluster.uuid] = {
+            replicas = {
+                [box.info().uuid] = {
+                    uri = uri,
+                    name = 'storage',
+                    master = true,
+                },
+            },
+        },
+    },
+}
+vshard.storage.cfg(cfg, box.info().uuid)
+vshard.router.cfg(cfg)
+vshard.router.bootstrap()
+
+-- Create the 'customers' space.
+box.once('customers', function()
+    box.schema.create_space('customers', {
+        format = {
+            {name = 'id', type = 'unsigned'},
+            {name = 'bucket_id', type = 'unsigned'},
+            {name = 'name', type = 'string'},
+            {name = 'age', type = 'number'},
+        }
+    })
+    box.space.customers:create_index('primary_index', {
+        parts = {
+            {field = 1, type = 'unsigned'},
+        },
+    })
+    box.space.customers:create_index('bucket_id', {
+        parts = {
+            {field = 2, type = 'unsigned'},
+        },
+        unique = false,
+    })
+    box.space.customers:create_index('age', {
+        parts = {
+            {field = 4, type = 'number'},
+        },
+        unique = false,
+    })
+
+    -- Fill the space.
+    box.space.customers:insert({1, 477, 'Elizabeth', 12})
+    box.space.customers:insert({2, 401, 'Mary', 46})
+    box.space.customers:insert({3, 2804, 'David', 33})
+    box.space.customers:insert({4, 1161, 'William', 81})
+    box.space.customers:insert({5, 1172, 'Jack', 35})
+    box.space.customers:insert({6, 1064, 'William', 25})
+    box.space.customers:insert({7, 693, 'Elizabeth', 18})
+end)
+
+-- Create the developers space.
+box.once('developers', function()
+    box.schema.create_space('developers', {
+        format = {
+            {name = 'id', type = 'unsigned'},
+            {name = 'bucket_id', type = 'unsigned'},
+            {name = 'name', type = 'string'},
+            {name = 'surname', type = 'string'},
+            {name = 'age', type = 'number'},
+        }
+    })
+    box.space.developers:create_index('primary_index', {
+        parts = {
+            {field = 1, type = 'unsigned'},
+        },
+    })
+    box.space.developers:create_index('bucket_id', {
+        parts = {
+            {field = 2, type = 'unsigned'},
+        },
+        unique = false,
+    })
+    box.space.developers:create_index('age_index', {
+        parts = {
+            {field = 5, type = 'number'},
+        },
+        unique = false,
+    })
+    box.space.developers:create_index('full_name', {
+        parts = {
+            {field = 3, type = 'string'},
+            {field = 4, type = 'string'},
+        },
+        unique = false,
+    })
+
+    -- Fill the space.
+    box.space.developers:insert({1, 477, 'Alexey', 'Adams', 20})
+    box.space.developers:insert({2, 401, 'Sergey', 'Allred', 21})
+    box.space.developers:insert({3, 2804, 'Pavel', 'Adams', 27})
+    box.space.developers:insert({4, 1161, 'Mikhail', 'Liston', 51})
+    box.space.developers:insert({5, 1172, 'Dmitry', 'Jacobi', 16})
+    box.space.developers:insert({6, 1064, 'Alexey', 'Sidorov', 31})
+end)
+
+-- Initialize crud.
+crud.init_storage()
+crud.init_router()
+
+-- Start a console.
+console.start()
+os.exit()
--- a/test/doc/playground_test.lua
+++ b/test/doc/playground_test.lua
@@ -1,0 +1,133 @@
+local yaml = require('yaml')
+local t = require('luatest')
+local g = t.group()
+
+local popen_ok, popen = pcall(require, 'popen')
+
+g.before_all(function()
+    t.skip_if(not popen_ok, 'no built-in popen module')
+    t.skip_if(jit.os == 'OSX', 'popen is broken on Mac OS: ' ..
+        'https://github.com/tarantool/tarantool/issues/6674')
+end)
+
+-- Run ./doc/playground.lua, execute a request and compare the
+-- output with reference return values.
+--
+-- The first arguments is the request string. All the following
+-- arguments are expected return values (as Lua values).
+--
+-- The function ignores trailing `null` values in the YAML
+-- output.
+local function check_request(request, ...)
+    local ph, err = popen.new({'./doc/playground.lua'}, {
+        stdin = popen.opts.PIPE,
+        stdout = popen.opts.PIPE,
+        stderr = popen.opts.DEVNULL,
+    })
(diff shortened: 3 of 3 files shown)


Totktonada:
Created for myself to experiment around crud, but decided to share for comments. If it'll look useful for somebody, I'll merge it. If not, I'll close the pull request.

doc/playground.lua
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/tarantool
```
oleg-jukovec, line 1: It will make your script more portable:
```suggestion
#!/usr/bin/env tarantool
```

doc/playground.lua
```diff
@@ -0,0 +1,165 @@
...
+
+-- How to run:
+--
+-- $ ./doc/playground.lua
```
oleg-jukovec, line 5: That should be enough:
```suggestion
-- $ doc/playground.lua
```

doc/playground.lua
Totktonada: Ouch, yep, I usually do, but forgot now. Fixed.

doc/playground.lua
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env tarantool
+
+-- How to run:
```
oleg-jukovec, line 3: We need a 'How-to run samples' section in README.md

doc/playground.lua
```diff
@@ -0,0 +1,165 @@
...
+local crud = require('crud')
+local console = require('console')
+
+-- Trick to don't leave *.snap, *.xlog files. See
+-- test/tuple_keydef.test.lua in the tuple-keydef module.
+if os.getenv('KEEP_DATA') ~= nil then
+    box.cfg()
+else
+    local tempdir = fio.tempdir()
+    box.cfg({
```
oleg-jukovec, line 32: I'm not clear with https://www.tarantool.io/en/doc/latest/dev_guide/lua_style_guide/#idioms-and-patterns

Is box.cfg{} necessary or box.cfg({}) is acceptable?

doc/playground.lua
```diff
@@ -0,0 +1,165 @@
...
+    if next_example - 2 < 1 then
+        print("You're on the first example")
+        return
+    end
+
+    print(examples[next_example - 2])
+    next_example = next_example - 1
+end
+
+add_help('doc/select.md')
```
oleg-jukovec, line 156: We need some test of the script on CI/CD. It's not a fact that developers will use it for each issue and the script can easily become outdated.

oleg-jukovec (changes requested):
Thank you for the pull request! It seems to me that this is a good idea. Now from README.md it is not clear how to manually test the crud. For example, I didn't understand how to run it manyally and tested it only with the luatest.

Manual testing will be very easy with this script.

I haven't looked at the code very carefully because it is a proposal. So I made my opinion and I noted only the obvious things:

1. Need a section in README.md as entry point.
2. Need autotests.

doc/playground.lua
Totktonada: I have to consider both variants as acceptable. To be honest, if I would write this style guide now, I would not recommend `func{<...>}` form for any function, because this way it is not perfectly clear what function should be called this way and what shouldn't.

doc/playground.lua
Totktonada: Are you have some tests in the mind?

I can propose a smoke test: console is run, a simple select works, `help()` and `prev()` show something (w/o a check what exactly).

doc/playground.lua
Totktonada: Hm. It would not work if it would be just `playground.lua` in the current directory (without `.` in `PATH`, which is unusual). Why `doc/playground.lua` works? It is a bit surprising for me.

doc/playground.lua
Totktonada: And add `customers` space too, to make examples from README.md workable. And add README.md examples into `help()`. Okay.

doc/playground.lua
oleg-jukovec: ```./``` means 'search the file in current directory (not in PATH)', not 'execute'.  Try to use absolute path with ```./``` instead of relative path.

I think this is due to the history of shell development. This is not critical and you can not change the line.

doc/playground.lua
oleg-jukovec: Thank you!

doc/playground.lua
oleg-jukovec: It seems to me that some kind of unit testing will be over-engineering here, so we need to test an output.

If we can split Tarantool's log output and the script output then we can compare the output with a reference output (a ready file). It can be something like that:

- Some kind of 'help()/prev()/next()' sequence as input. Output is saved to a file.
- Compare the output file and the reference file.

At first glance, they should always be same. The examples themselves do not change often, so it should not be a big problem to update a reference output time to time.

Totktonada: I would test the navigation on some testing samples. Duplicating of the documentation feels a bit too heavy here for me.

Totktonada:
Still in the draft state, but updated a bit.

Changes:

* Removed `help()` / `prev()`. It does not feel very convenient for me. I would rather show a list of sections / examples to read, enumerate them and let a user to choose what to look. Commands would look like `help list`, `help show 1`. I'll not dig into this direction within this pull request. Let's look into docs in a separate window. But I opened PR #291 to ease navigation over the documentation. BTW, we can expand the idea with `help list` / `help show` to make the examples testable.
* Added `customers` space. Its format and content more or less corresponds to what we can see in `README.md`. However there is no consistency in the readme itself: different examples assume different data (and even different bucket_count value). I decided to don't polish the readme in this pull request.
* Changed bucket_id values in the `developers` space to `bucket_id_strcrc32()` values calculated using primary keys. I would update them in `doc/select.md` and `doc/pairs.md` later. See https://github.com/tarantool/crud/pull/109#issuecomment-1119143504.
* Added `bucket_id` indexes. It does not change anything for playground, but a real sharded space should have this index to allow vshard rebalancer doing its work.
* Created a PoC patch to fix annoying prompt/log mixing: https://github.com/tarantool/tarantool/pull/7170.

TODO list:

* Mention in README.md.
* Add several simple tests: run playground, do some crud requests, check output.

doc/playground.lua
Totktonada: I would prefer to keep it. I feel myself a bit more safe with the more explicit syntax.

doc/playground.lua
Totktonada: I have added the `customers` space, but don't add a section into readme. I'll do after PR #291.

doc/playground.lua
Totktonada: `help()` / `prev()` part is not actual anymore, but the idea to test that the script is workeable still actual.

doc/playground.lua
Totktonada: Added 'Quickstart' section into readme.

doc/playground.lua
Totktonada: Added a test.

test/doc/playground_test.lua
```diff
@@ -59,56 +59,70 @@ local function check_request(request, ...)
...
 end
 
-g.test_select_customers = function()
-    check_request("crud.select('customers', {{'<=', 'age', 35}})", {
-        metadata = {
-            {name = 'id', type = 'unsigned'},
-            {name = 'bucket_id', type = 'unsigned'},
-            {name = 'name', type = 'string'},
-            {name = 'age', type = 'number'},
+local cases = {
```
Totktonada, line 62: Don't know, which variant is better: define cases as functions explicitly or do it in a declarative way. Kept both variants to collect a feedback.

Totktonada:
Changes:

* Added 'Quickstart' section into README.md.
* Added a test.

It is ready for review now, but I need to choose between two variants of the test.

README.md
```diff
@@ -37,6 +38,29 @@ It also provides the `crud-storage` and `crud-router` roles for
...
+
+Now you have the following options how to learn crud API and use it in a
+project:
+
+* Play with crud on a testing dataset on a single instance:
+
+  ```shell
+  $ git clone https://github.com/tarantool/crud.git
+  $ cd crud
+  $ tarantoolctl rocks install vshard
```
oleg-jukovec, line 53: $ tarantoolctl rocks install vshard
$ doc/playground.lua 
LuajitError: /home/ozi/tmp/crud/crud/cfg.lua:5: module 'checks' not found:
	no field package.preload['checks']
	no file '/home/ozi/tmp/crud/checks.lua'
	no file '/home/ozi/tmp/crud/checks/init.lua'
	no file '/home/ozi/tmp/crud/checks.so'
	no file '/home/ozi/tmp/crud/.rocks/share/tarantool/checks.lua'
	no file '/home/ozi/tmp/crud/.rocks/share/tarantool/checks/init.lua'
	no file '/home/ozi/tmp/.rocks/share/tarantool/checks.lua'
	no file '/home/ozi/tmp/.rocks/share/tarantool/checks/init.lua'
	no file '/home/ozi/.rocks/sha
fatal error, exiting the event loop
```suggestion
  $ ./deps.sh
```

README.md
```diff
@@ -37,6 +38,29 @@ It also provides the `crud-storage` and `crud-router` roles for
...
+  $ ./doc/playground.lua
+  tarantool> crud.select('customers', {{'<=', 'age', 35}})
+  tarantool> crud.select('developers', nil, {first = 6})
+  ```
+* Add crud into dependencies of a Cartridge application and add crud roles into
+  dependencies of your roles (see [Cartridge roles](#cartridge-roles) section).
+* Add crud into dependencies of your application (rockspec, RPM spec -- depends
+  of your choice) and call crud initialization code from storage and router
+  code (see [API](#api) section).
+
```
oleg-jukovec, line 63: We have a CMakeLists.txt file. Do we need to explain what can be done with it here? Or is it better to do it in a separate issue?

doc/playground.lua
```diff
@@ -0,0 +1,145 @@
...
+    })
+    box.space.customers:create_index('bucket_id', {
+        parts = {
+            {field = 2, type = 'unsigned'},
+        },
+    })
+    box.space.customers:create_index('age', {
+        parts = {
+            {field = 4, type = 'number'},
+        },
```
oleg-jukovec, line 84: We can add one non-unique index, I needed it a couple of times (integration tests have it).
```suggestion
        parts = {
            {field = 4, type = 'number'},
            unique = false,
        },
```

doc/playground.lua
```diff
@@ -0,0 +1,145 @@
...
+    })
+    box.space.developers:create_index('bucket_id', {
+        parts = {
+            {field = 2, type = 'unsigned'},
+        },
+    })
+    box.space.developers:create_index('age_index', {
+        parts = {
+            {field = 5, type = 'number'},
+        },
```
oleg-jukovec, line 121: And another one non-unique index here:
```suggestion
        parts = {
            {field = 5, type = 'number'},
            unique = false,
        },
```

test/doc/playground_test.lua
oleg-jukovec: I like the last one.

test/doc/playground_test.lua
```diff
@@ -0,0 +1,128 @@
...
+                {2, 401, 'Sergey', 'Allred', 21},
+                {3, 2804, 'Pavel', 'Adams', 27},
+                {4, 1161, 'Mikhail', 'Liston', 51},
+                {5, 1172, 'Dmitry', 'Jacobi', 16},
+                {6, 1064, 'Alexey', 'Sidorov', 31},
+            },
+        },
+    },
+    {
+        'test_insert',
```
oleg-jukovec, line 104: Names can be taken from keys, but it's really not critical:
```Lua
test_insert = {
    ...
},
```
[see an example](https://github.com/tarantool/crud/blob/20023a9a6faab1976997534e11c3b71f24d9061f/test/unit/stats_test.lua#L93-L209).

oleg-jukovec (approved):
Thanks for the update, I made a few notes, but they are not critical and can be easily corrected. So LGTM after resolving the notes.

test/doc/playground_test.lua
Totktonada: Kept it, squashed.

test/doc/playground_test.lua
Totktonada: It was my variant of declarative test case description, which I use, when the order was important. Here it is not the case, so I'll rewrite it as you suggested.

README.md
Totktonada: `deps.sh` installs everything that is needed for development. Here I had idea to install only what is needed for using crud. But, yep, I missed that we need checks and errors.

`tarantoolctl rocks make` here would be more suitable, I think.

README.md
Totktonada: In those two bullets: I think, no. A user either add a rock as the dependency (so tarantoolctl rocks will do the job) or add a package as the dependency (so the package will install everything appropriately).

I'll add `tarantoolctl rocks make` to the first bullet.

README.md
Totktonada: Replaced `tarantoolctl rocks install vshard` with `tarantoolctl rocks make`.

README.md
Totktonada: Replaced `tarantoolctl rocks install vshard` with `tarantoolctl rocks make`.

doc/playground.lua
Totktonada: You're right. Nice catch!

Bucket IDs, ages, names should be in a non-unique index. I fixed it.

(`unique` parameter is on the same level as `parts`, I have added it into the right place.)

doc/playground.lua
Totktonada: Fixed (see the comment above).

README.md
oleg-jukovec: It works, thank you!

doc/playground.lua
```diff
@@ -0,0 +1,150 @@
...
+    box.space.developers:insert({5, 1172, 'Dmitry', 'Jacobi', 16})
+    box.space.developers:insert({6, 1064, 'Alexey', 'Sidorov', 31})
+end)
+
+-- Initialize crud.
+crud.init_storage()
+crud.init_router()
+
+-- Start a console.
+console.start()
```
DifferentialOrange, line 149: image

Having explicit `null` as a second return val is interesting. Is this a side effect of `console`?

DifferentialOrange (approved):
LGTM after resolving remaining comments.

doc/playground.lua
Totktonada: It is from `sharding.wrap_select_method` (the same for insert). I spin on it for some time, but then decided that it is not a bug. The second return value is an error.

TBH, I would change it to make visually better in a console (not in this PR), but I'll not file an issue, because it is a thing of taste.

test/doc/playground_test.lua
```diff
@@ -0,0 +1,125 @@
...
+            rows = {
+                {100, 2976, 'Alfred', 'Hitchcock', 123},
+            },
+        }
+    },
+}
+
+for case_name, case in pairs(cases) do
+    g[case_name] = function()
+        -- We have no cases with more than one non-null return
```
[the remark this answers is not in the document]
Totktonada: Good idea, will do.

doc/playground.lua
```diff
@@ -0,0 +1,150 @@
...
+-- Or
+--
+-- $ KEEP_DATA=1 ./doc/playground.lua
+--
+-- What to do next:
+--
+-- Choose an example from README.md, doc/select.md or doc/pairs.md
+-- and run. For example:
+--
+-- tarantool> crud.select('customers', {{'<=', 'age', 35}})
```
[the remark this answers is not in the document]
Totktonada: I'll add `first`.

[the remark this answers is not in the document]
Totktonada: Done.

test/doc/playground_test.lua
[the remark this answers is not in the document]
Totktonada: Done. I return `nil` and `err.err` to get rid of volatile paths in backtraces.

PR #281, opened 2022-05-06
stats: make quantile tolerated error configurable

DifferentialOrange:
Make metrics quantile collector tolerated error [1] configurable. Change
metrics quantile collector default tolerated error from 1e-2 to 1e-3.

The motivation of this patch is a tarantool/metrics bug [2]. Sometimes
quantile values turn to `-Inf` under high load when observations are
small. It was reproduced in process of developing Grafana dashboard
panels for CRUD stats [3].

Quantile tolerated error could be changed with crud.cfg:
```lua
  crud.cfg{stats_quantile_tolerated_error = 1e-4}
```

1. https://www.tarantool.io/ru/doc/latest/book/monitoring/api_reference/#summary
2. https://github.com/tarantool/metrics/issues/189
3. https://github.com/tarantool/grafana-dashboard/tree/DifferentialOrange/crud-report

- [x] Tests
- [x] Changelog
- [x] Documentation

Issue #278, opened 2022-05-05
CRUD 0.11.0 incompatible with old-style apps when bucket_id is defined explicitly


no1seman:
We have many projects with old-style manner when bucket_id is defined explicitly and in some cases crud 0.11.0 fails with an error.

For example, if crud.upsert_object used got the folowing error:

UpsertError: Failed to call upsert on storage-side: ShardingHashMismatchError: crud: Sharding hash mismatch for space cd_flight. Sharding info will be refreshed after receiving this error. Please retry your request.

On crud 0.10.0 - all works fine.

PR #282, opened 2022-05-06
Release 0.11.1

DifferentialOrange:
### Overview

This release fixes critical bug introduced in 0.11.0. It is
recommended to use 0.11.1 instead of 0.11.0.

It also adds an ability to configure quantile collector tolerated
error and changes its default value.

### Breaking changes

There are no breaking changes in the release.

### New features

Set quantile collector tolerated error with crud.cfg (#281):
```
crud.cfg{ stats_quantile_tolerated_error = 1e-4 }
```
Decreasing tolerated error may fix getting `-Inf` values in
quantiles. Increasing tolerated error may improve performance a bit.

### Bugfixes

Requests no more fail with "Sharding hash mismatch" error
if ddl set and bucket_id is explicitly specified (#278).

- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+
+### Changed
+
+### Fixed
+
+## [0.11.1] - 06-05-22
+
+### Added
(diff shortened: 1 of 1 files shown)


Totktonada (approved):
LGTM.

Please, add the caution to 0.11.0 release on GitHub too.

PR #288, opened 2022-05-17
Fix nan crud.stats latency

### LINKED ISSUE #286: stats: select ok latency becomes -nan after 30 seconds
Imagine we have a Cartridge cluster with some data ... if we make the following operations in router console:

```log
unix/:./tmp/run/app-cartridge.router.control> do
                                            >     fiber=require('fiber')
                                            >     crud.cfg({stats = true})
                                            >     t={}
                                            > 
                                            >     res=crud.select('space')
                                            >     table.insert(t, crud.stats().spaces.space.select.ok.latency)
                                            >     fiber.sleep(30);
                                            >     table.insert(t, crud.stats().spaces.space.select.ok.latency)
                                            > end
---
...

unix/:./tmp/run/app-cartridge.router.control> t
---
- - 0.0036880000006931
  - -nan
...
```

Seems quite strange that "select" "ok" "latency" becomes -nan after 30 seconds

DifferentialOrange:
### stats: make quantile age params configurable

The main motivation of making these parameters configurable is to be
able to write time-adequate test cases for #286 issue, but they may be
useful in getting rid of #286 effect at all or in quantile time window
calibration.

After this patch, statistics summary quantile aging params
`age_bucket_count` and `max_age_time` [1] could be configured:
```lua
crud.cfg{
    stats_quantile_age_bucket_count = 3,
    stats_quantile_max_age_time = 30,
}
```

Only type validation is conducted in crud.cfg, every other validation is
performed by metrics itself.

1. https://www.tarantool.io/ru/doc/latest/book/monitoring/api_reference/#summary

### stats: separate quantile and average fields

Add separate quantile (`latency_quantile_recent`) and average (
`latency_average`) fields to `crud.stats()` output. `latency` field and
tarantool/metrics output remains unchanged.

Before this patch, `latency` displayed `latency_quantile_recent` or
`latency_average` and there wasn't any was to see pre-computed average
if quantiles are enabled. But it may be useful if quantile is `nan`.

Quantiles may display `-nan` if there were no observations for a several
ages. Such behavior is expected [1] and valid: for example, Grafana
should ignore such values and they will be displayed as `No data` for
a window when there wasn't any requests.

1. https://github.com/tarantool/metrics/issues/303

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* Make metrics quantile collector age params configurable (#286).
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 * Make metrics quantile collector age params configurable (#286).
+* Add separate `latency_average` and `latency_quantile_recent`
+  fields to `crud.stats()` output (#286).
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -721,11 +721,15 @@ crud.stats()
     my_space:
       insert:
         ok:
-          latency: 0.002
+          latency: 0.0015
+          latency_average: 0.002
+          latency_quantile_recent: 0.0015
           count: 19800
           time: 39.6
         error:
-          latency: 0.000001
+          latency: 0.0000008
+          latency_average: 0.000001
+          latency_quantile_recent: 0.0000008
           count: 4
           time: 0.000004
 ...
@@ -733,11 +737,15 @@ crud.stats('my_space')
 ---
 - insert:
     ok:
-      latency: 0.002
+      latency: 0.0015
+      latency_average: 0.002
+      latency_quantile_recent: 0.0015
       count: 19800
       time: 39.6
     error:
-      latency: 0.000001
+      latency: 0.0000008
+      latency_average: 0.000001
+      latency_quantile_recent: 0.0000008
       count: 4
       time: 0.000004
 ...
@@ -759,10 +767,17 @@ and `borders` (for `min` and `max` calls).
 Each operation section consists of different collectors
 for success calls and error (both error throw and `nil, err`)
 returns. `count` is the total requests count since instance start
-or stats restart. `latency` is the 0.99 quantile of request execution
-time if `metrics` driver used and quantiles enabled,
-otherwise `latency` is the total average.
-`time` is the total time of requests execution.
+or stats restart.  `time` is the total time of requests execution.
+`latency_average` is `time` / `count`.
+`latency_quantile_recent` is the 0.99 quantile of request execution
+time for a recent period (see 
+[`metrics` summary API](https://www.tarantool.io/ru/doc/latest/book/monitoring/api_reference/#summary)).
+It is computed only if `metrics` driver is used and quantiles are
+enabled. `latency_quantile_recent` value may be `-nan` if there
+wasn't any observations for several ages, see
+[tarantool/metrics#303](https://github.com/tarantool/metrics/issues/303).
+`latency` is a `latency_quantile_recent` if `metrics` driver is used
+and quantiles are enabled, otherwise it's `latency_average`.
 
 In [`metrics`](https://www.tarantool.io/en/doc/latest/book/monitoring/)
 registry statistics are stored as `tnt_crud_stats` metrics
--- a/README.md
+++ b/README.md
@@ -797,7 +797,17 @@ crud.cfg{stats_quantile_tolerated_error = 1e-4}
 ```
 See [tarantool/metrics#189](https://github.com/tarantool/metrics/issues/189) for
 details about the issue.
-
+You can also configure quantile `age_bucket_count` (default: 2) and
+`max_age_time` (in seconds, default: 60):
+```lua
+crud.cfg{
+    stats_quantile_age_bucket_count = 3,
+    stats_quantile_max_age_time = 30,
+}
+```
+See [`metrics` summary API](https://www.tarantool.io/ru/doc/latest/book/monitoring/api_reference/#summary)
+for details. These parameters can be used to smooth time window move
+or reduce the amount on `-nan` gaps for low request frequency applications.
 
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -8,8 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* Make metrics quantile collector age params configurable (#286).
 
 ### Changed
+* Display total average in `crud.stats` if quantile is `-nan` (#286).
```
no1seman, line 14: Cool, but may be it's better to add average as additional field of stats and leave latency as before? This will make possible to see overall average and also quantile in a period of time of 'max_age_time'

README.md
```diff
@@ -797,6 +797,16 @@ crud.cfg{stats_quantile_tolerated_error = 1e-4}
 ```
 See [tarantool/metrics#189](https://github.com/tarantool/metrics/issues/189) for
 details about the issue.
+You can also configure quantile `age_bucket_count` and `max_age_time`:
+```lua
+crud.cfg{
+    stats_quantile_age_bucket_count = 3,
+    stats_quantile_max_age_time = 30,
+}
```
Totktonada, line 808: What are default values?

crud/stats/metrics_registry.lua
```diff
@@ -193,7 +198,8 @@ end
 --
 --  If quantiles disabled, we need to compute
 --  latency as overall average from `time` and
---  `count` values.
+--  `count` values. We also need to compute
+--  latencies as average is quantiles is nan.
```
Totktonada, line 202: Typo: is -> if.

CHANGELOG.md
```diff
@@ -8,8 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
```
Totktonada, line 8: Please, consider CI results. It fails on jobs with metrics.

README.md
DifferentialOrange: https://github.com/tarantool/crud/blob/aca29f3bfb092f32e2996716d26a529c5b00d4c7/crud/stats/init.lua#L534-L538

Totktonada: I mean, I don't see it in the readme.

CHANGELOG.md
DifferentialOrange: For now there are three possible approaches.

### Fill `-nan` with average
(implemented now)

#### Pros
- API remains the same, backward compatible (in all adequate cases)
- The logic behind the values changes slightly

#### Cons
- No possible way to distinct what is displayed in `latency` now
- Whole `latency` logic is really complicated and hard to understand

### Add `average` field, do not change `latency`

#### Pros
- Backward compatible
- When quantiles enabled, possible to see both average and quantile latency

#### Cons
- The logic is still confusing since in many cases `average` = `latency` (local driver or quantiles disabled)


### Add `average` field, display only quantile value in `latency`

#### Pros
- Logic is rather clear and simple to understand
- When quantiles enabled, possible to see both average and quantile latency

#### Cons
- Breaking change

Options 2 and 3 really sounds more satisfying than 1. I think we'll discuss it a bit and choose a solution.

DifferentialOrange: One more possible approach

### Add separate `average` and `quantile` fields, deprecate (but not remove yet) `latency` field

It should clear up the logic behind the values and smooth up `latency` breaking change.

crud/stats/metrics_registry.lua
DifferentialOrange: Fixed, thanks!

README.md
DifferentialOrange: Added to README

CHANGELOG.md
Totktonada: Not relevant anymore.

README.md
```diff
@@ -759,10 +767,17 @@ and `borders` (for `min` and `max` calls).
...
+`latency_average` is `time` / `count`.
+`latency_quantile_recent` is the 0.99 quantile of request execution
+time for a recent period (see 
+[`metrics` summary API](https://www.tarantool.io/ru/doc/latest/book/monitoring/api_reference/#summary)).
+It is computed only if `metrics` driver used and quantiles enabled.
+`latency_quantile_recent` value may be `-nan` if there wasn't any
+observations for several ages, see
+[tarantool/metrics#303](https://github.com/tarantool/metrics/issues/303).
+`latency` is a `latency_quantile_recent`, if it presents,
+otherwise it's `latency_average`.
```
Totktonada, line 780: I can understand it as composition of the following three conditions:

1. Metrics is enabled.
2. Quantiles are enabled.
3. `latency_quantile_recent` is `-nan` at the moment (it means that no data is present, right?)

However, as I remember from our private discussion, you had idea to implement it based on the first two conditions.

It requires a clarification in the documentation to avoid misinterpretation.

Totktonada (approved):

README.md
```diff
@@ -721,23 +721,31 @@ crud.stats()
     my_space:
       insert:
         ok:
-          latency: 0.002
+          latency: 0.0015
+          latency_average: 0.002
+          latency_quantile_recent: 0.0015
```
Totktonada, line 726: I'll not insist here, but, as for me, `latency_recent` is descriptive enough.

README.md
DifferentialOrange: It's only the first two, yes. If `latency_quantile_recent` is `-nan`, `latency` is `-nan` too.

README.md
DifferentialOrange: I think it's better to leave it since "quantile" is what this value is.

README.md
DifferentialOrange: Replaced "if it presents" with "if the field in not empty". I'm not sure we should explicitly write about `-nan` here.

Totktonada: But why not 'if metrics and quantiles are enabled'?

CHANGELOG.md
Totktonada: It is implemented as suggested by Yaroslav, so I'll resolve the discussion.

README.md
DifferentialOrange: I just thought it will be tautologic. But your version is better, so I reworked it.

Totktonada: I think that the latest variant is more explicit. Thanks for the update.

PR #290, opened 2022-05-18
test: fix timeout error message for 2.10.0-rc1

Totktonada:
Otherwise the test cases fail on tarantool 2.10.0-beta2-165-gc463e56b9
and newer.

See https://github.com/tarantool/tarantool/pull/6538.

### DIFF
--- a/test/unit/call_test.lua
+++ b/test/unit/call_test.lua
@@ -141,7 +141,7 @@ g.test_timeout = function()
 
     t.assert_equals(results, nil)
     t.assert_str_contains(err.err, "Failed for %w+%-0000%-0000%-0000%-000000000000", true)
-    t.assert_str_contains(err.err, "Timeout exceeded")
+    helpers.assert_timeout_error(err.err)
 end
 
 local function check_single_vshard_call(g, exp_vshard_call, opts)
@@ -272,5 +272,5 @@ g.test_any_vshard_call_timeout = function()
 
     t.assert_equals(results, nil)
     t.assert_str_contains(err.err, "Failed for %w+%-0000%-0000%-0000%-000000000000", true)
-    t.assert_str_contains(err.err, "Timeout exceeded")
+    helpers.assert_timeout_error(err.err)
 end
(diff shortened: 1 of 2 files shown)


test/unit/call_test.lua
```diff
@@ -141,7 +141,10 @@ g.test_timeout = function()
 
     t.assert_equals(results, nil)
     t.assert_str_contains(err.err, "Failed for %w+%-0000%-0000%-0000%-000000000000", true)
-    t.assert_str_contains(err.err, "Timeout exceeded")
+    -- 'Timeout exceeded' or 'timed out'.
+    --
+    -- See https://github.com/tarantool/tarantool/pull/6538.
+    t.assert_str_contains(err.err, "[Tt]imed? ?out", true)
```
olegrok, line 147: Alternatively you can implement special helper for that.

To be consistent with vshard (https://github.com/tarantool/vshard/commit/f4e57ba2285ec4dfbebaccfc882390beb6c77764#diff-d7277fd5f081018a7522656bb0d6fa5b34b670ed5100ffc99e18373720e6e570R206) and cartridge (https://github.com/tarantool/cartridge/commit/73686c1cfd6a71c776ff2ebac125d295a9a3cde4#diff-aed3b7f006a0175c125574a5621fba5978870e2247cb6d719836f1386718062bR203-R207)

DifferentialOrange (approved):
LGTM after resolving (or ignoring) olegrok comment.

test/unit/call_test.lua
Totktonada: It looks more accurate in the code, but will not show the actual error message value in case of an error. Can we achieve both, what do you think?

olegrok: You can move this assert to separate helper. It will show error message and reduce amount of copy-paste in case if there will be more tests with timeout error check.

Also feel free to ignore this message, it's not so serious issue and current patch is also OK for me :)

Totktonada: I wrote a simple helper.

olegrok (approved):

PR #291, opened 2022-05-21
Tiny documentation enhancements

Totktonada:
- Added forgotten `crud-router` mention.
- Added tables of contents (using [doctoc](https://github.com/thlorenz/doctoc)).
- Added Vim Swap files into .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ luacov.report.out
 luacov.stats.out
 build/*.cmake
 build/Makefile
+
+# Vim Swap files.
+.*.s[a-w][a-z]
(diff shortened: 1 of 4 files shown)


olegrok (approved):

.gitignore
```diff
@@ -10,3 +10,6 @@ luacov.report.out
 luacov.stats.out
 build/*.cmake
 build/Makefile
+
+# Vim Swap files.
+.*.sw[a-z]
```
oleg-jukovec, line 15: http://vimdoc.sourceforge.net/htmldoc/recover.html#swap-file
```
If the ".swp" file name already exists, the last character is
	   decremented until there is no file with that name or ".saa" is
	   reached.  In the last [case](http://vimdoc.sourceforge.net/htmldoc/change.html#case), no swap file is created.
```
```suggestion
.*.s[a-w][a-z]
```

oleg-jukovec (approved):
Thanks for the pull request. I like it, but you can improve the regular expression for vim swap files if you want.

.gitignore
Totktonada: No problem, updated.

DifferentialOrange (approved):
I think two ToC commits could be merged into one, but it's a matter of taste

Issue #292, opened 2022-05-23
design document: how the query planner/executor work


Totktonada:
* Basic planner cases and executor algorithm.
* Meaning of the query plan fields.
* What the executor will do, when both 'scan value' and 'after scan value' are present? I guess, it'll use 'after scan value' with GT/LT and transform 'scan value' to a post-filter, but I'm not completely sure.
* Whether the 'after scan value' always corresponds to the index determined by the first select condition? What if the first condition is not backed by any index? Will the 'after' condition be transformed to a post-filter (rather than GT/LT on an index) in the case?

Initially asked in https://github.com/tarantool/crud/pull/277#discussion_r878778465

oleg-jukovec:
I left a note here, but I will add the documentation later.

The algorithm is not too complicated:

1. An index is selected according to a first condition with suitable fields. By default it is primary index (if there are no conditions with an index fields).
2. A traverse order is selected according to the condition (if it exists for the index). By default is ascending (GE, GT, EQ, ALL -> ASC, LE, LT, REQ -> DESC). At here we have a transformation EQ/REQ to GE/LE if `after` presents.
2.1 if `first` is negative value and `after` exists then we add a condition with `after` value as first one for the index and the traverse order.
3. A traverse start key (`scan value`) is selected as greater between the condition (if it exists for the index) and `after` according to the traverse order. By default is `nil` (if there are no suitable conditions for the index and no `after`).
4. Start a traverse from the key (`scan value`).
5. Skip some tuples before `after` value (based on secondary + primary key fields) while tuples < `after`.
6. Fetch tuples + filtering by conditions (except the first one).

So it three base cases:

1. If we have a condition (GT, GE, ALL, LT, LE) for an index + `after` then traverse starts from a greater one (according to traverse order). Other conditions will be used as post-filter.
2. If we have a EQ/REQ condition for an index + `after` then it will be full scan from the `after` value, EQ/REQ will be added to other conditions as post-filter.
3. If we have only  non-index condition + `after`, then a primary index with an ascending order will be used. The non-index conditions will be used as post-filter.

So for the issue #276 and the pull requres #277, we can consider `after` as a full scan in all cases without `first` <= 1000.

PR #295, opened 2022-05-24
Fix crud.select() + after behavior problems

oleg-jukovec:
It seems to me that now there are logical errors with `select` + `after`:

1. Always a smaller key is choosed to start the `select` when has_keydef.
2. Unsymmetric behavior on ascending and descending order with ER/REQ. 
3. `select(nil)` + `after` always traverse to the `after` from beginning.

The patch fixes the problems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
     - [6, 1064, 'Alexey', 'Sidorov', 31]
   ...
 
+* `crud.select()` results order with negative `first`.
+
 ## [0.11.2] - 23-05-22
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 ### Changed
+* Optimize `crud.select()` without conditions and with `after`.
 
 ### Fixed
 * `crud.select()` if a condition is '<=' and it's value < `after`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,27 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
     - [1, 477, 'Alexey', 'Adams', 20]
   ...
 
+* `crud.select()` filtration by a first condition if the condition is '>'
+  or '>=' and it's value > `after`.
+
+  Before this patch:
+
+  tarantool> crud.select('developers', {{'>=', 'id', 5}}, {first = 10, after = rows[2]}).rows
+  ---
+  - - [3, 2804, 'Pavel', 'Adams', 27]
+    - [4, 1161, 'Mikhail', 'Liston', 51]
+    - [5, 1172, 'Dmitry', 'Jacobi', 16]
+    - [6, 1064, 'Alexey', 'Sidorov', 31]
+  ...
+
+  After this patch:
+
+  tarantool> crud.select('developers', {{'>=', 'id', 5}}, {first = 10, after = rows[2]}).rows
+  ---
+    - [5, 1172, 'Dmitry', 'Jacobi', 16]
+    - [6, 1064, 'Alexey', 'Sidorov', 31]
+  ...
+
 ## [0.11.2] - 23-05-22
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,40 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   ...
 
 * `crud.select()` results order with negative `first`.
+* `crud.select()` if a condition is '=' or '==' with negative `first`.
+
+  Suppose we have a non-unique secondary index by the field `age` field and a
+  space:
+
+  tarantool> crud.select('developers', nil, {first = 10})
+  ---
+  - metadata: [{'name': 'id', 'type': 'unsigned'}, {'name': 'bucket_id', 'type': 'unsigned'},
+      {'name': 'name', 'type': 'string'}, {'name': 'surname', 'type': 'string'}, {'name': 'age',
+        'type': 'number'}]
+    rows:
+    - [1, 477, 'Alexey', 'Adams', 20]
+    - [2, 401, 'Sergey', 'Allred', 27]
+    - [3, 2804, 'Pavel', 'Adams', 27]
+    - [4, 1161, 'Mikhail', 'Liston', 27]
+    - [5, 1172, 'Dmitry', 'Jacobi', 27]
+    - [6, 1064, 'Alexey', 'Sidorov', 31]
+  - null
+  ...
+
+  Before this patch:
+
+  tarantool> crud.select('developers', {{'=', 'age', 27}}, {first = -10, after = rows[4]}).rows
+  ---
+  - []
+  ...
+
+  After this patch:
+
+  tarantool> crud.select('developers', {{'=', 'age', 27}}, {first = -10, after = rows[4]}).rows
+  ---
+  - - [2, 401, 'Sergey', 'Allred', 27]
+    - [3, 2804, 'Pavel', 'Adams', 27]
+  ...
 
 ## [0.11.2] - 23-05-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,24 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
+* `crud.select()` if a condition is '<=' and it's value < `after`.
+
+  Before this patch:
+
+  tarantool> crud.select('developers', {{'<=', 'id', 3}}, {first = 10, after = rows[5]}).rows
+  ---
+  - - [2, 401, 'Sergey', 'Allred', 21]
+    - [1, 477, 'Alexey', 'Adams', 20]
+  ...
+
+  After this patch:
+
+  tarantool> crud.select('developers', {{'<=', 'id', 3}}, {first = 10, after = rows[5]}).rows
+  ---
+  - - [3, 2804, 'Pavel', 'Adams', 27]
+    - [2, 401, 'Sergey', 'Allred', 21]
+    - [1, 477, 'Alexey', 'Adams', 20]
+  ...
 
 ## [0.11.2] - 23-05-22
 
--- a/README.md
+++ b/README.md
@@ -440,8 +440,8 @@ where:
 * `conditions` (`?table`) - array of [select conditions](#select-conditions)
 * `opts`:
   * `first` (`?number`) - the maximum count of the objects to return.
-     If negative value is specified, the last objects are returned
-     (`after` option is required in this case).
+     If negative value is specified, the objects behind `after` are returned
+     (`after` option is required in this case). [See pagination examples](doc/select.md#pagination).
   * `after` (`?table`) - tuple after which objects should be selected
   * `batch_size` (`?number`) - number of tuples to process per one request to storage
   * `bucket_id` (`?number|cdata`) - bucket ID
--- a/crud/select/executor.lua
+++ b/crud/select/executor.lua
@@ -43,8 +43,11 @@ local generate_value
 
 if has_keydef then
     generate_value = function(after_tuple, scan_value, index_parts, tarantool_iter)
-        local cmp_operator = select_comparators.get_cmp_operator(tarantool_iter)
         local key_def = keydef_lib.new(index_parts)
+        if #scan_value == 0 and after_tuple ~= nil then
+            return key_def:extract_key(after_tuple)
+        end
+        local cmp_operator = select_comparators.get_cmp_operator(tarantool_iter)
         local cmp = key_def:compare_with_key(after_tuple, scan_value)
         if (cmp_operator == '<' and cmp < 0) or (cmp_operator == '>' and cmp > 0) then
             return key_def:extract_key(after_tuple)
@@ -52,9 +55,12 @@ if has_keydef then
     end
 else
     generate_value = function(after_tuple, scan_value, index_parts, tarantool_iter)
+        local after_tuple_key = utils.extract_key(after_tuple, index_parts)
+        if #scan_value == 0 and after_tuple ~= nil then
+            return after_tuple_key
+        end
         local cmp_operator = select_comparators.get_cmp_operator(tarantool_iter)
         local scan_comparator = select_comparators.gen_tuples_comparator(cmp_operator, index_parts)
-        local after_tuple_key = utils.extract_key(after_tuple, index_parts)
         if scan_comparator(after_tuple_key, scan_value) then
             return after_tuple_key
         end
--- a/crud/select/executor.lua
+++ b/crud/select/executor.lua
@@ -79,9 +79,41 @@ function executor.execute(space, index, filter_func, opts)
 
     local value = opts.scan_value
     if opts.after_tuple ~= nil then
-        local new_value = generate_value(opts.after_tuple, opts.scan_value, index.parts, opts.tarantool_iter)
-        if new_value ~= nil then
-            value = new_value
+        local iter = opts.tarantool_iter
+        if iter == box.index.EQ or iter == box.index.REQ then
+            -- we need to make sure that the keys are equal
+            -- the code is correct even if value is a partial key
+            local parts = {}
+            for i, _ in ipairs(value) do
+                -- the code required for tarantool 1.10.6 at least
+                table.insert(parts, index.parts[i])
+            end
+
+            local is_eq = iter == box.index.EQ
+            local is_after_bigger
+            if has_keydef then
+                local key_def = keydef_lib.new(parts)
+                local cmp = key_def:compare_with_key(opts.after_tuple, value)
+                is_after_bigger = (is_eq and cmp > 0) or (not is_eq and cmp < 0)
+            else
+                local comparator
+                if is_eq then
+                    comparator = select_comparators.gen_func('<=', parts)
+                else
+                    comparator = select_comparators.gen_func('>=', parts)
+                end
+                local after_key = utils.extract_key(opts.after_tuple, parts)
+                is_after_bigger = not comparator(after_key, value)
+            end
+            if is_after_bigger then
+                -- it makes no sence to continue
+                return resp
+            end
+        else
+            local new_value = generate_value(opts.after_tuple, value, index.parts, iter)
+            if new_value ~= nil then
+                value = new_value
+            end
         end
     end
 
--- a/crud/select/executor.lua
+++ b/crud/select/executor.lua
@@ -42,9 +42,11 @@ end
 local generate_value
 
 if has_keydef then
-    generate_value = function(after_tuple, scan_value, index_parts)
+    generate_value = function(after_tuple, scan_value, index_parts, tarantool_iter)
+        local cmp_operator = select_comparators.get_cmp_operator(tarantool_iter)
         local key_def = keydef_lib.new(index_parts)
-        if key_def:compare_with_key(after_tuple, scan_value) < 0 then
+        local cmp = key_def:compare_with_key(after_tuple, scan_value)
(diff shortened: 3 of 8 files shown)


test/integration/select_test.lua
```diff
@@ -217,32 +217,32 @@ pgroup.test_negative_first = function(g)
     t.assert_equals(objects, helpers.get_objects_by_idxs(customers, {2, 3, 4}))
 
     -- id >= 2
-    -- first -2 after 5
+    -- first -4 after 5
```
DifferentialOrange, line 220: I think it's better to leave a more human-readable version, like "last 4"

test/integration/select_test.lua
```diff
@@ -217,32 +217,32 @@ pgroup.test_negative_first = function(g)
...
-    local first = -2
+    local first = -4
     local after = crud_utils.flatten(customers[5], g.space_format)
     local result, err = g.cluster.main_server.net_box:call(
        'crud.select', {'customers', conditions, {first=first, after=after}})
 
     t.assert_equals(err, nil)
     local objects = crud.unflatten_rows(result.rows, result.metadata)
-    t.assert_equals(objects, helpers.get_objects_by_idxs(customers, {3, 4}))
+    t.assert_equals(objects, helpers.get_objects_by_idxs(customers, {2, 3, 4}))
```
DifferentialOrange, line 231: There are 7 tuples satisfying conditions: 2, 3, 4, 5, 6, 7, 8. Why 4 last tuples after 5 is 2, 3, 4 and not 6, 7, 8? I don't see any REQ iterators in the request options.

test/integration/select_test.lua
```diff
@@ -258,6 +258,20 @@ pgroup.test_negative_first = function(g)
     local objects = crud.unflatten_rows(result.rows, result.metadata)
     t.assert_equals(objects, helpers.get_objects_by_idxs(customers, {6}))
 
+    -- id <= 6
```
DifferentialOrange, line 261: But it's `id <= 2`

test/integration/select_test.lua
oleg-jukovec: after = customers[5]
first = -4
condition = age >= 2

So, it's traverse from customers[5] in backward order because first < 0 (1, 2, 3, 4) and filter results with id >= 2 (2, 3, 4).

test/integration/select_test.lua
oleg-jukovec: Yes, my mistake. There was another same problem, fixed.

test/integration/select_test.lua
oleg-jukovec: I don't see any value in these comments at all. This is just a duplication of the code below. I just copied the nearest code to make it look the same (you can see the same comments in the whole file).

test/integration/select_test.lua
DifferentialOrange: Description of after + negative first is fixed in separate commit

crud/select/executor.lua
```diff
@@ -45,16 +45,22 @@ if has_keydef then
     generate_value = function(after_tuple, scan_value, index_parts, tarantool_iter)
         local cmp_operator = select_comparators.get_cmp_operator(tarantool_iter)
         local key_def = keydef_lib.new(index_parts)
+        if #scan_value == 0 and after_tuple ~= nil then
```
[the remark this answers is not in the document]
oleg-jukovec: I have no idea how to test the internal behavior. I'm not sure if the internal behavior needs to be tested too.
But we already have integration tests for select(nil, {after}).

test/integration/select_test.lua
```diff
@@ -537,6 +612,14 @@ pgroup.test_eq_condition_with_index = function(g)
     t.assert_equals(err, nil)
     local objects = crud.unflatten_rows(result.rows, result.metadata)
     t.assert_equals(objects, helpers.get_objects_by_idxs(customers, {1})) -- in id order
+
+    -- after obj 3
```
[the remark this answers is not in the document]
oleg-jukovec: I've updated the tests. Outdated.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
+* `crud.select()` pagination with a EQ/REQ condition and `after` and
```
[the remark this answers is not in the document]
oleg-jukovec: Yeah.. I tried to reformulate the change log entries.

oleg-jukovec:
> Does it block PR about `select(nil)` warning or they exist independently?

It does not. But after the pull request we can delete a doc entry:
```
**Note:**

Pagination with parameter ``after`` leads to a full scan even with the condition '=' or '==' on the index fields.
...
```
from #277.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 ### Changed
+* `crud.select()` without conditions starts with an `after` key if
```
DifferentialOrange, line 13: Starts "with an `after` key" or starts "after `after` key"? Is starting after `after` key is added to conditions as usual?

crud/select/executor.lua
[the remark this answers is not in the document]
DifferentialOrange: Because when I first saw this commit, my first thought was "before this patch `select(nil)` with `after` extracted all records, now it extracts only records starting after `after`"

test/unit/select_plan_test.lua
```diff
@@ -276,7 +276,7 @@ g.test_first = function()
     t.assert_equals(plan.scan_value, {90}) -- after_tuple id
     t.assert_equals(plan.after_tuple, after_tuple)
     t.assert_equals(plan.scan_condition_num, 1)
-    t.assert_equals(plan.tarantool_iter, box.index.GE) -- inverted iterator
+    t.assert_equals(plan.tarantool_iter, box.index.EQ) -- inverted iterator
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, removed.

CHANGELOG.md
oleg-jukovec: I rewrote the message, but I don't mind to delete it.

DifferentialOrange: If it's an optimization, I think it's better to write something like "optimize `crud.select({}, {after = tuple})`" without digging too deep on what exactly was optimized. However, to prove that it's really an optimization we need to run some performance tests after and before the fix.

If it isn't something that users will notice even in terms of optimization, it's better to remove changelog entry at all to not confuse them.

oleg-jukovec: It seems to me that this is not a very common case of using a crud.select to write and mantain a separate performance test. So i've just remove the entry from the changelog.

oleg-jukovec: But you can see the difference with the code:

```lua
box.schema.create_space('developers', {
    format = {
    }
})
box.space.developers:create_index('primary_index', {
    parts = {
        {field = 1, type = 'unsigned'},
    },
})
-- Fill the space.
for i=1, 1000000 do
     box.space.developers:insert({i})
end
crud.select('developers', {}, {after = {999990}})
```

DifferentialOrange: > It seems to me that this is not a very common case of using a crud.select to write and maintain a separate performance test.

It's quite the contrary: most performance tests in crud module measure `select` performance. But it's okay if you don't want to write and maintain a performance test: it's really bothersome. Moreover, it's not like it will be widely used after this patch.

DifferentialOrange: I've tried to run benchmark tests, but dispersion is too high to really catch the performance upgrade.

I think it's still better to leave a changelog entry even if we couldn't specify what actual improvement is in numbers.

DifferentialOrange (approved):
LGTM after restoring CHANGLELOG entry and rebasing to master.

Please, request additional approve from Alexander.

CHANGELOG.md
oleg-jukovec: Thank you, I've added:
```patch
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 ### Changed
+* Optimize `crud.select()` without conditions and with `after`.
 
```

README.md
```diff
@@ -416,8 +416,8 @@ where:
 * `conditions` (`?table`) - array of [select conditions](#select-conditions)
 * `opts`:
   * `first` (`?number`) - the maximum count of the objects to return.
-     If negative value is specified, the last objects are returned
-     (`after` option is required in this case).
+     If negative value is specified, the objects behind `after` are returned
+     (`after` option is required in this case). [See pagination examples](https://github.com/tarantool/crud/blob/master/doc/select.md#pagination).
```
Totktonada, line 420: Nit: We can use a relative link here: `[pagination examples](doc/select.md#pagination)`.

oleg-jukovec: Wow, It works. Thanks!

Totktonada:
## General feedback

To be honest, it is not easy to spot ideas behind those commits. I'll not insist, but I think that more detailed descriptions would be a good step toward good query planner/executor.

I'll share my thoughts on how I would describe such changes.

<details>
<summary>Abstract example of description of planner/executor change</summary>

```markdown
select: fix <...>

Let's consider the following space:

| id | age |
| -- | --- |
|  1 |  32 |
|  2 |  22 |
|  3 |  33 |
|  4 |  24 |
|  5 |  23 |

We're traverse over the space with sorting by the `age` field with the
following condition: `age <= 30`. The page size is 2.

To make the explanation simpler, let's look on the same table, but
sorted by `age`:

| id | age |
| -- | --- |
|  2 |  22 |
|  5 |  23 |
|  4 |  24 |
|  1 |  32 |
|  3 |  33 |

And filtered by `age <= 30`:

| id | age |
| -- | --- |
|  2 |  22 |
|  5 |  23 |
|  4 |  24 |

First, we got the first page:

> crud.select('<...>', {{'<=', 'age', 30}}, {first = 2})

| id | age |
| -- | --- |
|  2 |  22 |
|  5 |  23 |

Now we want to fetch the next page:

> crud.select('<...>', {{'<=', 'age', 30}}, {first = 2, after = {5, 23}})

| id | age |
| -- | --- |
|  4 |  24 |

Then we step back to the previous page:

> crud.select('<...>', {{'<=', 'age', 30}}, {first = -2, after = {4, 24}})

| id | age |
| -- | --- |
|  2 |  22 |
|  5 |  23 |

It is where the problem appears. Before the change the response was the
following:

| id | age |
| -- | --- |
|  2 |  22 |

Several details about the query planner and executor are important here.

1. The plan contains `foo` and `bar` fields, which normally corresponds
   to a first select condition backed by an index. However in presence
   of `after` it is changed to <...>. Extra post-filter is added to the
   `baz` field. A stop condition is added to the `fiz` field.
2. When `first` value is negative the planner performs yet another
   transformation: <...>.
3. The executor just uses `foo` and `bar` disregarding `after` and
   `first` presence. However if `after` is present, it <...>.

The problem was in the wrong transformation in the query planner: it
takes into account <...>, but missed the case, when <...>. The patch
fixes the planner to support both cases.

There is another scenario, which should be taken into account: if the
index is non-unique, then <...>. Let's consider the example table
(sorted by `age`):

| id | age |
| -- | --- |
|  2 |  22 |
|  6 |  22 |
|  7 |  22 |
|  5 |  23 |
|  4 |  24 |
|  1 |  32 |
|  3 |  33 |

If we're on the second page and want to get the first one:

> crud.select('<...>', {{'>=', 'age', 30}}, {first = -2, after = {7, 22}})

| id | age |
| -- | --- |
|  2 |  22 |
|  6 |  22 |

The result is anyway correct, because it is handled by the following
code in the executor:

<..code here..>

As result the changed code produces correct result here as well.
```
</details>

Maybe my example is too wordy, but I would like to see some explanations. Not necessary so large ones.

We discussed the changes with @oleg-jukovec and @DifferentialOrange and now I understood them better. I'll summarize my understanding and why I think that the changes are okay.

## 1. select: fix LE and after results when the condition value < after

If the index is non-unique, the page border can be in a middle of sequence tuples with the same values of the key. Before the fix we would jump over all those tuples. Now it works correctly.

It was not obvious on the first glance, why we can relax the condition and internally use `<=` instead of `<`. Naive pagination implementation would just use `<` to get the next page (I assume descending order, that's why it is the next page, not previous), but it would work only for unique indexes. (Partial keys are irrelevant here, because we always can extract a full key from the `after` tuple).

The executor performs scrolling to the `after` tuple in the `scroll_to_after_tuple()` function. So in the case of the unique index we'll lookup one more tuple. In case of non-unique index, the scrolling is inevitable (while https://github.com/tarantool/tarantool/issues/5282 is not implemented).

There is `generate_value()` machinery in the executor: it replaces the key to start index scan with the key extracted from the `after` tuple. It actually performs the replacement when the key from the `after` tuple is *after* the key from the condition (with correspondence to basic direction imposed by conditions). IOW, when replacing the value allows to save some CPU cycles on linear scanning and pass the better start value to the index scan (the index works logarithmically).

Since `<` (before the patch) and `<=` (after the patch) index iterator is applied to the resulting scan value, including one we get from the `generate_value()` function, we should make sure that the proposed change works good for both cases: when we replaced the scan value and when we don't.

When we replaced it, `scroll_to_after_tuple()` will compensate it. Okay.

When we don't do that, the `after` tuple is either *before* the original scan value (and so `scroll_to_after_tuple()` will be no-op) or we have no `after` tuple and the `<` to `<=` replacement was not performed in the query planner. Okay too.

The patch looks correct for me.

Here we have the headroom for a little optimization:

* Use `>` and `<` for paginating over an *unique* index (when `after` is passed).

I would like to see this possibility mentioned somewhere in developer docs (say, in the scope of #292).

## 2. select: fix traverse start key in case condition + after

It's logical bugfix for travising in the descending order (`<` or `<=` in the first indexed condition). Okay.

## 3. select: fix EQ/REQ and after results

Wow. This change is not simple. I'll put here how I explained it for myself (based on the code and what Oleg tells me on the meeting). Some context is necessary here.

How a EQ/REQ condition + `after` was handled before the patch?

* The condition is added as a post-filter.

  Technically it is performed by this block of the code (from `plan.new()`):

  ```lua
  -- Since we use pagination we should continue iteration since after tuple.
  -- Such iterator could be run only with index:pairs(key(after_tuple), 'GT/LT').
  -- To preserve original condition we should manually inject it in `filter_conditions`
  -- See function `parse` in crud/select/filters.lua file for details.
  if scan_after_tuple ~= nil then
      if scan_iter == box.index.REQ or scan_iter == box.index.EQ then
          table.insert(conditions, conditions[scan_condition_num])
      end
  end
  ```

  We add the condition once again into the array of conditions. It is the trick to overrun the following if-condition in the `parse()` function in `crud/compare/filters.lua`:

  ```lua
  local filter_conditions = {}
  for i, condition in ipairs(conditions) do
      if i ~= opts.scan_condition_num then          -- <-- THIS ONE
          <...>
          table.insert(filter_conditions, {<...>})
      end
  end
  ```

  So the condition is present twice in the array of conditions, but ignored only once.
* The iterator is changed to `GE` or `LT` (yes, `LE` should be here, at least for non-unique index; it's a bug).
* A stop condition should logically stop the traversal if we met a tuple that does not correspond to the given EQ/REQ condition. However we have no code that would do that. It is the bug.

  See the `is_early_exit_possible()` function in `crud/compare/filters.lua`. There is no `condition_iter == box.index.EQ` or `condition_iter == box.index.REQ` case.

  I guess we can add it, but it should be enabled only for the condition that corresponds to the index we iterate over. Adding it for arbitrary condition would make usual filtering conditions be stop conditions, it is incorrect.

There are several problems in this algorithm:

* `LT` instead of `LE`.
* No stop condition.
* The 'potentially long select' warning (PR #277) is confused by the `GE`/`LT` in the query plan.

The first problem is the trivial. We can fix the remaining problems in different ways:

* Mark stop conditions explicitly in the query planner and use this information in `is_early_exit_possible()`. Mark 'GE bounded by EQ' and 'LE bounded by REQ' queries in the query plan to use this information in the code that emits the warning.
* Don't add the post-filter, don't replace EQ/REQ with GE/LE, don't replace the key from the first index-backed condition with the key extracted from the `after` tuple.

The second approach was choosen here. As I know from Oleg's words, he tried to implement something in the spirit of the first one, but met problems.

There are downsides in the choosen approach. We always start from the same key from conditions and linearly traverse the tuples to the `after` tuple. It is possible to have many same-by-the-key tuples in the case of partial key or non-unique index. While it is inevitable for a non-unique index (till we'll have https://github.com/tarantool/tarantool/issues/5282 implemented), it is technically possible to extract the full key from the `after` tuple (in case of the partial key in the condition).

I guess that the requests with partial keys and non-unique index with large sequences of the-same-key tuples are rare, so I can accept it as the temporary solution. Please, file an issue regarding the problem described above.

I'll also add an inline comment regarding the implementation (see below).

To sum up: this patch is okay as the temporary solution.

## 4. select: improve behavior of select(nil) + after

That's nice.

## 5. doc: fix crud.select the first parameter desription

It's okay.

(Typo in the commit header: desription -> description.)

## Summary

The patchset is okay for me.

I expect that you'll work on the patch descriptions to some extent (until it'll be too tiresome), that you'll file a follow up issue regarding pagination with the partial key condition (see comments to the the fourth patch), that you'll mention the little optimization opportunity for unique index scan during work on #292 (see comments to the first patch).

Totktonada:
What also possibly worth to mention in the developer documentation (#292): the key replacement[^1] code is spread across the query planner and the executor. There is the code that handles the `first < 0` case in the query planner:

```lua
if opts.first ~= nil then
    total_tuples_count = math.abs(opts.first)

    if opts.first < 0 then
        scan_iter = utils.invert_tarantool_iter(scan_iter)

        -- scan condition becomes border consition
        scan_condition_num = nil

        if scan_after_tuple ~= nil then                                -- <-- FROM HERE
            if has_keydef then
                local key_def = keydef_lib.new(scan_index.parts)
                scan_value = key_def:extract_key(scan_after_tuple)
            else
                scan_value = utils.extract_key(scan_after_tuple, scan_index.parts)
            end
        else
            scan_value = nil
        end                                                            -- <-- TO HERE
    end
end
```

However the tightly coupled (or even duplicating) code is in the `generate_value()` in the executor. That's strange. We should at least explain it or, better, rearrange the code to make it less strange.

[^1]: Replace the key extracted from the first condition backed by an index with the key extracted from the `after` tuple.

Totktonada (approved):

oleg-jukovec:
> There is the code that handles the `first < 0` case in the query planner:

it's for case:

iterator: >
scan value: 1
after value: 5

after the code:

iterator: <
scan value: 5
after value: 5
filter (and stop condition): > 1

The executor at now just choose a bigger value from scan value and after value to start execution from a nearest value to expected results. There is some logic in that, but it seems to me that the planner code and the executor code are too closely depend on each other. Maybe I will do a small refactoring in #292 to make it clearer.

oleg-jukovec:
@Totktonada Thank you, I've updated the pull request:

1. Rewrote the commit message for the LT->LE commit.
2. Split EQ/REQ commit into two separate (the bug fix + change in an internal behavior). Rewrote the commit messages.
3. Added more examples in CHANGELOG.md for non-obvious cases.
4. Added a new fix, see commit `select: fix result order with negative first`.

I didn't go into too much detail in the commit messages to keep them short as possible. But I hope that the patches have become clearer together with the examples in CHANGELOG.md. I will write more detailed documentation about the planner/executor logic in #292.

But I can add something if you think it is necessary.

crud/select/executor.lua
```diff
@@ -79,9 +79,34 @@ function executor.execute(space, index, filter_func, opts)
...
+                is_not_eq = key_def:compare_with_key(opts.after_tuple, value) ~= 0
+            else
+                local comparator = select_comparators.gen_func('==', parts)
+                local after_key = utils.extract_key(opts.after_tuple, parts)
+                is_not_eq = not comparator(after_key, value)
+            end
+            if is_not_eq then
+                -- it makes no sence to continue
+                return resp
+            end
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, you are right! I've added additional tests for this cases and fix the problem.

Totktonada:
> I guess that the requests with partial keys and non-unique index with large sequences of the-same-key tuples are rare, so I can accept it as the temporary solution. Please, file an issue regarding the problem described above.

It is the only thing that I expect from you.

The patchset looks okay for me.

PR #297, opened 2022-05-30
Make CRUD select error message more clearer.

psergee:
There is a difference between error description for select and other CRUD calls in case of CRUD storage unavailability.
```
crud.get('profile', 1)
---
- null
- line: 128
  class_name: GetError
  err: "Failed to call get on storage-side: CallError: Failed for 50b73728-f3e5-4363-bfaa-011beef0d20c:
    Function returned an error: NotInitialized: crud isn't initialized on replicaset:
    \"nil\"\nstack traceback:\n\t...ge_app/myapp/.rocks/share/tarantool/crud/common/call.lua:52:
```
```
crud.select('profile')
---
- null
- line: 173
  class_name: SelectError
  err: Procedure '_crud.select_on_storage' is not defined
  file: '...app/.rocks/share/tarantool/crud/common/sharding/init.lua'
  stack: "stack traceback:\n\t...app/.rocks/share/tarantool/crud/common/sharding/init.lua:173:
```

The select error description does not give any clues what might be the cause of this error.
Here is the updated error text:
```
crud.get('profile', 1)
---
- null
- line: 128
  class_name: GetError
  err: "Failed to call get on storage-side: CallError: Failed for 50b73728-f3e5-4363-bfaa-011beef0d20c:
    Function returned an error: NotInitialized: crud isn't initialized on replicaset:
    \"50b73728-f3e5-4363-bfaa-011beef0d20c\"\nstack traceback:\n\t...e_app/myapp/.rocks/share/tarantool/crud/common/utils.lua:701:
```
```
crud.select('profile')
---
- null
- line: 173
  class_name: SelectError
  err: "NotInitialized: crud isn't initialized on replicaset: \"50b73728-f3e5-4363-bfaa-011beef0d20c\"\nstack
    traceback:\n\t...e_app/myapp/.rocks/share/tarantool/crud/common/utils.lua:701:
```
The new error description for select also contains information about replicaset id.
Additionally, nil replicaset id is fixed in error messages for single calls: get, delete, etc.

See also #229

Issue #95, opened 2020-12-04
Wait for master until a timeout


filonenko-mikhail:
- Setup cartridge cluster with failover
- Setup separated crud-router and crud-storage
- Make a write load
- Switch master under load
- One time 

```
LuajitError: ...rk/starwars/.rocks/share/tarantool/crud/common/utils.lua:35: attempt to index field 'master' (a nil value)
```

Totktonada:
It seems, it was crud-0.0.4. The relevant code is the following:

https://github.com/tarantool/crud/blob/1fa7ffc052852c98eea15ef4c48b4411554520dc/crud/common/utils.lua#L33-L37

The code remain unchanged since 0.0.4.

It seems logical that there is a time frame, when the old master is already demoted, but the new one is not promoted yet. Also I see that vshard contains 'whether there is a master?' [checks](https://github.com/tarantool/vshard/blob/93febbf368f381179ec56254b0c34f725120dcd8/vshard/replicaset.lua#L465-L477) that confirms my guess.

We should not lean on master presence unconditionally and should report a proper error if there is no master. The error should give a user hint that it is a transient error and the request can be safely retried.

BTW, crud has no autoretries / retry strategy settings?

Totktonada:
BTW, as I see, vshard's call automatically waits for master (until a timeout). It seems, it would be good to use the same approach.

NB: Don't forget to substract time we waiting for a space from net.box request timeout.

PR #299, opened 2022-06-02
Provide an API to get storages initialization state

### LINKED ISSUE #229: Calls to routers and storages when crud-router and crud-storage roles initialization is not finished yet
There are exceptions trown in cartridge-java driver due to crud-router and crud-storage roles initialization issues.

How to repoduce:
- run some load test to call crud.select (with prefer_replica=true and balance=true) and crud.upsert
- stop/start a few routers
- stop/start a few replicas (not leaders) in replicasets

Result for select:
```
org.springframework.dao.DataRetrievalFailureException: InnerErrorMessage:
code: 33
message: Procedure '_crud.select_on_storage' is not defined; nested exception is io.tarantool.driver.exceptions.TarantoolInternalException: InnerErrorMessage:
code: 33
message: Procedure '_crud.select_on_storage' is not defined
```
Result for upsert:
```
Caused by: io.tarantool.driver.exceptions.TarantoolInternalException: InnerErrorMessage:
code: 32
message: ...example.lua:100: variable 'crud' is not declared
```

psergee:
There is an issue with using CRUD functionality if not all storages are
up. New function is added to get the information about storages
state: initialized or not. So, a user can poll state and wait for
storages to be initialized before making CRUD calls.

Looks like state.lua module is unused. Removed.

Resolves #229


crud/common/utils.lua
```diff
@@ -14,6 +14,8 @@ local ParseOperationsError = errors.new_class('ParseOperationsError', {capture_s
 local ShardingError = errors.new_class('ShardingError', {capture_stack = false})
 local GetSpaceFormatError = errors.new_class('GetSpaceFormatError', {capture_stack = false})
 local FilterFieldsError = errors.new_class('FilterFieldsError', {capture_stack = false})
+local GetReplicaStateError = errors.new_class('GetStorageStateError')
+local fiber_clock = require('fiber').clock
```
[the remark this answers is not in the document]
Totktonada: If `fiber.time()` calls are splitted by a yield, it should be okay, I guess.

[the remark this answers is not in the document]
psergee: According to doc, looks like it is recommended to use fiber.clock for timeouts because it is monotonic and returns elapsed wall-clock time. As I can see, it is implemented using _ev_monotonic_now_ call and is also used for timeouts in tarantool source code.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* `crud.get_storages_state` function to get storages initialization state.
```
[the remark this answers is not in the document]
psergee: Done.

psergee:
> Original ticket discusses some possibility to implement simple retries, while this solution provide a manual handle to check the state. Was this design approved by issue authors or product managers? Also, to be honest, I don't think current approach to state check is satisfying (see comments).

We have discussed this approach with Java connector team and @Totktonada 
Currently we could not come up with a better option to provide the information about CRUD storage init state. Timeout approach does not look useful, because storage start may take a long time and CRUD calls like select or get will "hang" if timeout is big enough. In case of small timeout, the original issue will not fade away.

test/integration/storages_state_test.lua
```diff
@@ -0,0 +1,89 @@
...
+        use_vshard = true,
+        replicasets = helpers.get_test_replicasets(),
+        env = {
+            ['ENGINE'] = g.params.engine,
+        },
+    })
+    g.cluster:start()
+
+    -- wait for storages to initialize
+    all_storages_initialized = wait_storages_init(g)
```
DifferentialOrange, line 49: Do we need it here? Isn't luatest wait for instance roles init?

DifferentialOrange (approved):
Thank you, LGTM!

I'll merge it after https://github.com/tarantool/crud/pull/232 . Maybe you'll need to rebase this PR and solve remaining conflict before merge.

test/integration/storages_state_test.lua
DifferentialOrange: Seems like it is
https://github.com/tarantool/cartridge/blob/9ea29aed1a480df0ebb4bf42b0bc0c5903a24023/cartridge/test-helpers/cluster.lua#L338-L346

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* `crud.get_storages_state` function to get storages initialization state (#229).
```
Totktonada, line 11: Nits:

1. AFAIK usage of plural nouns in a function/field name is discouraged.
2. We usually leave getters without `get_` prefix.
3. tarantool uoften uses `.info()` in this sense, so, maybe, `crud.storage_info()`?

crud.lua
```diff
@@ -111,6 +111,10 @@ crud.stats = stats.get
 -- @function reset_stats
 crud.reset_stats = stats.reset
 
+-- @refer utils.get_storages_state
+-- @function get_storages_state
+crud.get_storages_state = utils.get_storages_state
```
Totktonada, line 116: We should add it into the README (and don't forget to re-run `doctoc`).

test/integration/storages_state_test.lua
psergee: Yes you are right, cluster start waits for roles configured. I wanted to rely on this also, but... I need this code because sometimes I get "peer closed" error for some connections right after the cluster start. Retry is required to give a small timeout to reconnect. As I can see in logs, disconnect and reconnect always occur right after cluster start in all tests. And in my case reconnect does not happen fast enough sometimes.
Suppose, we should investigate this behavior. But I can do nothing with this from CRUD side.

crud/common/utils.lua
```diff
@@ -711,4 +713,55 @@ function utils.update_storage_call_error_description(err, func_name, replicaset_
...
+    end
+
+    opts = opts or {}
+
+    local futures_by_replicas = {}
+    local replicas_state = {}
+    local async_opts = {is_async = true}
+    local timeout = opts.timeout or const.DEFAULT_VSHARD_CALL_TIMEOUT
+
+    for _, replicasets in pairs(replicasets) do
```
Totktonada, line 737: replicasets -> replicaset?

crud/common/utils.lua
```diff
@@ -711,4 +713,55 @@ function utils.update_storage_call_error_description(err, func_name, replicaset_
...
+        for replica_uuid, replica in pairs(replicasets.replicas) do
+            replicas_state[replica_uuid] = {name = replica.name, storage_initialized = false}
+            local ok, future = pcall(replica.conn.eval, replica.conn,
+                                [[return rawget(rawget(_G, "_crud") or {}, "is_init") ~= nil]],
+                                {}, async_opts)
```
Totktonada, line 742: Please, use `call`. You need a function on the storage side, yep.

`eval` assumes ability to do everything. We should not require it to call the status check function.

crud/common/utils.lua
```diff
@@ -711,4 +713,55 @@ function utils.update_storage_call_error_description(err, func_name, replicaset_
...
+            if ok then
+                futures_by_replicas[replica_uuid] = future
+            end
```
Totktonada, line 745: What kind of errors are handled there? Why we don't add the storage into the result?

crud/common/utils.lua
```diff
@@ -711,4 +713,55 @@ function utils.update_storage_call_error_description(err, func_name, replicaset_
...
+        if wait_timeout < 0 then
+            wait_timeout = 0
+        end
+
+        local result = future:wait_result(wait_timeout)
+        if result then
+            replicas_state[replica_uuid].storage_initialized = result[1]
+        else
+            future:discard()
+        end
```
Totktonada, line 761: The same question as above: why we don't add infomation about this storage into the result?

I would also note that in this case we should also add what is not good with the storage: IOW, the error message we received.

crud/common/utils.lua
```diff
@@ -711,4 +713,55 @@ function utils.update_storage_call_error_description(err, func_name, replicaset_
...
+    opts = opts or {}
+
+    local futures_by_replicas = {}
+    local replicas_state = {}
+    local async_opts = {is_async = true}
+    local timeout = opts.timeout or const.DEFAULT_VSHARD_CALL_TIMEOUT
+
+    for _, replicasets in pairs(replicasets) do
+        for replica_uuid, replica in pairs(replicasets.replicas) do
+            replicas_state[replica_uuid] = {name = replica.name, storage_initialized = false}
```
Totktonada, line 739: From where `name` comes? I don't see it in vshard's documntation.

Totktonada:
> Original ticket discusses some possibility to implement simple retries, while this solution provide a manual handle to check the state. Was this design approved by issue authors or product managers?

Discussed between me, Pavel. S and cartridge-java developers. We agreed that a function that will return status of storages will suit the needs.

I can share my vision of the problem. There are a bit different situations:

1. Waiting for the cluster boostrap (full or partial).
2. Run queries.

There are different possible mechanisms that can help in those situations:

1. A status endpoint.
2. Timeouts for queries.
3. Contract on errors that allows a client to spot the non-bootstrapped storage situation and retry.

Timeouts for queries is the most used mechanish and it allows to overcome small periods of a peer unavailability. We support it, but there are flaws: #95.

The also unified error messages for the non-bootstrapped storage case, see PR #296 and PR #297. (It is not great that we propose to match the error message to make programmatic decisions, but errors in crud -- and even in tarantool itself -- requires a redesign and rework IMHO. At least, some revision.)

However we still don't offer anything to wait for a potentially long times (over usual query timeout) until the cluster will go into the ready state. It is what we want to solve here.

crud/common/utils.lua
psergee: So, we will must process "undefined function" error, which will be returned if a storage is not initialized yet. What is the point of calling the function if the fact of its existence means initialized storage?

crud/common/utils.lua
psergee: Closed connection, for example. We add a storage into the result right before pcall. But there is no future in case of eval call error. So, storage status of this replica remains unchanged: uninit.

crud/common/utils.lua
psergee: Non-public api. Removed.

CHANGELOG.md
psergee: Done.

Totktonada (approved):
Thanks! The patchset is okay for me in general. However, please, consider several comments and thoughts.

test/integration/storages_state_test.lua
Totktonada: Let's just add your current observations into a comment in the code and move beyond. It is fair to let future readers know why the function is here.

crud/common/utils.lua
```diff
@@ -748,4 +751,65 @@ function utils.list_slice(list, start_index, end_index)
...
+        if wait_timeout < 0 then
+            wait_timeout = 0
+        end
+
+        local result, err = future:wait_result(wait_timeout)
+        if result == nil then
+            future:discard()
+            if err ~= nil then
+                local str_err = tostring(err)
+                if (string.find(str_err, " is not defined") ~= nil) then
```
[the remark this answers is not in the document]
psergee: Done

crud.lua
```diff
@@ -165,6 +169,8 @@ function crud.init_storage()
     count.init()
     borders.init()
     sharding_metadata.init()
+
+    _G._crud.is_initialized = function() return true end
```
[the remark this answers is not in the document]
psergee: Re-worked to return a table and added _on_storage suffix.

README.md
```diff
@@ -1074,6 +1075,47 @@ crud.len('customers')
...
+**Example:**
+
+```lua
+crud.storage_info()
+```
+```
+---
+- 5c3392a3-ce89-4aec-83f3-6cb5f18e60c3:
+    status: error
+    is_master: true
```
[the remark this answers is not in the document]
psergee: Done.

crud/common/utils.lua
```diff
@@ -748,4 +751,75 @@ function utils.list_slice(list, start_index, end_index)
...
+            replica_state_by_uuid[replica_uuid] = {status = "error",
+                                                   is_master = replicaset.master == replica}
+            local ok, res = pcall(replica.conn.call, replica.conn, "_crud.storage_info_on_storage",
+                                  {}, async_opts)
+            if ok then
+                futures_by_replicas[replica_uuid] = res
+            elseif res ~= nil then
+                log.error("Error getting storage info for %s: %s", replica_uuid, res)
+                replica_state_by_uuid[replica_uuid].message = tostring(res)
+            end
```
Totktonada, line 786: Shouldn't we fill `message` if an error occurs here (if `ok` is `false` disregarding whether `res` is `nil`)?

psergee: We need to log it, but I'm not sure about usefulness of common error message.
Done.

Totktonada: Is it ever possible that we'll get `false, nil` result from `pcall(replica.conn.call, <...>)`? Anyway, even if so, I think that it would be misleading to see `status = 'error'` without any message.

Thanks for the update.

test/integration/storages_state_test.lua
```diff
@@ -0,0 +1,170 @@
...
+end)
+
+pgroup.after_all(function(g)
+    helpers.stop_cluster(g.cluster)
+    fio.rmtree(g.cluster.datadir)
+end)
+
+pgroup.test_crud_storage_status_of_stopped_servers = function(g)
+    t.assert_equals(all_storages_initialized, true)
```
DifferentialOrange, line 61: I see in

```lua
t.assert_equals(all_storages_initialized, true)
```
```lua
t.assert_equals(wait_storages_init(g), true)
```

that you rely on tests order and this is not good. Moreover, it seems like you should extract this thing to `before_each` luatest hook. Moreover, if something went wrong in test asserts, instances won't be restarted, so you should extract these parts to `after_each` or `after_test(test_name, ...)` hooks.

test/integration/storages_state_test.lua
```diff
@@ -0,0 +1,170 @@
...
+
+    -- stop crud storage role on one replica
+    local server = g.cluster:server("s1-replica")
+    local results = server.net_box:eval([[
+        local serviceregistry = require("cartridge.service-registry")
+        serviceregistry.get("crud-storage").stop()
+        return true
+    ]])
+
+    t.assert_not_equals(results, nil, "Fail to disable storage role")
```
DifferentialOrange, line 101:
```suggestion
    t.assert_not_equals(results, nil, "Failed to disable storage role")
```

test/integration/storages_state_test.lua
```diff
@@ -0,0 +1,170 @@
...
+        }
+    })
+
+    local results = server.net_box:eval([[
+        local serviceregistry = require("cartridge.service-registry")
+        serviceregistry.get("crud-storage").init()
+        return true
+    ]])
+
+    t.assert_not_equals(results, nil, "Fail to enable storage role")
```
DifferentialOrange, line 131:
```suggestion
    t.assert_not_equals(results, nil, "Failed to enable storage role")
```

test/integration/storages_state_test.lua
```diff
@@ -0,0 +1,170 @@
...
+
```
DifferentialOrange, line 132: Extra newline

test/integration/storages_state_test.lua
```diff
@@ -0,0 +1,170 @@
...
+            status = "running",
+            is_master = true
+        },
+        [helpers.uuid('c', 2)] = {
+            status = "error",
+            is_master = false,
+            message = "attempt to call a table value"
+        }
+    })
+
```
DifferentialOrange, line 169: Extra newline

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* `crud.storage_info` function to get storages status (#229).
```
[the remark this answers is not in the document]
psergee: Issue is linked with PR. Why do we need redundant references here?

test/integration/storages_state_test.lua
DifferentialOrange: There is a two separate comments. I should've add two comments instead of one.

1. First test asserts `t.assert_equals(all_storages_initialized, true)` (because `all_storages_initialized` was already computed in `before_all`), all other tests use `t.assert_equals(wait_storages_init(g), true)` (because status has changed and they need to recalculate it). Relying on tests order is not good (for example, if one uses `--shuffle` the logic may break), so let's use honest computation `t.assert_equals(wait_storages_init(g), true)` before each test. And since we want to use it before each test, let's use `before_each` hook instead of copypaste.

2. Some test cases stop servers or disable roles to arrange non-trivial cluster status and validate handle logic. To restore cluster state to a default one, test cases restart servers/enable roles. But since the code is the part of a test case, if some assert will fail, servers won't be restarted/roles won't enabled. Thus next test case will start with inconsistent cluster and will fail too. Let's handle restarts/reenables in `after_test` hooks: they can be used to clean up specific test setups even if test has failed.

README.md
```diff
@@ -1074,6 +1075,51 @@ crud.len('customers')
...
+
+```lua
+-- Get storages status
+local result, err = crud.storage_info(opts)
+```
+
+where:
+
+* `opts`:
+  * `timeout` (`?number`) -  maximum time (in seconds) to wait for response from
```
DifferentialOrange, line 1088: By default we wait infinitely?

README.md
```diff
@@ -1074,6 +1075,51 @@ crud.len('customers')
...
+  cluster instances.
+
+Returns storages status table by instance UUID or nil with error. Status table fields:
+"status" contains a string representing the status:
+* "running" - storage is initialized and running.
+* "uninitialized" - storage is not initialized or disabled.
+* "error" - error getting the status from a storage. Connection error, for example.
+"is_master" is true if an instance is a master. False - otherwise.
```
DifferentialOrange, line 1096:
```suggestion
"is_master" is `true` if an instance is a master. `false` - otherwise.
```

README.md
```diff
@@ -1074,6 +1075,51 @@ crud.len('customers')
...
+```
+
+where:
+
+* `opts`:
+  * `timeout` (`?number`) -  maximum time (in seconds) to wait for response from
+  cluster instances.
+
+Returns storages status table by instance UUID or nil with error. Status table fields:
+"status" contains a string representing the status:
```
DifferentialOrange, line 1092: Maybe it's better to do a two-level list here?

DifferentialOrange, line 1092: I think it's better to write table field names and values as inline code (`status`, `"running"` etc).

README.md
```diff
@@ -1074,6 +1075,51 @@ crud.len('customers')
...
+* "running" - storage is initialized and running.
+* "uninitialized" - storage is not initialized or disabled.
+* "error" - error getting the status from a storage. Connection error, for example.
+"is_master" is true if an instance is a master. False - otherwise.
+"message" is nil unless a problem occurs with getting storage status.
```
DifferentialOrange, line 1097:
```suggestion
"message" is `nil` unless a problem occurs with getting storage status.
```

DifferentialOrange:
Since there are some GitHub folding, maybe you've missed this unresolved comments so I'll write them down here:
- https://github.com/tarantool/crud/pull/299#discussion_r924394773
- https://github.com/tarantool/crud/pull/299#discussion_r924396533
- https://github.com/tarantool/crud/pull/299#discussion_r924396979
- https://github.com/tarantool/crud/pull/299#discussion_r924398270

test/integration/storages_state_test.lua
```diff
@@ -0,0 +1,170 @@
...
+})
+
+local all_storages_initialized = false
+
+local function wait_storages_init(g)
+    local storages_initialized = false
+    local attempts_left = 5
+    local wait_for_init_timeout = 1
+    while (attempts_left > 0 and not storages_initialized) do
+        local results, err = g.cluster.main_server.net_box:call("crud.storage_info", {})
```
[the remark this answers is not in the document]
psergee: What's wrong with this?

test/integration/storages_state_test.lua
```diff
@@ -0,0 +1,170 @@
...
+        t.assert_equals(err, nil, "Error getting storage status")
+        storages_initialized = true
+        for _,v in pairs(results) do
```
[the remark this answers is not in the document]
psergee: I don't think there is a big difference between not-yet-enabled vshard-storage role and not started server. I observe all servers in replicasets regardless of their roles enabled states.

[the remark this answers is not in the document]
DifferentialOrange: I mean, don't we need to check that there are 4 servers in results?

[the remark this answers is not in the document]
psergee: Done.

DifferentialOrange (approved):

PR #315, opened 2022-07-22
sharding func: fix specifying `vshard` sharding funcs

### LINKED ISSUE #314: Fix specifying `vshard` sharding func
Starting from 0.11.0 user can specify sharding function to calculate bucket_id
with sharding func definition as a part of
[DDL schema](https://github.com/tarantool/ddl#input-data-format)
or insert manually to the space `_ddl_sharding_func`.

Right now `ddl` has a bug https://github.com/tarantool/ddl/issues/91
But even if this bug is fixed, there is also a bug on CRUD side.
Inserting manually to the space `_ddl_sharding_func` showed
that CRUD search `vshard` sharding func in `_G` but this
approach doesn't work with `vshard` case.

AnaNek:
Starting from 0.11.0 user can specify sharding func to
calculate bucket_id with sharding func definition as a part of
DDL schema or insert manually to the space `_ddl_sharding_func`.

Right now ddl fails with setting schema with vshard sharding function
(https://github.com/tarantool/ddl/issues/91).
But even if this bug is fixed, there is also a bug on CRUD side.
Inserting manually to the space `_ddl_sharding_func` showed
that CRUD search vshard sharding func in `_G` but this
approach doesn't work with vshard case.

This patch allows to specify `vshard` sharding func
inserting manually to the space `_ddl_sharding_func`.

I didn't forget about

- [x] Tests
- [x] Changelog
- Documentation

Closes #314

### DIFF
--- a/crud/common/sharding/sharding_func.lua
+++ b/crud/common/sharding/sharding_func.lua
@@ -9,6 +9,10 @@ local ShardingFuncError = errors.new_class('ShardingFuncError',  {capture_stack 
 
 local sharding_func_module = {}
 
+local sharding_module_names = {
+    ['vshard'] = true,
+}
+
 local function is_callable(object)
     if type(object) == 'function' then
         return true
@@ -43,13 +47,33 @@ end
 local function get_function_from_G(func_name)
     local chunks = string.split(func_name, '.')
     local sharding_func = _G
+    local sharding_module = false
+    local ok
+
+    if sharding_module_names[chunks[1]] then
+        ok, sharding_func = pcall(require, chunks[1])
+        if not ok then
+            return nil
+        end
+
+        sharding_module = true
+        table.remove(chunks, 1)
+    end
 
     -- check is the each chunk an identifier
     for _, chunk in pairs(chunks) do
         if not utils.check_name_isident(chunk) or sharding_func == nil then
             return nil
         end
-        sharding_func = rawget(sharding_func, chunk)
+
+        -- `vshard` store sharding functions in metatable,
+        -- this metatable is common for all `vshard` routers.
+        -- That's why for `vshard` case we can't use rawget.
+        if sharding_module then
+            sharding_func = sharding_func[chunk]
+        else
+            sharding_func = rawget(sharding_func, chunk)
+        end
     end
 
     return sharding_func
--- a/test/entrypoint/srv_ddl.lua
+++ b/test/entrypoint/srv_ddl.lua
@@ -161,6 +161,14 @@ package.preload['customers-storage'] = function()
             local customers_G_func_schema = table.deepcopy(customers_id_key_schema)
             customers_G_func_schema.sharding_func = 'some_module.sharding_func'
 
+            local customers_empty_sharding_func_schema = table.deepcopy(customers_id_key_schema)
+
+            local customers_vshard_mpcrc32_schema = table.deepcopy(customers_id_key_schema)
+            customers_vshard_mpcrc32_schema.sharding_func = 'vshard.router.bucket_id_mpcrc32'
+
+            local customers_vshard_strcrc32_schema = table.deepcopy(customers_id_key_schema)
+            customers_vshard_strcrc32_schema.sharding_func = 'vshard.router.bucket_id_strcrc32'
+
             local schema = {
                 spaces = {
                     customers = customers_id_schema,
@@ -173,6 +181,9 @@ package.preload['customers-storage'] = function()
                     customers_name_age_key_three_fields_index = customers_name_age_key_three_fields_index_schema,
                     customers_G_func = customers_G_func_schema,
                     customers_body_func = customers_body_func_schema,
+                    customers_empty_sharding_func = customers_empty_sharding_func_schema,
+                    customers_vshard_mpcrc32 = customers_vshard_mpcrc32_schema,
+                    customers_vshard_strcrc32 = customers_vshard_strcrc32_schema,
                 }
             }
 
@@ -188,7 +199,9 @@ package.preload['customers-storage'] = function()
                 box.space['_ddl_sharding_key']:update(space_name, {{'=', fieldno_sharding_key, sharding_key_def}})
             end)
             rawset(_G, 'set_sharding_func', function(space_name, fieldno_sharding_func, sharding_func_def)
-                box.space['_ddl_sharding_func']:update(space_name, {{'=', fieldno_sharding_func, sharding_func_def}})
+                local record = {space_name, box.NULL, box.NULL}
+                record[fieldno_sharding_func] = sharding_func_def
+                box.space['_ddl_sharding_func']:replace(record)
             end)
         end,
     }
--- a/test/integration/ddl_sharding_func_test.lua
+++ b/test/integration/ddl_sharding_func_test.lua
@@ -21,6 +21,11 @@ local cache_group = t.group('ddl_sharding_func_cache', {
     {engine = 'vinyl'},
 })
 
+local vshard_group = t.group('ddl_vshard_sharding_func', {
+    {engine = 'memtx'},
+    {engine = 'vinyl'},
+})
+
 pgroup.before_all(function(g)
     g.cluster = helpers.Cluster:new({
         datadir = fio.tempdir(),
@@ -75,6 +80,35 @@ cache_group.after_all(function(g) helpers.stop_cluster(g.cluster) end)
 cache_group.before_each(function(g)
     helpers.truncate_space_on_cluster(g.cluster, 'customers_G_func')
     helpers.truncate_space_on_cluster(g.cluster, 'customers_body_func')
+end)
+
+vshard_group.before_all(function(g)
+    g.cluster = helpers.Cluster:new({
+        datadir = fio.tempdir(),
+        server_command = helpers.entrypoint('srv_ddl'),
+        use_vshard = true,
+        replicasets = helpers.get_test_replicasets(),
+        env = {
+            ['ENGINE'] = g.params.engine,
+        },
+    })
+    g.cluster:start()
+    local result, err = g.cluster.main_server.net_box:eval([[
+        local ddl = require('ddl')
+
+        local ok, err = ddl.get_schema()
+        return ok, err
+    ]])
+    t.assert_equals(type(result), 'table')
+    t.assert_equals(err, nil)
+end)
+
+vshard_group.after_all(function(g) helpers.stop_cluster(g.cluster) end)
+
+vshard_group.before_each(function(g)
+    helpers.truncate_space_on_cluster(g.cluster, 'customers_vshard_mpcrc32')
+    helpers.truncate_space_on_cluster(g.cluster, 'customers_vshard_strcrc32')
+    helpers.truncate_space_on_cluster(g.cluster, 'customers_empty_sharding_func')
 end)
 
 pgroup.test_insert_object = function(g)
@@ -703,7 +737,7 @@ cache_group.test_update_cache_with_incorrect_func = function(g)
 
     -- records for all spaces exist
     local cache_size = helpers.get_sharding_func_cache_size(g.cluster)
-    t.assert_equals(cache_size, 2)
+    t.assert_equals(cache_size, 4)
 
     -- no error just warning
     local space_name = 'customers_G_func'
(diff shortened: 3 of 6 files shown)


test/entrypoint/srv_ddl.lua
```diff
@@ -188,7 +191,9 @@ package.preload['customers-storage'] = function()
                 box.space['_ddl_sharding_key']:update(space_name, {{'=', fieldno_sharding_key, sharding_key_def}})
             end)
             rawset(_G, 'set_sharding_func', function(space_name, fieldno_sharding_func, sharding_func_def)
-                box.space['_ddl_sharding_func']:update(space_name, {{'=', fieldno_sharding_func, sharding_func_def}})
+                local record = {space_name}
```
DifferentialOrange, line 194:
```suggestion
                local record = {space_name, box.NULL, box.NULL}
```

test/entrypoint/srv_ddl.lua
```diff
@@ -188,7 +191,9 @@ package.preload['customers-storage'] = function()
...
+                record[fieldno_sharding_func] = sharding_func_def
+                box.space['_ddl_sharding_func']:upsert(record, {{'=', fieldno_sharding_func, sharding_func_def}})
```
DifferentialOrange, line 196: Why not replace? If there is a both second and third field will be filled, it would be an invalid ddl schema.

crud/common/sharding/sharding_func.lua
```diff
@@ -43,13 +47,25 @@ end
...
     local sharding_func = _G
+    local sharding_module = false
+
+    if sharding_module_names[chunks[1]] then
+        sharding_func = require(chunks[1])
+        sharding_module = true
+        table.remove(chunks, 1)
+    end
 
     -- check is the each chunk an identifier
```
DifferentialOrange, line 58: I don't get how this "multiple chunk" load works (and not sure it is valid).

test/entrypoint/srv_ddl.lua
```diff
@@ -188,7 +191,9 @@ package.preload['customers-storage'] = function()
                 box.space['_ddl_sharding_key']:update(space_name, {{'=', fieldno_sharding_key, sharding_key_def}})
             end)
             rawset(_G, 'set_sharding_func', function(space_name, fieldno_sharding_func, sharding_func_def)
```
DifferentialOrange, line 193: To be honest, I would've prefer merging https://github.com/tarantool/ddl/pull/98 first and then setting up a full integration test with honest `ddl.set_schema` (like I did in sharding reload). https://github.com/tarantool/crud/blob/a7a3413ba283a1b4f5c64ec1fa87c3ef13b9b6cc/test/entrypoint/srv_ddl_reload.lua#L156-L174

DifferentialOrange (commented):
Seems LGTM, but see my comment below

test/entrypoint/srv_ddl.lua
AnaNek: Changed:

https://github.com/tarantool/crud/blob/1d98a493fda2765dd4a6ecb8d93a38bd9293282d/test/entrypoint/srv_ddl.lua#L194

test/entrypoint/srv_ddl.lua
AnaNek: Good idea, changed:

https://github.com/tarantool/crud/blob/1d98a493fda2765dd4a6ecb8d93a38bd9293282d/test/entrypoint/srv_ddl.lua#L196

test/entrypoint/srv_ddl.lua
AnaNek: Sounds logical, but I would like to have tests with inserting sharding func manually to the space `_ddl_sharding_func` as well. Let's merge https://github.com/tarantool/ddl/pull/98 and then add tests with `ddl` here.

crud/common/sharding/sharding_func.lua
AnaNek: Before this patch we got record by `chunk` key from `_G`, but this approach doesn't work with `vshard` because `require('vshard')` doesn't set `vshard` key in `_G`. So, in vshard case we got record directly from `vshard` table, not from `_G`.

crud/common/sharding/sharding_func.lua
```diff
@@ -43,13 +47,25 @@ end
 local function get_function_from_G(func_name)
     local chunks = string.split(func_name, '.')
     local sharding_func = _G
+    local sharding_module = false
+
+    if sharding_module_names[chunks[1]] then
+        sharding_func = require(chunks[1])
```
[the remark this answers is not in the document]
AnaNek: Good point! Thank you! Here is the fix:

```diff
     local sharding_func = _G
     local sharding_module = false
+    local ok
 
     if sharding_module_names[chunks[1]] then
-        sharding_func = require(chunks[1])
+        ok, sharding_func = pcall(require, chunks[1])
+        if not ok then
+            return nil
+        end
+
         sharding_module = true
```

test/integration/ddl_sharding_func_test.lua
```diff
@@ -938,3 +939,68 @@ pgroup.test_gh_278_count_with_explicit_bucket_id_and_ddl = function(g)
...
+    },
+    strcrc32 = {
+        sharding_func_name = 'vshard.router.bucket_id_strcrc32',
+        bucket_id = 477,
+        srv_with_data = 's2-master',
+        srv_without_data = 's1-master',
+    }
+}
+
+for name, case in pairs(vshard_cases) do
```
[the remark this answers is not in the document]
AnaNek: This is a common parameterization approach for CRUD tests https://github.com/tarantool/crud/blob/d053623d616682bc9d8dc233a141c1b0063a22d9/test/integration/ddl_sharding_info_reload_test.lua#L455-L478, I suppose the reason for using this option is the following - naming for tests with many parameters is hard readable, for example:

```
    ddl_sharding_func_vshard.engine:"memtx".vshard_sharding_case:{
    bucket_id = 477,
    sharding_func_name = "vshard.router.bucket_id_strcrc32",
    srv_with_data = "s2-master",
    srv_without_data = "s1-master",
}.test_vshard_sharding_func ... (0.049s) Ok
```

I agree that luatest parametrization is good for https://github.com/tarantool/ddl/pull/98, but in this case I'm not sure.

crud/common/sharding/sharding_func.lua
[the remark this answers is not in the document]
DifferentialOrange: > but vshard module is not available

https://github.com/tarantool/crud/blob/d053623d616682bc9d8dc233a141c1b0063a22d9/crud-scm-1.rockspec#L12

test/entrypoint/srv_ddl.lua
AnaNek: Now there are tests integrated with `ddl.set_schema`.

test/integration/ddl_sharding_func_test.lua
```diff
@@ -938,3 +972,89 @@ pgroup.test_gh_278_count_with_explicit_bucket_id_and_ddl = function(g)
...
+            local record_exist, err = helpers.update_sharding_func_cache(g.cluster, space_name)
+            t.assert_equals(err, nil)
+            t.assert_equals(record_exist, true)
+        end
+
+        -- Insert a tuple.
+        local result, err = g.cluster.main_server.net_box:call(
+                'crud.insert', {space_name, {1, box.NULL, 'Ivan', 25}})
+        t.assert_equals(err, nil)
+        t.assert_equals(result.metadata, {
```
DifferentialOrange, line 1032: I think it's excessive

test/integration/ddl_sharding_func_test.lua
```diff
@@ -938,3 +972,89 @@ pgroup.test_gh_278_count_with_explicit_bucket_id_and_ddl = function(g)
...
+            {is_nullable = false, name = 'id', type = 'unsigned'},
+            {is_nullable = false, name = 'bucket_id', type = 'unsigned'},
+            {is_nullable = false, name = 'name', type = 'string'},
+            {is_nullable = false, name = 'age', type = 'number'},
+        })
+        t.assert_equals(#result.rows, 1)
+        t.assert_equals(result.rows[1], {1, case.bucket_id, 'Ivan', 25})
+
+        -- There is a tuple on server that we inserted before using crud.insert().
```
DifferentialOrange, line 1041:
```suggestion
        -- There is a tuple on server that we inserted with crud.insert().
```

crud/common/sharding/sharding_func.lua
```diff
@@ -43,13 +47,30 @@ end
...
+        sharding_module = true
+        table.remove(chunks, 1)
+    end
 
     -- check is the each chunk an identifier
     for _, chunk in pairs(chunks) do
         if not utils.check_name_isident(chunk) or sharding_func == nil then
             return nil
         end
-        sharding_func = rawget(sharding_func, chunk)
```
DifferentialOrange, line 52: Shouldn't `rawget` satisfy both cases?

test/entrypoint/srv_ddl.lua
DifferentialOrange: Well, there is no any `ddl.set_schema` calls. As far as I understand, `ddl` isn't good with migrations, but the thing that I used in `srv_ddl_reload` seems close to a real ddl migration. I think it would be better to use the whole `set_sharding_func_name` implementation from `srv_ddl_reload`.

DifferentialOrange (commented):
Should be ok after resolving remaining comments

test/entrypoint/srv_ddl.lua
AnaNek: `ddl.set_schema` call is here https://github.com/tarantool/crud/blob/90f0c714b7daf29240a71e3804a81d734b1f9aa7/test/entrypoint/srv_ddl.lua#L191

test/integration/ddl_sharding_func_test.lua
AnaNek: What exactly?

AnaNek: Checking metadata?

test/integration/ddl_sharding_func_test.lua
AnaNek: Fixed, thanks!

test/integration/ddl_sharding_func_test.lua
DifferentialOrange: Yes

crud/common/sharding/sharding_func.lua
DifferentialOrange: Let's add a comment in code so it will be obvious for future generations

test/entrypoint/srv_ddl.lua
DifferentialOrange: `ddl.set_schema` is called without any sharding functions in schema and then we set up sharding functions later in test cases with `set_sharding_func` helper.

DifferentialOrange: So if you try to test this with older ddl without any vshard support, it still will be ok.

DifferentialOrange: The idea is that user will commonly use a single `ddl.set_schema` with vshard sharding functions in the beginning and then make crud calls. Since we build many different test cases here, it is hard to use the same approach without any reworks (and recreating a new Tarantool instance for each case), but using the same helpers as in `srv_ddl_reload`, in my opinion, is a bit much honest.

AnaNek: Maybe I don't get your remark, but `ddl.set_schema` is called with `vshard` sharding functions in schema https://github.com/tarantool/crud/blob/90f0c714b7daf29240a71e3804a81d734b1f9aa7/test/entrypoint/srv_ddl.lua#L166-L170

crud/common/sharding/sharding_func.lua
AnaNek: Done:

https://github.com/tarantool/crud/blob/a479871fd0e793990aab7e0b0ca86ae59277e5ce/crud/common/sharding/sharding_func.lua#L69-L71

test/integration/ddl_sharding_func_test.lua
AnaNek: Okay, removed.

test/entrypoint/srv_ddl.lua
DifferentialOrange: Oh, sorry, you are right. I looks like I've confused the approach in ddl and ddl_reload tests.

DifferentialOrange (approved):

PR #317, opened 2022-08-03
doc: describe schema reload implementation

### LINKED ISSUE #253: design document: how database schema information is renewed

DifferentialOrange:
This patch adds a small document describing how space and sharding schemas works in the module. It covers what schemas are, how we store, use and reload them in requests.

I didn't forget about

- Tests (not required)
- Changelog (since it's internal, I don't think it's worth it)
- [x] Documentation

Closes #253

### DIFF
--- a/doc/dev/schema.md
+++ b/doc/dev/schema.md
@@ -1,0 +1,227 @@
+# Database schema information design document
+
+Two types of schema are used in ``crud`` requests: ``net.box`` spaces
+schema and ``ddl`` sharding schema. If a change had occurred in one of
+those, router instances should reload the schema and reevaluate
+a request using an updated one. This document clarifies how schema
+is obtained, used and reloaded.
+
+## Table of Contents
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Space schema](#space-schema)
+  - [How schema is stored](#how-schema-is-stored)
+  - [When schema is used](#when-schema-is-used)
+  - [How schema is reloaded](#how-schema-is-reloaded)
+  - [When schema is reloaded and operation is retried](#when-schema-is-reloaded-and-operation-is-retried)
+  - [Alternative approaches](#alternative-approaches)
+- [Sharding schema](#sharding-schema)
+  - [How schema is stored](#how-schema-is-stored-1)
+  - [When schema is used](#when-schema-is-used-1)
+  - [How schema is reloaded](#how-schema-is-reloaded-1)
+  - [When schema is reloaded and operation is retried](#when-schema-is-reloaded-and-operation-is-retried-1)
+  - [Alternative approaches](#alternative-approaches-1)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Space schema
+
+Related links: [#98](https://github.com/tarantool/crud/issues/98),
+[PR#111](https://github.com/tarantool/crud/pull/111).
+
+### How schema is stored
+
+Every ``crud`` router is a [``vshard``](https://www.tarantool.io/en/doc/latest/reference/reference_rock/vshard/)
+router, the same applies to storages. In ``vshard`` clusters, spaces are
+created on storages. Thus, each storage has a schema (space list,
+space formats and indexes) on it.
+
+Every router has a [``net.box``](https://www.tarantool.io/en/doc/latest/reference/reference_lua/net_box/)
+connection to each storage it could interact with. (They can be
+retrieved with [``vshard.router.routeall``](https://www.tarantool.io/en/doc/latest/reference/reference_rock/vshard/vshard_router/#lua-function.vshard.router.routeall)
+call.) Each ``net.box`` connection has space schema for an instance it
+is connected to. Router can access space schema by using ``net.box``
+connection object contents.
+
+### When schema is used
+
+Space schema is used
+- to flatten objects on `crud.*_object*` requests;
+- to resolve update operation fields if updating
+  by field name is not supported natively;
+- to calculate ``bucket_id`` to choose replicaset for a request
+  (together with sharding info);
+- in `metadata` field of request response, so it could be used later
+  for `crud.unflatten_rows`.
+
+### How schema is reloaded
+
+``net.box`` schema reload works as follows. For each request (we use ``call``s
+to execute our procedures) from client (router) to server (storage) we
+receive a schema version together with response data. If schema versions mismatch,
+client reloads schema. The request is not retried. ``net.box`` reloads
+schema before returning synchronous ``call`` result to a user, so a next
+request in the fiber will use the updated schema. (See [tarantool/tarantool/6169](https://github.com/tarantool/tarantool/issues/6169).)
+
+``crud`` cannot implicitly rely on ``net.box`` reloads: ``crud`` requests
+use ``net.box`` spaces schema to build ``net.box`` ``call`` request data,
+so if something is not relevant anymore, a part of this request data need
+to be recalculated. ``crud`` uses connection ``reload_schema`` handle
+(see [PR#111 comment](https://github.com/tarantool/crud/pull/111#issuecomment-765811556))
+to ping the storage and updates `net.box` schema if it is outdated. ``crud``
+reloads each replicaset schema. If there are several requests to reload info,
+the only one reload fiber is started and every request waits for its completion.
+
+### When schema is reloaded and operation is retried
+
+The basic logic is as follows: if something had failed and space schema
+mismatch could be the reason, reload the schema and retry. If it didn't
+help after ``N`` retries (now ``N`` is ``1``), pass the error to the
+user.
+
+Retry with reload is triggered
+- before a network request if an action that depends on the schema has failed:
+  - space not found;
+  - object flatten on `crud.*_object*` request has failed;
+  - ``bucket_id`` calculation has failed due to any reason;
+  - if updating by field name is not supported natively and id resolve
+    has failed;
+- network operation had failed on storage, hash check is on and hashes mismatch.
+
+Let's talk a bit more about the last one. To enable hash check, the user
+should pass `add_space_schema_hash` option to a request. This option is
+always enabled for `crud.*_object*` requests. If hashes mismatch, it
+means the router schema of this space is inconsistent with the storage
+space schema, so we reload it. For ``*_many`` methods, reload and retry
+happens only if all tuples had failed with hash mismatch; otherwise,
+errors are passed to a user.
+
+Retries with reload are processed by wrappers that wraps "pure" crud operations
+and utilities. Retries are counted per function, so in fact there could be more
+reloads than ``N`` for a single request. For example, `crud.*_object*` code looks
+like this:
+```lua
+local function reload_schema_wrapper(func, ...)
+    for i = 1, N do
+        local status, res, err, reload_needed = pcall(func, ...)
+        if err ~= nil and reload_needed then
+            process_reload()
+            -- ...
+        end
+    end
+end
+
+function crud.insert_object(args)
+    local flatten_obj, err = reload_schema_wrapper(flatten_obj_func, space_name, obj)
+    -- ...
+    return reload_schema_wrapper(call_insert_func, space_name, flatten_obj, opts)
+end
+```
+
+For `crud.*_object_many`, each tuple flatten is retried separately (though it
+is hard to imagine a real case of multiple successful flatten retries):
+```lua
+function crud.insert_object_many(args)
+    for _, obj in ipairs(obj_array) do
+        local flatten_obj, err = reload_schema_wrapper(flatten_obj_func, space_name, obj)
+        -- ...
+    end
+    -- ...
+    return reload_schema_wrapper(call_insert_many_func, space_name, flatten_obj_array, opts)
+end
+```
+
+### Alternative approaches
+
+One of the alternatives considered was to ping a storage instance on
+each request to refresh schema (see [PR#111 comment](https://github.com/tarantool/crud/pull/111#discussion_r562757016)),
+but it was declined due to performance degradation.
+
+
+## Sharding schema
+
+Related links: [#166](https://github.com/tarantool/crud/issues/166),
+[PR#181](https://github.com/tarantool/crud/pull/181),
+[#237](https://github.com/tarantool/crud/issues/237),
+[PR#239](https://github.com/tarantool/crud/pull/239),
+[#212](https://github.com/tarantool/crud/issues/212),
+[PR#268](https://github.com/tarantool/crud/pull/268).
+
+### How schema is stored
+
+Again, ``crud`` cluster is a [``vshard``](https://www.tarantool.io/en/doc/latest/reference/reference_rock/vshard/)
+cluster. Thus, data is sharded based on some key. To extract the key
+from a tuple and compute a [``bucket_id``](https://www.tarantool.io/en/doc/latest/reference/reference_rock/vshard/vshard_architecture/),
+we use [``ddl``](https://github.com/tarantool/ddl) module data.
+``ddl`` module schema describes sharding key (how to extract data
+required to compute ``bucket_id`` from a tuple based on space schema)
+and sharding function (how to compute ``bucket_id`` based on the data).
+This information is stored on storages in some tables and Tarantool
+spaces: ``_ddl_sharding_key`` and ``_ddl_sharding_func``. ``crud``
+module uses `_ddl_sharding_key` and `_ddl_sharding_func` spaces to fetch
+sharding schema: thus, you don't obliged to use ``ddl`` module and can
+setup only ``_ddl_*`` spaces manually, if you want. ``crud`` module uses
+plain Lua tables to store sharding info on routers and storages.
+
+### When schema is used
+
+Sharding schema is used
+- to compute ``bucket_id``. Thus, it is used each time we need to execute
+  a non-map-reduce request[^1] and `bucket_id` is not specified by user.
+  If there is no sharding schema specified, we use defaults: sharding key
+  is primary key, sharding function is ``vshard.router.bucket_id_strcrc32``.
+
+[^1]: It includes all ``insert_*``, ``replace_*``, ``update_*``,
+      ``delete_*``, ``upsert_*``, ``get_*`` requests. ``*_many`` requests
+      also use sharding schema to find a storage for each tuple. ``select``,
+      ``count`` and ``pairs`` requests use sharding info if user conditions
+      have equal condition for a sharding key. (User still can force map-reduce
+      with `force_map_call`. In this case sharding schema won't be used.)
+
+### How schema is reloaded
+
+Storage sharding schema (internal ``crud`` Lua tables) is updated on
+initialization (first time when info is requested) and each time someone
+changes ``_ddl_sharding_key`` or ``_ddl_sharding_func`` data — we use
+``on_replace`` triggers.
+
+Routers fetch sharding schema if cache wasn't initialized yet or each
+time reload was requested. Reload could be requested with ``crud``
+itself (see below) or by user (with ``require('crud.common.sharding_key').update_cache()``
+or ``require('crud.common.sharding_func').update_cache()`` handles).
+The handles was deprecated after introducing automatic reload.
+
+The sharding information reload procedure always fetches all sharding keys
+and all sharding functions disregarding of a reason that triggers the reload.
+If there are several requests to reload info, the only one
+reload fiber is started and every request waits for its completion.
+
+### When schema is reloaded and operation is retried
+
+Retry with reload is triggered
+- if router and storage schema hashes mismatch on request that
+  uses sharding info.
+
+Each request that uses sharding info passes sharding hashes from
+a router with a request. If hashes mismatch with storage ones, we return
+a specific error. The request retries if it receives a hash mismatch
+error from the storage. If it didn't help after ``N`` retries (now ``N``
+is ``1``), the error is passed to the user. For ``*_many`` methods,
+reload and retry happens only if all tuples had failed with hash
+mismatch; otherwise, errors are passed to the user.
+
+Retries with reload are processed by wrappers that wraps "pure" crud
+operations, same as in space schema reload. Since there is only one
+use case for sharding info, there is only one wrapper per operation:
+around the main "pure" crud function with network call.
+
+### Alternative approaches
+
+There were different implementations of working with hash: we tried
+to compute it instead of storing pre-computed values (both on router
+and storage), but pre-computed approach with triggers was better in
+terms of performance. It was also an option to ping storage before
+sending a request and verify sharding info relevance before sending
+the request with separate call, but it was also declined due to
+performance degradation.


AnaNek (approved):

doc/dev/schema.md
```diff
@@ -0,0 +1,174 @@
...
+### When schema is used
+
+Space schema is used to flatten maps on `crud.*_object*` requests.
+Together with sharding info, the space schema is used to calculate
+``bucket_id`` to choose replicaset for a request.
+
+### How schema is reloaded
+
+Now there are no automatic reload of space schema supported by
+``net.box`` itself (see [tarantool/tarantool/6169](https://github.com/tarantool/tarantool/issues/6169)).
```
Totktonada, line 56: It is not fully correct. AFAIR, net.box performs a request and if a schema version in the response lower than the schema version on the client, it starts schema reloading in background (I don't remember, whether other requests will wait for the new schema). However it is true that net.box does not repeat the request after the schema reloading: so it is possible that space/index IDs will be resolved from names using an old schema (client's one), but the request will be performed against the new schema.

doc/dev/schema.md
```diff
@@ -0,0 +1,174 @@
...
+Retry with reload is triggered if
+- space not found;
+- ``bucket_id`` calculation has failed due to any reason;
+- operation had failed on storage, hash check is on and hashes mismatch.
+
+Let's talk a bit more about the last one. To enable hash check, the user
+should pass `add_space_schema_hash` option to a request. This option is
+always enabled for `crud.*_object*` requests. If hashes mismatch, it
+means the router space schema is inconsistent with the storage space
+schema, so we reload it. For ``*_many`` methods, reload and retry
```
Totktonada, line 79: I would highlight that it is for a particular space:

> the router schema of *this* space

doc/dev/schema.md
```diff
@@ -0,0 +1,174 @@
...
+happens only if all tuples had failed with hash mismatch; otherwise,
+errors are passed to a user.
+
+Retries are counted per function it wraps, so in fact there could
+be more reloads than ``N`` for a single request. For example, if a user
+calls `insert_object`, object flatten has failed (and has succeeded
+after reload) and then storage insert has failed and hashes mismatch,
```
Totktonada, line 86: The flatten fail is not mentioned above as a reason to trigger schema reload for the space.

doc/dev/schema.md
```diff
@@ -0,0 +1,174 @@
...
+
+Let's talk a bit more about the last one. To enable hash check, the user
+should pass `add_space_schema_hash` option to a request. This option is
+always enabled for `crud.*_object*` requests. If hashes mismatch, it
+means the router space schema is inconsistent with the storage space
+schema, so we reload it. For ``*_many`` methods, reload and retry
+happens only if all tuples had failed with hash mismatch; otherwise,
+errors are passed to a user.
+
+Retries are counted per function it wraps, so in fact there could
```
Totktonada, line 83: It seems, some words about the wrapping are needed before this paragraph. I understand what do you meant, because I seen this code, but I don't remember precise set of such wrappers. It would be good to list them.

doc/dev/schema.md
```diff
@@ -0,0 +1,174 @@
...
+calls `insert_object_many`, each tuple flatten is wrapped independently
+(though it is hard to imagine a real case of multiple flatten retries:
+user will need to pass a batch of tables with different structure and
+change the schema several times to match the current one; in reality it
+always will be success after a single retry or a fast fail).
+
+### Alternative approaches
+
+One of the alternatives considered was to ping a storage instance on
+each request to refresh schema (see [PR#111 comment](https://github.com/tarantool/crud/pull/111#discussion_r562757016)).
```
Totktonada, line 98: It worth to briefly say, why an alternative was declined.

doc/dev/schema.md
```diff
@@ -0,0 +1,174 @@
...
+[PR#239](https://github.com/tarantool/crud/pull/239),
+[#212](https://github.com/tarantool/crud/issues/212),
+[PR#268](https://github.com/tarantool/crud/pull/268).
+
+### How schema is stored
+
+Again, ``crud`` cluster is a [``vshard``](https://www.tarantool.io/en/doc/latest/reference/reference_rock/vshard/)
+cluster. Thus, data is sharded based on some key. To extract the key
+from a tuple and compute a [``bucket_id``](https://www.tarantool.io/en/doc/latest/reference/reference_rock/vshard/vshard_architecture/),
+[``ddl``](https://github.com/tarantool/ddl) module is used. ``ddl``
```
Totktonada, line 115: It sounds like 'crud uses a function from ddl to compute bucket_id', a bit confusing.

doc/dev/schema.md
```diff
@@ -0,0 +1,174 @@
...
+setup only ``_ddl_*`` spaces manually, if you want. ``crud`` module uses
+plain Lua tables to store sharding info on routers and storages.
+
+### When schema is used
+
+Sharding schema is used to compute ``bucket_id``. Thus, it is used each
+time we need to execute a non-map-reduce request (``insert_*``,
+``replace_*``, ``update_*``, ``delete_*``,  ``upsert_*``, ``get_*``
+requests, most ``select`` and ``count`` requests, some ``pairs``
+requests), ``bucket_id`` is not specified explicitly and there are no
```
Totktonada, line 132: are -> is

doc/dev/schema.md
```diff
@@ -0,0 +1,174 @@
...
+Storage sharding schema (internal ``crud`` Lua tables) is updated on
+initialization (first time when info is requested) and each time someone
+changes ``_ddl_sharding_key`` or ``_ddl_sharding_func`` data -- we use
+``on_replace`` triggers.
+
+Routers fetch sharding schema if cache wasn't initialized yet or each
+time reload was requested. Reload could be requested with ``crud``
+itself (see below) or by user (with ``require('crud.common.sharding_key').update_cache()``
+or ``require('crud.common.sharding_func').update_cache()`` handles).
+The latter was deprecated after introducing automatic reload.
```
Totktonada, line 148: 'The latter' may be interpreted as the second function, so I would say 'the handles'.

doc/dev/schema.md
DifferentialOrange: But space retrieval is not a request, thus if we use only net_box.space or net_box.call, it is not refreshes? Or it refreshes for calls?

doc/dev/schema.md
DifferentialOrange: Changed

doc/dev/schema.md
DifferentialOrange: Fixed

doc/dev/schema.md
DifferentialOrange: Added some description

doc/dev/schema.md
DifferentialOrange: Mentioned

doc/dev/schema.md
DifferentialOrange: Reworked

doc/dev/schema.md
DifferentialOrange: Fixed

doc/dev/schema.md
DifferentialOrange: Fixed

doc/dev/schema.md
```diff
@@ -0,0 +1,174 @@
...
+
+It is worth noting that reload/initialization affects both sharding
+key and function caches at the same time, but fetch itself is triggered
+only if one cache requested by some function is not initialized yet.
```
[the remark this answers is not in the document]
DifferentialOrange: Fixed

doc/dev/schema.md
Totktonada: > But space retrieval is not a request, thus if we use only net_box.space or net_box.call, it is not refreshes?

It is correct. Accessing a net.box space object or its fields does not issue a request (so it is fast) and does not trigger the schema reload.

> Or it refreshes for calls?

For any request except authorization (and except internal schema fetching requests). It includes a call.

But it is important that the new schema will be 'seen' only for future requests (I'm not sure whether it assumes all requests issued after given one or all issued after the new schema will arrive). The request, which triggers the schema reload, will use old space/index/collation IDs.

DifferentialOrange: I've tried to summarize our discussion to the document.

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+to execute our procedures) from client (router) to server (storage) we
+receive a schema version together with response data. If schema versions mismatch,
+client reloads schema. The request is not retried. ``net.box`` reloads
+schema before returning ``call`` result to a user, so the next request will use
+updated schema. (See [tarantool/tarantool/6169](https://github.com/tarantool/tarantool/issues/6169)).
+``crud`` cannot implicitly rely on ``net.box`` reloads: ``crud`` requests
+use ``net.box`` spaces schema to build ``net.box`` ``call`` request data,
+so if something is not relevant anymore, a part of this request data need
+to be recalculated. ``crud`` uses connection ``reload_schema`` handle
+(``ping`` wrapper, see[PR#111 comment](https://github.com/tarantool/crud/pull/111#issuecomment-765811556))
```
Totktonada, line 69: Typo: No space before a link.

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+for `crud.unflatten_rows`.
+
+### How schema is reloaded
+
+``net.box`` schema reload works as follows. For each request (we use ``call``s
+to execute our procedures) from client (router) to server (storage) we
+receive a schema version together with response data. If schema versions mismatch,
+client reloads schema. The request is not retried. ``net.box`` reloads
+schema before returning ``call`` result to a user, so the next request will use
+updated schema. (See [tarantool/tarantool/6169](https://github.com/tarantool/tarantool/issues/6169)).
```
Totktonada, line 64: 1. Next requests *in this fiber*.
2. If the request is synchronous.

Both points are my guesses about the implementation (it is likely so, but I don't check it specifically).

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+``crud`` cannot implicitly rely on ``net.box`` reloads: ``crud`` requests
```
Totktonada, line 65: I suggest to split `net.box` and `crud` parts into its own paragraphs (just add an empty line before this line).

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+use ``net.box`` spaces schema to build ``net.box`` ``call`` request data,
+so if something is not relevant anymore, a part of this request data need
+to be recalculated. ``crud`` uses connection ``reload_schema`` handle
+(``ping`` wrapper, see[PR#111 comment](https://github.com/tarantool/crud/pull/111#issuecomment-765811556))
+to reload schema explicitly when needed. ``crud`` reloads each replicaset
```
Totktonada, line 70: Don't know, whether it should be written here or it is out of scope of this document, but we can add that `reload_schema()` always sends a ping request (and waits for a response), but not necessarily reloads the schema (content of `_vspace`, `_vindex` and `vcollation` system views). It may give a good guess about a cost of this function.

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+
+### When schema is used
+
+Space schema is used to flatten maps on `crud.*_object*` requests.
+Space schema is used to resolve update operation fields if updating
+by field is not supported natively. Together with sharding info,
+the space schema is used to calculate ``bucket_id`` to choose
+replicaset for a request. Finally, schema is returned to a user
+in `metadata` field of request response, so it could be used later
+for `crud.unflatten_rows`.
```
Totktonada, line 55: Nit: I think it'll look more structured if will be formatted as a list.

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+fiber is started and every request waits for its completion.
+
+### When schema is reloaded and operation is retried
+
+The basic logic is as follows: if something had failed and space schema
+mismatch could be the reason, reload the schema and retry. If it didn't
+help after ``N`` retries (now ``N`` is ``1``), pass the error to the
+user.
+
+Retry with reload is triggered if
```
Totktonada, line 81: If I'll generalize it in the following way, will it be correct?

> Retry with reload is triggered
>
> - before a network request if an action that depends on the schema has failed:
>   - space not found;
>   - <...> flatten <...>;
>   - <...> `bucket_id` <...>;
> - operation had failed on storage, hash check is on and hashes mismatch.

If actions that depend on the schema, but failed without schema reload, exist, this generalization is incorrect. However if it is correct, we'll have a good simple rule what should trigger the reload.

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+always enabled for `crud.*_object*` requests. If hashes mismatch, it
+means the router schema of this space is inconsistent with the storage
+space schema, so we reload it. For ``*_many`` methods, reload and retry
+happens only if all tuples had failed with hash mismatch; otherwise,
+errors are passed to a user.
+
+Retries with reload are processed with wrappers. Wrapper is a Lua
+function that wraps the call of some business logic Lua function (for
+example, a function that transforms `crud.insert_object` map to a tuple-like
+table or a function that performs insert request with this tuple-like table).
```
Totktonada, line 98: The explanation in the parentheses is redundant IMHO. I mean, the word 'wrapper' is quite usual in programming, but the explanation requires to understand several not very obvious things (what is '`crud.insert_object` map', what is 'a function that performs insert request').

Also the definition 'a wrapper is <...> that wraps' uses a term to explain itself. I would say: a wrapper of a function X is a function Y, which makes semantically same job as X, which calls X, and adds to X's execution something -- preprocessing of the inputs, postprocessing of the result, retrying. However I think it is redundant.

What is worth to say here: The reload-schema-and-retry logic is implemented as a wrapper around 'pure' `crud` request (insert, get, ...).

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+Depending on return values, it decides to call some reload util and retry
+the function call or pass the error value to the user. Retries are counted
+per function it wraps, so in fact there could be more reloads than ``N``
+for a single request. For example, if a user calls `insert_object`, object
+flatten has failed (and has succeeded after reload) and then storage insert
+has failed and hashes mismatch, reload will happen twice and there will be
+two retries (though it's two different blocks of code will be retried
+independently). Or if a user calls `insert_object_many`, each tuple
```
Totktonada, line 106: To be honest, I completely missed the point. I understand that when we have two-level retrying wrappers, we can get N*M retries at max. But here we don't say anything about sharding info wrappers. And we can't get more than N retries if the logic is like so:

```lua
local function pure_crud_request()
    <...>
end

local function crud_request_with_schema_reload()
    while i = 1, N do
        pure_crud_request()
        if <..ok..> then return <..result..>
        if <..schema should be reloaded..> then continue end
        return <..error..>
    end
    return <..last error..>
end
```

Maybe it worth to illustrate the words with a pseudocode?

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+flatten is wrapped independently (though it is hard to imagine a real
+case of multiple flatten retries: user will need to pass a batch of
+tables with different structure and change the schema several times
+to match the current one; in reality it always will be success after
+a single retry or a fast fail).
```
Totktonada, line 111: It is cryptic for me too. I guess a pseudocode would help here, but I'll share what confuses me:

* We perform a request against a schema of version V1, then, if flatten fails, we guess that it is because the schema now on the version V2 and we hope that schema version V3 will help to flatten result? It seems counter-intuitive.
* What are 'tables with different structure'? Where I (a user) should pass it?
* Why should I (a user) need to 'change the schema several times'? To observe some effect? It is not clear from this text.

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+setup only ``_ddl_*`` spaces manually, if you want. ``crud`` module uses
+plain Lua tables to store sharding info on routers and storages.
+
+### When schema is used
+
+Sharding schema is used to compute ``bucket_id``. Thus, it is used each
+time we need to execute a non-map-reduce request (``insert_*``,
+``replace_*``, ``update_*``, ``delete_*``,  ``upsert_*``, ``get_*``
+requests, most ``select`` and ``count`` requests, some ``pairs``
+requests), ``bucket_id`` is not specified explicitly and there is no
```
Totktonada, line 151: If you have a three-line out-of-flow inline comment, it likely will look better as a [footnote](https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/) rather than as a clause in parentheses.

Totktonada: > Thus, it is used each time we need to execute a non-map-reduce request, (?) ``bucket_id`` is not specified explicitly and there is no ``force_map_call`` option.

(Removed the clause in the parentheses.)

I guess something like 'IOW, when' is missed in the place, which I marked with (?).

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+time reload was requested. Reload could be requested with ``crud``
+itself (see below) or by user (with ``require('crud.common.sharding_key').update_cache()``
+or ``require('crud.common.sharding_func').update_cache()`` handles).
+The handles was deprecated after introducing automatic reload.
+
+It is worth noting that reload/initialization affects both sharding
+key and function caches at the same time, but fetch itself is triggered
+even if one cache requested by some function is not initialized yet.
+(I.e. if function cache is ok and key cache is not, they still both be
+reloaded.). If there are several requests to reload info, the only one
```
Totktonada, line 173: I think we can say it simpler. For example:

> Thes harding information reload procedure always fetches all sharding keys and all sharding functions disregarding of a reason that triggers the reload.

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+Sharding schema is used to compute ``bucket_id``. Thus, it is used each
+time we need to execute a non-map-reduce request (``insert_*``,
+``replace_*``, ``update_*``, ``delete_*``,  ``upsert_*``, ``get_*``
+requests, most ``select`` and ``count`` requests, some ``pairs``
+requests), ``bucket_id`` is not specified explicitly and there is no
+``force_map_call`` option. If there is no sharding schema specified, we
+use defaults: sharding key is primary key, sharding function is
+``vshard.router.bucket_id_strcrc32``.
+
+### How schema is reloaded
```
Totktonada, line 156: Shouldn't we say here something about wrappers? AFAIR, it is implemented similarly to how space format/indexes are reloaded.

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+error from the storage. If it didn't help after ``N`` retries (now ``N``
+is ``1``), the error is passed to the user. For ``*_many`` methods,
+reload and retry happens only if all tuples had failed with hash
+mismatch; otherwise, errors are passed to the user.
+
+### Alternative approaches
+
+There were different implementations of working with hash: we tried
+to compute it instead of storing pre-computed values, but pre-computed
+approach with triggers was better in terms of performance. It was also
```
Totktonada, line 190: On storage? On router? Both?

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+an option to ping storage before sending a request and verify sharding
+info relevance before sending the request with separate call, but it was
+discarded.
```
Totktonada, line 193: Why?

Totktonada: I mean, I understand why, but the alternatives are listed to explain why an approach that may look simpler/better was not choosen. If it is so obvious to don't mention, it is not an alternative and should not be written here. If it is an alternative, it worth to say why it is worse than choosen approach.

doc/dev/schema.md
```diff
@@ -0,0 +1,193 @@
...
+The handles was deprecated after introducing automatic reload.
+
+It is worth noting that reload/initialization affects both sharding
+key and function caches at the same time, but fetch itself is triggered
+even if one cache requested by some function is not initialized yet.
+(I.e. if function cache is ok and key cache is not, they still both be
+reloaded.). If there are several requests to reload info, the only one
+reload fiber is started and every request waits for its completion.
+
+### When schema is reloaded and operation is retried
```
Totktonada, line 176: Here we have only one situation that triggers the reload-and-retry? Nothing like 'separate wrappers for flattens in a `*_many` operation' as in the 'Space schema' section?

Anyway, even if we have the only situation that triggers it, I would format is as a list. This way it is more obvious that it is really the one.

Totktonada (changes requested):
Thanks for the updates! You have added more details and now not everything looks obvious for me. So I take back my approve back and will continue to look into the document with you.

doc/dev/schema.md
DifferentialOrange: But next request in another fiber will also use updated schema since there is the only one net.box schema object.

doc/dev/schema.md
DifferentialOrange: But we rely on `reload_schema` to actually reload schema -- there are no explicit checks on `_vspace` copy and etc.

doc/dev/schema.md
DifferentialOrange: Since every object flatten is retried independently it is **theoretically** possible that retry may happen for each object. To actually trigger this, the input should be as follows
```lua
{
    {--object with schema v2},
    {--object with schema v3},
    {--object with schema v4},
    -- ...
}
```

doc/dev/schema.md
DifferentialOrange: Fixed

doc/dev/schema.md
DifferentialOrange: Added

doc/dev/schema.md
DifferentialOrange: Thank you for the idea, reworked

doc/dev/schema.md
DifferentialOrange: Thanks, reworked to two-level list

doc/dev/schema.md
DifferentialOrange: Reworked

doc/dev/schema.md
DifferentialOrange: Actually it is not as you expected. Reworked with code example, hope it will be enough to understand.

doc/dev/schema.md
DifferentialOrange: I think it's better to remove an explanation on why multiple success reloads is not possible. If one really will be interested to prove that everything should be alright, he'll quickly solve the exercise by themselves.

doc/dev/schema.md
DifferentialOrange: Used a footnote, thank you.

doc/dev/schema.md
DifferentialOrange: Reworked

doc/dev/schema.md
DifferentialOrange: Both

doc/dev/schema.md
DifferentialOrange: Added to the same section as in "space schema".

doc/dev/schema.md
DifferentialOrange: Well, I've tried to format it as a list.

doc/dev/schema.md
DifferentialOrange: Added a reason

doc/dev/schema.md
Totktonada: Here you describe `net.box` behaviour, so it would be good it make this description correct and complete.

DifferentialOrange: So if we use asynchronous request, it doesn't reload the schema if user fetches the result?

Totktonada: I'll cite current text:

> ``net.box`` reloads schema before returning ``call`` result to a user, so the next request will use updated schema.

Let's consider an example:

```lua
conn:call('foo', {<..foo args..>}, {is_async = true})
local result = conn.space.s:select('bar', {<..select args..>})
```

According to your text if calling 'foo' reloads the schema, we'll see its effect in the select (because the `call` returns at this point). I doubt that it is guaranteed (but I didn't check the net.box's code).

Totktonada: I guess that sync request works in the same way as async request + `:wait_result()`.

doc/dev/schema.md
Totktonada: The question is not about that. I'll remove extra information from it:

`reload_schema()` always sends a ping request and waits for a response, but not necessarily reloads the schema. Should we reflect this fact it is the document?

doc/dev/schema.md
Totktonada: I got it now, thanks!

doc/dev/schema.md
Totktonada: I would like to see this clarification in the document, not just for me.

doc/dev/schema.md
DifferentialOrange: > 

well

> I'll cite current text:
> 
> > `net.box` reloads schema before returning `call` result to a user, so the next request will use updated schema.
> 
> Let's consider an example:
> 
> ```lua
> conn:call('foo', {<..foo args..>}, {is_async = true})
> local result = conn.space.s:select('bar', {<..select args..>})
> ```
> 
> According to your text if calling 'foo' reloads the schema, we'll see its effect in the select (because the `call` returns at this point). I doubt that it is guaranteed (but I didn't check the net.box's code).

Well, it seems that I need to replace `result` with something like `synchronous result`

doc/dev/schema.md
DifferentialOrange: Isn't `net.box` developers designed it in such a way that `reload_schema` reload the schema as it is stated by its name?

Totktonada: Reloads if necessary. If the schema is actual, it'll not be reloaded.

doc/dev/schema.md
DifferentialOrange: I've tried to improve the text based on this discussion.

doc/dev/schema.md
DifferentialOrange: I've tried to improve the text based on this discussion.

doc/dev/schema.md
Totktonada: Thank you!

> so the next request in a fiber will use updated schema

I think it should be:

> so **a** next request in **the** fiber will use **the** updated schema

Totktonada (approved):
Thanks for the careful work!

PR #318, opened 2022-08-12
tests: fix call options

DifferentialOrange:
`{fullscan = true}` option was supposed to be passed to request options
instead of net.box ones. Starting from Tarantool 2.10.1, net.box checks
its options and throws error if illegal parameter is passed [1].

1. https://github.com/tarantool/tarantool/commit/d7da59b4a33c9547da68431a22ee256e50a3df62

I didn't forget about

- [x] Tests
- Changelog (not needed, test update)
- Documentation (not needed, test update)

### DIFF
--- a/test/integration/count_test.lua
+++ b/test/integration/count_test.lua
@@ -766,7 +766,7 @@ pgroup.test_composite_index = function(g)
     }
 
     -- no after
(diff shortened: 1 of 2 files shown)


oleg-jukovec (approved):
Thank you for the patch!

PR #329, opened 2022-10-28
Flatten for sequence

### LINKED ISSUE #328: *_object does not support sequences
Plain tuples insert/replace/upsert works ok with sequence fields, but their *_object versions are not support them due to map flatten fail. We should consider if it is possible to improve flatten or describe that it is unsupported in doc.

DifferentialOrange:
Add `skip_nullability_check_on_flatten` option for `insert_object`, `insert_object_many`, `replace_object`, `replace_object_many`. `false` by default. By setting the option to `true` one allows setting null values to non-nullable fields, which can be useful if non-nullable field value is generated by sequence [1].

**Warning**: there is no native support of sequences in sharded systems since each replicaset has its own sequence. If sequence field is a part of the sharding key (which is true by default), choosing the bucket id is the sole responsibility of the developer (#328).

The option wasn't added to `upsert_object*` methods since upsert operation doesn't support sequence generated fields:

```lua
box.schema.space.create('seq', {
    format = {
        {name = 'payload', type = 'string'},
    },
})

box.schema.sequence.create('id', {if_not_exists = true})

box.space.seq:create_index('id',
    {
        parts = {{field = 'id'}},
        unique = true,
        sequence = 'id'
    }
)

box.space.seq:upsert({nil, 'payload'}, {{'=', 'payload', 'payload'}})
```

```
---
- error: 'Tuple field 1 (id) type does not match one required by operation: expected
    unsigned, got nil'
...
```

Refer to #328 discussion regarding solution design.

1. https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_schema_sequence/create_index/

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #328


AnaNek (approved):
LGTM.

PR #335, opened 2022-12-14
ci: add reusable testing workflow for crud module

slavakirichenko:
Add reusable testing workflow for integration
testing of crud module.

Part of tarantool/tarantool#6620

I didn't forget about

- [Tests](https://github.com/tarantool/tarantool/actions/runs/3749354667/jobs/6367850717)
- [Tested negatively here](https://github.com/tarantool/tarantool/actions/runs/3749505598/jobs/6368187258)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,12 @@ add_custom_target(luatest
   COMMENT "Run regression tests"
 )
 
+add_custom_target(luatest-no-coverage
+  COMMAND ${LUATEST} -v
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  COMMENT "Run regression tests without coverage"
+)
+
 set(PERFORMANCE_TESTS_SUBDIR "test/performance")
 
 add_custom_target(performance
--- a/.github/workflows/reusable_test.yml
+++ b/.github/workflows/reusable_test.yml
@@ -1,0 +1,47 @@
+name: Reusable test
+
+on:
+  workflow_call:
+    inputs:
+      artifact_name:
+        description: The name of the Tarantool build artifact
+        default: ubuntu-focal
+        required: false
+        type: string
+
+jobs:
+  run_tests:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Clone the crud module
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ github.repository_owner }}/crud
+
+      - name: Download the Tarantool build artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ inputs.artifact_name }}
+
+      # All dependencies for tarantool are already installed in Ubuntu 20.04.
+      # Check package dependencies when migrating to other OS version.
+      - name: Install Tarantool
+        run: |
+          sudo dpkg -i tarantool_*.deb tarantool-common_*.deb tarantool-dev_*.deb
+          tarantool --version
+
+      - name: Install requirements
+        run: ./deps.sh
+
+      - name: Install metrics
+        run: tarantoolctl rocks install metrics 0.12.0
+
+      # This server starts and listen on 8084 port that is used for tests
+      - name: Stop Mono server
+        run: sudo kill -9 $(sudo lsof -t -i tcp:8084) || true
+
+      - run: cmake -S . -B build
+
+      - name: Run regression tests
+        run: make -C build luatest-no-coverage


.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+          ref: 'slavakirichenko/add-reusable-testing-workflow'
+
+      - name: Download the Tarantool build artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ inputs.artifact_name }}
+
+      - name: Install Tarantool
+        # Now we're lucky: all dependencies are already installed. Check package
+        # dependencies when migrating to other OS version.
```
NickVolynkin, line 32: What does this mean? Does this workflow check migration?

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: tarantoolctl rocks install metrics ${{ matrix.metrics-version }}
+
+      # This server starts and listen on 8084 port that is used for tests
+      - name: Stop Mono server
+        run: sudo kill -9 $(sudo lsof -t -i tcp:8084) || true
+
+      - run: cmake -S . -B build
+
+      - name: Run tests and code coverage analysis
+        run: make -C build coverage
```
NickVolynkin, line 52: We don't need coverage in an integration test. Is there a way to run tests without coverage?

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
+name: reusable_testing
+
+on:
+  workflow_call:
+    inputs:
+      artifact_name:
+        description: The name of the Tarantool build artifact
+        default: ubuntu-focal
+        required: false
```
NickVolynkin, line 9: Is `ubuntu-focal` a valid artifact name? It seems that it isn't.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+          tarantool --version
+          ./deps.sh
+
+      - run: echo $PWD/.rocks/bin >> $GITHUB_PATH
+
+      - name: Install metrics
+        run: tarantoolctl rocks install metrics ${{ matrix.metrics-version }}
+
+      # This server starts and listen on 8084 port that is used for tests
+      - name: Stop Mono server
```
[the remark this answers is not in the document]
DifferentialOrange: On GitHub Ubuntu 20.04 runner

.github/workflows/reusable_testing.yml
DifferentialOrange: As far as I understand, there is no CMake target to run tests without coverage. But you may add one, it shouldn't be too complicated.

.github/workflows/reusable_testing.yml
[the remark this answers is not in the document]
NickVolynkin: Weird, but okay. Is it this mono: https://www.mono-project.com/docs/web/aspnet/?

.github/workflows/reusable_testing.yml
NickVolynkin: Running tests without coverage is usually faster than with it. @slavakirichenko could you please add a new target?

.github/workflows/reusable_testing.yml
NickVolynkin: Indeed, it's the same as in other similar workflows. Okay. https://github.com/tarantool/expirationd/blob/master/.github/workflows/reusable_testing.yml

.github/workflows/reusable_testing.yml
NickVolynkin: Try this:
```cmake
add_custom_target(test
  COMMAND ${LUATEST} -v
  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
  COMMENT "Run regression tests without coverage"
)
```

CMakeLists.txt
```diff
@@ -68,6 +68,12 @@ add_custom_target(coveralls
   COMMENT "Send code coverage data to the coveralls.io service"
 )
 
+add_custom_target(test
```
NickVolynkin, line 71: Let's add this just below the `luatest` target. And on second thought, the name should be something similar, like `luatest-no-coverage`.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,50 @@
...
+  run_tests:
+    strategy:
+      matrix:
+        metrics-version: [ "0.12.0" ]
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone the crud module
+        uses: actions/checkout@v3
+        with:
+          repository: slavakirichenko/crud
```
NickVolynkin, line 22: change to main repo before merging

NickVolynkin: I was wrong )

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,49 @@
...
+        metrics-version: [ "0.12.0" ]
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone the crud module
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ github.repository_owner }}/crud
+
+      - name: Download the Tarantool build artifact
+        uses: actions/download-artifact@v2
```
ylobankov, line 25: Looks like we should use v3.

CMakeLists.txt
```diff
@@ -36,6 +36,12 @@ add_custom_target(luatest
   COMMENT "Run regression tests"
 )
 
+add_custom_target(luatest-no-coverage
```
ylobankov, line 39: @DifferentialOrange @slavakirichenko Shouldn't we have make targets like these?

- luatest
- luatest-coverage (or something like this)

I think it is more logical and convenient.

DifferentialOrange: It sounds better, but let's not break backward compatibility yet (for example, we may rename the target to `luatest-coverage` and add an alias if CMake can do so)

ylobankov: Actually, I am ok with the current approach, it was just a question.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,49 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ github.repository_owner }}/crud
+
+      - name: Download the Tarantool build artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ inputs.artifact_name }}
+
+      - name: Install Tarantool
```
ylobankov, line 29: Please add this comment before this step.

```
# Now we're lucky: all dependencies are already installed. Check package
# dependencies when migrating to other OS version.
```

.github/workflows/reusable_testing.yml
ylobankov: It means that we are lucky with dependencies for now. Pay attention, we use `dpkg` here, not `apt`. If we run this workflow on `ubuntu-22.04`, we may get an error about unresolved dependencies. As far as I know, since version 1.1 `apt` can install local packages via `sudo apt install ./foo-1.2.3.deb`. Probably we can do it here.

ylobankov: But I am ok if you leave `dpkg` here.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  run_tests:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Clone the crud module
+        uses: actions/checkout@v3
+        with:
+          repository: slavakirichenko/crud
```
DifferentialOrange, line 20: So... when this one will be replaced with the real one?

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,48 @@
...
+          sudo dpkg -i tarantool_*.deb tarantool-common_*.deb tarantool-dev_*.deb
+          tarantool --version
+
+      - name: Install requirements
+        run: ./deps.sh
+
+      - name: Install metrics
+        run: tarantoolctl rocks install metrics 0.12.0
+
+      - run: echo $PWD/.rocks/bin >> $GITHUB_PATH
```
DifferentialOrange, line 39: ?

CMakeLists.txt
DifferentialOrange: Yeah, let's at least use `luatest` here.

.github/workflows/reusable_testing.yml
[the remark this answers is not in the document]
DifferentialOrange: I think so

.github/workflows/reusable_testing.yml
NickVolynkin: @ylobankov what does it mean? How are they already installed if we install them in the next step?

NickVolynkin:
```suggestion
      # All dependencies for tarantool are already installed in Ubuntu 20.04.
      # Check package dependencies when migrating to other OS version.
      - name: Install Tarantool
```

.github/workflows/reusable_testing.yml
NickVolynkin: There's no such step in the original workflow https://github.com/tarantool/crud/blob/master/.github/workflows/check_on_push.yaml

.github/workflows/reusable_testing.yml
slavakirichenko: Fixed.

.github/workflows/reusable_testing.yml
slavakirichenko: Fixed.

.github/workflows/reusable_testing.yml
slavakirichenko: Fixed.

NickVolynkin (approved):
LGTM

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  run_tests:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Clone the crud module
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ github.repository_owner }}/crud
```
DifferentialOrange, line 20: What's this one for?

CMakeLists.txt
DifferentialOrange: I mean `luatest` instead of `luatest-no-coverage`

slavakirichenko: @DifferentialOrange but we alerady have one with coverage at CMakeLists.txr add_custom_target(**luatest**

DifferentialOrange: Ok, sorry. Let's leave it as is for now

DifferentialOrange:
@slavakirichenko you have unresolved comments

.github/workflows/reusable_testing.yml
slavakirichenko: This one to checkout crud module like @ylobankov did at [expirationd](https://github.com/tarantool/expirationd/blob/3bbc7fdb9f53d1de01457a07a093f313e2752135/.github/workflows/reusable_testing.yml#L19)

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,47 @@
+name: reusable_testing
```
ylobankov, line 1: I would prefer to keep consistency in such names around all workflows in the repo. Looks like it should be `Reusable test`. Also, I would name the workflow file `reusable_test.yml` as we have `test_on_push.yml`, not `testing_on_push.yml`.

slavakirichenko: Fixed.

.github/workflows/reusable_testing.yml
DifferentialOrange: Is it a v3 thing? There is no any additional args with v2 checkout

.github/workflows/reusable_testing.yml
ylobankov: Almost there, but not. `reusable_test` -> `Reusable test`.

slavakirichenko: Fixed.

ylobankov (approved):
LGTM.

.github/workflows/reusable_testing.yml
ylobankov: It is a limitation of reusable workflows. Reusable workflows don't have the own contexts, so if you call this workflow inside `tarantool/tarantool`, this step will check out the `tarantool/tarantool` repo instead of `tarantool/crud`. That's why we need the repository here.

DifferentialOrange (approved):

.github/workflows/reusable_testing.yml
ylobankov: From GitHub [docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#overview):

>If you reuse a workflow from a different repository, any actions in the called workflow run as if they were part of the caller workflow. For example, if the called workflow uses `actions/checkout`, the action checks out the contents of the repository that hosts the caller workflow, not the called workflow.

>When a reusable workflow is triggered by a caller workflow, the `github` context is always associated with the caller workflow. The called workflow is automatically granted access to `github.token` and `secrets.GITHUB_TOKEN`. 
```

PR #336, opened 2022-12-28
internal: master conn check for get space utils

### LINKED ISSUE #331: Unclear error if no leader
**Problem**:
if
1) failover is disabled
2) a replicaset master dies
3) a router is restarted
4) you then try to execute any crud operation

A cryptic error is returned.

**Error**:

<details>

```lua
myapp.router> crud.select('test', nil, {first=1})
---
- null
- line: 183
  class_name: SelectError
  err: '...app/.rocks/share/tarantool/crud/common/sharding/init.lua:183: ...repro/myapp/.rocks/share/tarantool/crud/common/utils.lua:100:
    attempt to index field ''space'' (a nil value)'
  file: '...app/.rocks/share/tarantool/crud/common/sharding/init.lua'
  stack: "stack traceback:\n\t...app/.rocks/share/tarantool/crud/common/sharding/init.lua:183:
    in function <...app/.rocks/share/tarantool/crud/common/sharding/init.lua:168>\n\t[C]:
    in function 'xpcall'\n\t.../LocalCode/repro/myapp/.rocks/share/tarantool/errors.lua:145:
    in function <.../LocalCode/repro/myapp/.rocks/share/tarantool/errors.lua:139>\n\t[C]:
    in function 'pcall'\n\tbuiltin/box/console.lua:402: in function <builtin/box/console.lua:378>\n\t[string
    \"local function func(...) return require('cons...\"]:1: in main chunk\n\t[C]:
    in function 'pcall'\n\tbuiltin/box/console.lua:402: in function 'eval'\n\tbuiltin/box/console.lua:708:
    in function 'repl'\n\tbuiltin/box/console.lua:842: in function <builtin/box/console.lua:828>\n\t[C]:
    in function 'pcall'\n\tbuiltin/socket.lua:1081: in function <builtin/socket.lua:1079>"
  str: 'SelectError: ...app/.rocks/share/tarantool/crud/common/sharding/init.lua:183:
    ...repro/myapp/.rocks/share/tarantool/crud/common/utils.lua:100: attempt to index
    field ''space'' (a nil value)'
...
```
</details>

**How to reproduce**:
```bash
# any cartridge app will do
cd myapp
cartridge build
cartridge start -d
cartridge replicasets setup --bootstrap-vshard
cartridge stop s1-master
cartridge stop router && cartridge start -d router
cartridge enter router

# in console
crud.select('test', nil, {first=1})
```

**Repoducer package**:
[repro.zip](https://github.com/tarantool/crud/files/10056494/repro.zip)

**Line that triggers the error**:

https://github.com/tarantool/crud/blob/7504da391c20ac1d2dfbbb7de6d3c32d73d75f60/crud/common/utils.lua#L100

GRISHNOV:
Added validation of the connection to the `utils.get_space` method before receiving the space through the connection.

Error before patch:

```
...
str: 'SelectError: ...app/.rocks/share/tarantool/crud/common/sharding/init.lua:183:
    ...repro/myapp/.rocks/share/tarantool/crud/common/utils.lua:100: attempt to index
    field ''space'' (a nil value)'
...
```

Now:

```
...
str: 'SelectError: An error occurred during the operation: "GetSpaceError: The connection
    to the master of replicaset 548fb6ff-d686-4298-ad17-7f81fa19588c is not valid:
    connect, called on fd 40, aka 127.0.0.1:58334: Connection refused"'
...
```

Closes #331

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Added validation of the master presence in replicaset and the 
   master connection to the `utils.get_space` method before 
   receiving the space from the connection (#331).
+* Fixed fiber cancel on schema reload timeout in `call_reload_schema`.
 
 ## [0.14.1] - 10-11-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
+* Added validation of the master presence in replicaset and the 
+  master connection to the `utils.get_space` method before 
+  receiving the space from the connection (#331).
 
 ## [0.14.1] - 10-11-22
 
--- a/crud/common/schema.lua
+++ b/crud/common/schema.lua
@@ -37,7 +37,7 @@ local function call_reload_schema(replicasets)
     for _ = 1,replicasets_num do
         if channel:get(const.RELOAD_SCHEMA_TIMEOUT) == nil then
             for _, f in ipairs(fibers) do
-                if fiber:status() ~= 'dead' then
+                if f:status() ~= 'dead' then
                     f:cancel()
                 end
             end
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -15,6 +15,7 @@ local FlattenError = errors.new_class("FlattenError", {capture_stack = false})
 local UnflattenError = errors.new_class("UnflattenError", {capture_stack = false})
 local ParseOperationsError = errors.new_class('ParseOperationsError', {capture_stack = false})
 local ShardingError = errors.new_class('ShardingError', {capture_stack = false})
+local GetSpaceError = errors.new_class('GetSpaceError')
 local GetSpaceFormatError = errors.new_class('GetSpaceFormatError', {capture_stack = false})
 local FilterFieldsError = errors.new_class('FilterFieldsError', {capture_stack = false})
 local NotInitializedError = errors.new_class('NotInitialized')
@@ -97,13 +98,37 @@ end
 
 function utils.get_space(space_name, replicasets)
     local replicaset = select(2, next(replicasets))
+
+    if replicaset == nil then
+        return nil, GetSpaceError:new(
+            'The router returned empty replicasets: ' ..
+            'perhaps other instances are unavailable or you have configured only the router')
+    end
+
+    if replicaset.master == nil then
+        local error_msg = string.format(
+            'The master was not found in replicaset %s, ' ..
+            'check status of the master and repeat the operation later',
+             replicaset.uuid)
+        return nil, GetSpaceError:new(error_msg)
+    end
+
+    if replicaset.master.conn.error ~= nil then
+        local error_msg = string.format(
+            'The connection to the master of replicaset %s is not valid: %s',
+             replicaset.uuid, replicaset.master.conn.error)
+        return nil, GetSpaceError:new(error_msg)
+    end
     local space = replicaset.master.conn.space[space_name]
 
     return space
 end
 
 function utils.get_space_format(space_name, replicasets)
-    local space = utils.get_space(space_name, replicasets)
+    local space, err = utils.get_space(space_name, replicasets)
+    if err ~= nil then
+        return nil, GetSpaceFormatError:new("An error occurred during the operation: %s", err)
+    end
     if space == nil then
         return nil, GetSpaceFormatError:new("Space %q doesn't exist", space_name)
     end
--- a/crud/replace.lua
+++ b/crud/replace.lua
@@ -66,9 +66,8 @@ local function call_replace_on_router(vshard_router, space_name, original_tuple,
 
     local space, err = utils.get_space(space_name, vshard_router:routeall())
     if err ~= nil then
-        return nil, ReplaceError:new("Failed to get space %q: %s", space_name, err), const.NEED_SCHEMA_RELOAD
+        return nil, ReplaceError:new("An error occurred during the operation: %s", err), const.NEED_SCHEMA_RELOAD
     end
-
     if space == nil then
         return nil, ReplaceError:new("Space %q doesn't exist", space_name), const.NEED_SCHEMA_RELOAD
     end
--- a/crud/borders.lua
+++ b/crud/borders.lua
@@ -74,7 +74,10 @@ local function call_get_border_on_router(vshard_router, border_name, space_name,
     })
 
     local replicasets = vshard_router:routeall()
-    local space = utils.get_space(space_name, replicasets)
+    local space, err = utils.get_space(space_name, replicasets)
+    if err ~= nil then
+        return nil, BorderError:new("An error occurred during the operation: %s", err), const.NEED_SCHEMA_RELOAD
+    end
     if space == nil then
         return nil, BorderError:new("Space %q doesn't exist", space_name), const.NEED_SCHEMA_RELOAD
     end
--- a/crud/count.lua
+++ b/crud/count.lua
@@ -137,7 +137,10 @@ local function call_count_on_router(vshard_router, space_name, user_conditions, 
         return nil, CountError:new("Failed to get router replicasets: %s", err)
     end
 
-    local space = utils.get_space(space_name, replicasets)
+    local space, err = utils.get_space(space_name, replicasets)
+    if err ~= nil then
+        return nil, CountError:new("An error occurred during the operation: %s", err), const.NEED_SCHEMA_RELOAD
+    end
     if space == nil then
         return nil, CountError:new("Space %q doesn't exist", space_name), const.NEED_SCHEMA_RELOAD
     end
--- a/crud/delete.lua
+++ b/crud/delete.lua
@@ -62,7 +62,10 @@ local function call_delete_on_router(vshard_router, space_name, key, opts)
         vshard_router = '?string|table',
     })
 
-    local space = utils.get_space(space_name, vshard_router:routeall())
+    local space, err = utils.get_space(space_name, vshard_router:routeall())
+    if err ~= nil then
+        return nil, DeleteError:new("An error occurred during the operation: %s", err), const.NEED_SCHEMA_RELOAD
+    end
     if space == nil then
         return nil, DeleteError:new("Space %q doesn't exist", space_name), const.NEED_SCHEMA_RELOAD
     end
--- a/crud/get.lua
+++ b/crud/get.lua
@@ -65,7 +65,10 @@ local function call_get_on_router(vshard_router, space_name, key, opts)
         vshard_router = '?string|table',
     })
 
-    local space = utils.get_space(space_name, vshard_router:routeall())
+    local space, err = utils.get_space(space_name, vshard_router:routeall())
+    if err ~= nil then
+        return nil, GetError:new("An error occurred during the operation: %s", err), const.NEED_SCHEMA_RELOAD
+    end
     if space == nil then
         return nil, GetError:new("Space %q doesn't exist", space_name), const.NEED_SCHEMA_RELOAD
     end
--- a/crud/insert.lua
+++ b/crud/insert.lua
@@ -64,7 +64,10 @@ local function call_insert_on_router(vshard_router, space_name, original_tuple, 
         vshard_router = '?string|table',
     })
 
-    local space = utils.get_space(space_name, vshard_router:routeall())
+    local space, err = utils.get_space(space_name, vshard_router:routeall())
+    if err ~= nil then
+        return nil, InsertError:new("An error occurred during the operation: %s", err), const.NEED_SCHEMA_RELOAD
+    end
     if space == nil then
         return nil, InsertError:new("Space %q doesn't exist", space_name), const.NEED_SCHEMA_RELOAD
     end
--- a/crud/insert_many.lua
+++ b/crud/insert_many.lua
@@ -131,7 +131,12 @@ local function call_insert_many_on_router(vshard_router, space_name, original_tu
         vshard_router = '?string|table',
     })
 
-    local space = utils.get_space(space_name, vshard_router:routeall())
+    local space, err = utils.get_space(space_name, vshard_router:routeall())
+    if err ~= nil then
+        return nil, {
+            InsertManyError:new("An error occurred during the operation: %s", err)
+        }, const.NEED_SCHEMA_RELOAD
+    end
     if space == nil then
         return nil, {InsertManyError:new("Space %q doesn't exist", space_name)}, const.NEED_SCHEMA_RELOAD
     end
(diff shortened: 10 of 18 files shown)


crud/common/utils.lua
```diff
@@ -97,6 +97,12 @@ end
 
 function utils.get_space(space_name, replicasets)
     local replicaset = select(2, next(replicasets))
+
+    if replicaset.master.conn.error ~= nil then
+        local error_msg = string.format('The connection to the master is not valid: %s',
+                                         replicaset.master.conn.error)
+        error(error_msg)
+    end
```
Totktonada, line 105: According to #95 it is also possible that `replicaset.master` is `nil`. The issue is pretty old, but maybe it worth to re-check it.

GRISHNOV: Thanks! So far I've implemented a simple check `replicaset.master` is not `nil`, but it is not entirely clear to me where the timeout for this check is set from

crud/common/utils.lua
```diff
@@ -97,6 +97,20 @@ end
 
 function utils.get_space(space_name, replicasets)
```
DifferentialOrange, line 98: While expecting `utils.get_space` use cases, I've met some unsettling code: there are examples of
```lua
    local space = utils.get_space(space_name, vshard_router:routeall())
    if space == nil then
        return nil, LenError:new("Space %q doesn't exist", space_name)
    end
```
and examples of
```lua
    local space, err = utils.get_space(space_name, vshard_router:routeall())
    if err ~= nil then
        return nil, ReplaceError:new("Failed to get space %q: %s", space_name, err), const.NEED_SCHEMA_RELOAD
    end

    if space == nil then
        return nil, ReplaceError:new("Space %q doesn't exist", space_name), const.NEED_SCHEMA_RELOAD
    end
```
but `utils.get_space` never returns errors (at least in current versions). I think we may rework `utils.get_space` to work as follows:
- it's either returns `space` if everything is alright
- or it returns `nil, err` if something went wrong, like
    - missing master,
    - broken master connection,
    - missing space with error message similar to the current one.

This approach should make existing code more consistent, solve #331 and do not break existing behavior (maybe it will change error messages a bit, but regex error checks in tests should succeed nonetheless).

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* Added validation of the master availability to the `utils.get_space`
+  method before receiving the space through the connection (#95 #331).
```
DifferentialOrange, line 12: After discussion, I think we should implement master waiting with timeout or not close #95 here (we may refer to it with `Part of #95` since it partially solves it).

crud/common/utils.lua
```diff
@@ -97,6 +97,20 @@ end
...
     local replicaset = select(2, next(replicasets))
+
+    if replicaset.master == nil then
+        local error_msg = string.format(
+            'The master is not configured for replicaset %s, ' ..
+            'check status of the master and repeat the operation later',
+             replicaset.uuid)
+        error(error_msg)
+    end
+    if replicaset.master.conn.error ~= nil then
```
DifferentialOrange, line 108: But does it prevent us from schema fetch in some way?

crud/common/utils.lua
```diff
@@ -97,6 +97,20 @@ end
...
+        local error_msg = string.format(
+            'The connection to the master of replicaset %s is not valid: %s',
+             replicaset.uuid, replicaset.master.conn.error)
+        error(error_msg)
+    end
     local space = replicaset.master.conn.space[space_name]
```
DifferentialOrange, line 114: Have you considered a possibility to test it? (I don't insist, but maybe there are some good recipes in vshard tests already).

DifferentialOrange, line 114: We also may consider to use not only master to fetch schema (optional question, feel free to ignore)

crud/common/utils.lua
GRISHNOV: I've reworked `utils.get_space`, now it returns two values:
* space
* err (instance of GetSpaceError:new(error_msg))

In my current patch the error `"Space %q doesn't exist"`  is not processed through `err`. 
To handle this error, `space == nil` is checked, and after that, an instance of the error is created, depending on the location in the code (for example, `SelectError:new('Space %q doesn't exist ...')` in `select.lua`, or `CountError:new('Space %q doesn't exist ...')` in `count.lua`)

At the same time, now all `utils.get_space`  calls process both `space` and `err` returns.
In specific places `const.NEED_SCHEMA_RELOAD` is used

crud/common/utils.lua
```diff
@@ -97,13 +98,30 @@ end
...
+    if replicaset.master.conn.error ~= nil then
+        local error_msg = string.format(
+            'The connection to the master of replicaset %s is not valid: %s',
+             replicaset.uuid, replicaset.master.conn.error)
+        return nil, GetSpaceError:new(error_msg)
+    end
     local space = replicaset.master.conn.space[space_name]
 
-    return space
+    return space, nil
```
DifferentialOrange, line 117: It is not a Lua way

crud/replace.lua
```diff
@@ -66,9 +66,8 @@ local function call_replace_on_router(vshard_router, space_name, original_tuple,
 
     local space, err = utils.get_space(space_name, vshard_router:routeall())
     if err ~= nil then
-        return nil, ReplaceError:new("Failed to get space %q: %s", space_name, err), const.NEED_SCHEMA_RELOAD
+        return nil, err, const.NEED_SCHEMA_RELOAD
```
DifferentialOrange, line 69: I think all errors returned from `replace` should be of the `ReplaceError` class, at least it works like this for any other errors and we don't planned to rework it anytime soon.

Same for other API methods.

crud/common/utils.lua
GRISHNOV: Fixed

crud/replace.lua
GRISHNOV: I propose to consider this:

```lua
return nil, InsertError:new("An error occurred during the operation: %q", err.str), const.NEED_SCHEMA_RELOAD
```

Output from this:

```yaml
myapp.router> crud.insert('customer',{2,100,'Mike'})
---
- null
- line: 69
  class_name: InsertError
  err: 'An error occurred during the operation: "GetSpaceError: The connection to
    the master of replicaset 548fb6ff-d686-4298-ad17-7f81fa19588c is not valid: connect,
    called on fd 37, aka 127.0.0.1:51219: Connection refused"'
  file: '...r_bug/repro/myapp/.rocks/share/tarantool/crud/insert.lua'
  str: 'InsertError: An error occurred during the operation: "GetSpaceError: The connection
    to the master of replicaset 548fb6ff-d686-4298-ad17-7f81fa19588c is not valid:
    connect, called on fd 37, aka 127.0.0.1:51219: Connection refused"'
```

GRISHNOV:
Also, while working on the task, I noticed that [this](https://github.com/tarantool/crud/blob/master/crud/common/schema.lua#L40) line of the code triggers an error:

```yaml
myapp.router> crud.insert('customer', {1,100,'Mike'})
---
- error: '...epro/myapp/.rocks/share/tarantool/crud/common/schema.lua:40: calling
    ''status'' on bad self (fiber expected, got table)'
...
```

AFAIU, it  should be `f:status()`, not `fiber:status()`

crud/replace.lua
DifferentialOrange: Regarding `%q` and `err.str` -- I think for now you should simply stick to the existing wrappers.

GRISHNOV: Done, now using `%s` and `err`

DifferentialOrange:
> Also, while working on the task, I noticed that [this](https://github.com/tarantool/crud/blob/master/crud/common/schema.lua#L40) line of the code triggers an error:
> 
> ```yaml
> myapp.router> crud.insert('customer', {1,100,'Mike'})
> ---
> - error: '...epro/myapp/.rocks/share/tarantool/crud/common/schema.lua:40: calling
>     ''status'' on bad self (fiber expected, got table)'
> ...
> ```
> 
> AFAIU, it should be `f:status()`, not `fiber:status()`

It would be nice to fix it in a separate commit, especially if you've got a repro (should be ok even without one, if this is complicated).

GRISHNOV:
Thanks for the feedback! Now the PR is prepared for the #331

crud/common/utils.lua
```diff
@@ -97,13 +98,23 @@ end
 
 function utils.get_space(space_name, replicasets)
     local replicaset = select(2, next(replicasets))
+
+    if replicaset.master.conn.error ~= nil then
```
DifferentialOrange, line 102: Can you be sure that replicaset always has a master and master always has a non-nil connection?

DifferentialOrange (commented):
Thank you for your updates! There are some older questions that are still unresolved: I've either never got the final answer or have missed it -- please, respond to them here so all of our decisions would be preserved here.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* Added validation of the master connection to the `utils.get_space`
+  method before receiving the space through the connection (#331).
```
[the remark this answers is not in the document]
GRISHNOV: Done

crud/common/schema.lua
```diff
@@ -37,7 +37,7 @@ local function call_reload_schema(replicasets)
     for _ = 1,replicasets_num do
         if channel:get(const.RELOAD_SCHEMA_TIMEOUT) == nil then
             for _, f in ipairs(fibers) do
-                if fiber:status() ~= 'dead' then
+                if f:status() ~= 'dead' then
```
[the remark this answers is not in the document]
GRISHNOV: Added info to `CHANGELOG.md`.
Also changed the target in the name of the commit: `internal: ...` -> `bugfix: ...`

crud/common/utils.lua
GRISHNOV: At the same time, it will not be difficult to add a check `replicaset.master.con ~= nil` and `replicaset ~= nil`, just in case. ( `<attempt to index local 'replicaset' (a nil value)>` error occurs, for example, when we only have a router)

DifferentialOrange: > At the same time, the need of `replicaset.master.con ~= nil`checking is not entirely obvious to me, vshard implementation use net_box.connect that does not simply return `nil` in case of an unsuccessful connection

Ok, I think we may rely on this.

> AFAIU, checking for replicaset.master ~= nil will be implemented in the https://github.com/tarantool/crud/issues/95.

I don't think it should. Either there is something similar to the above one ("vshard guarantees that there is always a `master`") or we should check it explicitly. Otherwise, it would be still possible to get an unclear error similar to #331 one.

DifferentialOrange: > `replicaset ~= nil`

Let's add this one.

GRISHNOV: Done, added check `replicaset ~= nil`

DifferentialOrange: I'm not an expert in vshard pipeline, but I think it's theretically possible to get a `nil` master. Let's checks this one too.

https://github.com/tarantool/vshard/blob/f5f386a5a35e6e5efd8f4f2ed1b3d208fdae9095/vshard/replicaset.lua#L223-L241

GRISHNOV: Done, added check `replicaset.master ~= nil`
Also updated `CHANGELOG.md` and commit name/description

CHANGELOG.md
```diff
@@ -8,10 +8,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+  receiving the space from the connection (#331).
 
 ### Changed
 
 ### Fixed
+* Fixed a typo of the fiber status call when
+  processing `RELOAD_SCHEMA_TIMEOUT` in `call_reload_schema`.
+  Before that, an error occurred in this case with description:
+  `calling 'status' on bad self (fiber expected, got table)`.
+  To reproduce the error before this fix, it is enough to
```
DifferentialOrange, line 22: It shouldn't be in changelog.

CHANGELOG.md
```diff
@@ -8,10 +8,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 
 ### Added
+* Added validation of the master presence in replicaset and the 
+  master connection to the `utils.get_space` method before 
+  receiving the space from the connection (#331).
 
 ### Changed
 
 ### Fixed
+* Fixed a typo of the fiber status call when
```
DifferentialOrange, line 18: I don't think that users are interesting in the nature of bug ("typo"), it's more interesting to hear what behavior has been fixed (something about fiber cancel on reload)

CHANGELOG.md
GRISHNOV: Done, reproduce description deleted

CHANGELOG.md
GRISHNOV: Done, `typo` -> `bug`, and added  fixed behavior description

CHANGELOG.md
```diff
@@ -8,10 +8,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 
 ### Added
+* Added validation of the master presence in replicaset and the 
+  master connection to the `utils.get_space` method before 
+  receiving the space from the connection (#331).
 
 ### Changed
 
 ### Fixed
+* Fixed a bug of the fiber status call when
```
DifferentialOrange, line 18: Let's make it simpler. Something like "fix fiber cancel on schema reload timeout"

GRISHNOV: Simplified `CHANGELOG.md` and commit msg

crud/common/utils.lua
DifferentialOrange: Well, ok, let's not use a malformed connection.

crud/common/utils.lua
```diff
@@ -97,13 +98,37 @@ end
...
 function utils.get_space(space_name, replicasets)
     local replicaset = select(2, next(replicasets))
+
+    if replicaset == nil then
+        return nil, GetSpaceError:new(
+            'The router returned empty replicasets: ' ..
+            'perhaps other instances are unavailable or you have configured only the router')
+    end
+
+    if  replicaset.master == nil then
```
DifferentialOrange, line 108:
```suggestion
    if replicaset.master == nil then
```

CHANGELOG.md
```diff
@@ -8,10 +8,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* Added validation of the master presence in replicaset and the
```
DifferentialOrange, line 11: I think it belongs to the `Fixed`

DifferentialOrange (approved):
Thank you for the updates! Let's change a couple of cosmetic things and merge it.

crud/common/utils.lua
GRISHNOV: Done

CHANGELOG.md
GRISHNOV: Done. Sorry for the wait, there were some problems with rebase

PR #337, opened 2023-01-18
feature: master presence check with timeout

### LINKED ISSUE #95: Wait for master until a timeout

GRISHNOV:
Added timeout condition for the validation of master presence in replicaset and for the master connection to the `utils.get_space` method.

Closes #95

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* Added timeout condition for the validation of master presence in 
+  replicaset and for the master connection (#95).
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -204,7 +204,8 @@ where:
 * `space_name` (`string`) - name of the space to insert an object
 * `tuple` / `object` (`table`) - tuple/object to insert
 * `opts`:
-  * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
+  * `timeout` (`?number`) - `vshard.call` timeout and vshard master
+    discovery timeout (in seconds), default value is 2
   * `bucket_id` (`?number|cdata`) - bucket ID
   * `fields` (`?table`) - field names for getting only a subset of fields
   * `vshard_router` (`?string|table`) - Cartridge vshard group name or
@@ -263,7 +264,8 @@ where:
 * `space_name` (`string`) - name of the space to insert an object
 * `tuples` / `objects` (`table`) - array of tuples/objects to insert
 * `opts`:
-  * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
+  * `timeout` (`?number`) - `vshard.call` timeout and vshard master
+    discovery timeout (in seconds), default value is 2
   * `fields` (`?table`) - field names for getting only a subset of fields
   * `stop_on_error` (`?boolean`) - stop on a first error and report error
     regarding the failed operation and error about what tuples were not
@@ -411,7 +413,8 @@ where:
 * `opts`:
   * `fields` (`?table`) - field names for getting only a subset of fields
   * `bucket_id` (`?number|cdata`) - bucket ID
-  * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
+  * `timeout` (`?number`) - `vshard.call` timeout and vshard master
+    discovery timeout (in seconds), default value is 2
   * `mode` (`?string`, `read` or `write`) - if `write` is specified then `get` is
     performed on master, default value is `read`
   * `prefer_replica` (`?boolean`) - if `true` then the preferred target is one of
@@ -450,7 +453,8 @@ where:
 * `key` (`any`) - primary key value
 * `operations` (`table`) - update [operations](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/update/)
 * `opts`:
-  * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
+  * `timeout` (`?number`) - `vshard.call` timeout and vshard master
+    discovery timeout (in seconds), default value is 2
   * `bucket_id` (`?number|cdata`) - bucket ID
   * `fields` (`?table`) - field names for getting only a subset of fields
   * `vshard_router` (`?string|table`) - Cartridge vshard group name or
@@ -485,7 +489,8 @@ where:
 * `space_name` (`string`) - name of the space
 * `key` (`any`) - primary key value
 * `opts`:
-  * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
+  * `timeout` (`?number`) - `vshard.call` timeout and vshard master
+    discovery timeout (in seconds), default value is 2
   * `bucket_id` (`?number|cdata`) - bucket ID
   * `fields` (`?table`) - field names for getting only a subset of fields
   * `vshard_router` (`?string|table`) - Cartridge vshard group name or
@@ -522,7 +527,8 @@ where:
 * `space_name` (`string`) - name of the space
 * `tuple` / `object` (`table`) - tuple/object to insert or replace exist one
 * `opts`:
-  * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
+  * `timeout` (`?number`) - `vshard.call` timeout and vshard master
+    discovery timeout (in seconds), default value is 2
   * `bucket_id` (`?number|cdata`) - bucket ID
   * `fields` (`?table`) - field names for getting only a subset of fields
   * `vshard_router` (`?string|table`) - Cartridge vshard group name or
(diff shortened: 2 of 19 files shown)


GRISHNOV:
I did a rebase with master and also corrected the timeout condition

AFAIU, it is enough to check during the timeout whether the master appeared in the replicaset after another `vshard_router:routeall` call

crud/common/const.lua
```diff
@@ -5,6 +5,8 @@ const.RELOAD_SCHEMA_TIMEOUT = 3 -- 3 seconds
 const.FETCH_SHARDING_METADATA_TIMEOUT = 3 -- 3 seconds
 const.SHARDING_RELOAD_RETRIES_NUM = 1
 
+const.GET_SPACE_TIMEOUT = 3 -- 3 seconds
```
DifferentialOrange, line 8: @Totktonada suggested to sync this one with vshard call timeout:
https://github.com/tarantool/crud/issues/95#issuecomment-968180577 

I think it worth considering. For example, we really may subtract the space wait timeout from the call timeout or we may set get space timeout to be less or equal than this timeout. (We need to think about this before implementing, maybe discuss a bit.)

crud/common/utils.lua
```diff
@@ -105,19 +106,44 @@ function utils.get_space(space_name, replicasets)
...
-        local error_msg = string.format(
-            'The master was not found in replicaset %s, ' ..
-            'check status of the master and repeat the operation later',
-             replicaset.uuid)
-        return nil, GetSpaceError:new(error_msg)
-    end
+    if replicaset.master == nil or replicaset.master.conn.error ~= nil then
+        -- Try to get master again with timeout.
+        local deadline = fiber_clock() + const.GET_SPACE_TIMEOUT
+        local vshard_router, err = utils.get_vshard_router_instance()
```
DifferentialOrange, line 112: Calling this util without a proper argument is wrong. Either you reuse the router you've got from the outside or extract it the same way it was extracted before. Otherwise it won't work with sharding groups.

crud/common/utils.lua
```diff
@@ -105,19 +106,44 @@ function utils.get_space(space_name, replicasets)
...
+        if err ~= nil then
+            return nil, GetSpaceError:new(err)
+        end
+        while true do
+            ::continue::
+            fiber.yield()
+            replicasets = vshard_router:routeall()
+            replicaset = select(2, next(replicasets))
+
+            if replicaset.master ~= nil and replicaset.master.conn.error ~= nil then
```
DifferentialOrange, line 122: Reducing code copypaste also is a way to reduce the possibility of bugs. For example, here you've forgot to checks that replicasets aren't nil even though this check is already implemented.

DifferentialOrange (commented):
That you for this PR. See my comments below.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* Added timeout condition for the validation of master presence in 
+  replicaset and for the master connection to the `utils.get_space`
```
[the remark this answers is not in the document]
GRISHNOV: Done, mention of the method deleted

crud/common/utils.lua
```diff
@@ -22,6 +22,7 @@ local NotInitializedError = errors.new_class('NotInitialized')
 local StorageInfoError = errors.new_class('StorageInfoError')
 local VshardRouterError = errors.new_class('VshardRouterError', {capture_stack = false})
 local fiber_clock = require('fiber').clock
```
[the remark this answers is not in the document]
GRISHNOV: Done, `fiber_clock` -> `fiber.clock`

crud/common/const.lua
GRISHNOV: Thanks, I'm thinking about it now. If the proposed version of the loop is suitable, a more complex timeout condition can be implemented there

crud/common/utils.lua
GRISHNOV: I propose to consider a current version that passes `vshard_router` inside the `utils.get_space`, after which `vshard_router:routeall()` is called inside timeout loop in `utils.get_space`.

However, there are nuances for `select.lua`, `select_old.lua`, `borders.lua`, `count.lua`:
Before my changes where was one call `local replicasets, err = vshard_router:routeall()`, and `replicasets` passed to `utils.get_space`.

But now there is `vshard_router:routeall()` call inside `utils.get_space` too for these files. 
Perhaps there will be two different versions of `replicasets` values. Now I'm trying to figure out if there is a potential problem here

crud/common/utils.lua
GRISHNOV: Done, now `replicaset ~= nil` check there is inside timeout loop

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+* Added timeout condition for the validation of master presence in
```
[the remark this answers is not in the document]
GRISHNOV: Done, now using `feature`

CHANGELOG.md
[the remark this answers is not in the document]
DifferentialOrange: Please, update the PR name too

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
```
DifferentialOrange, line 8: You've got typo in PR name

crud/borders.lua
```diff
@@ -102,7 +101,7 @@ local function call_get_border_on_router(vshard_router, border_name, space_name,
 
     local call_opts = {
         mode = 'read',
-        replicasets = replicasets,
+        replicasets = vshard_router:routeall(),
```
DifferentialOrange, line 104: What's about err check here?

crud/common/utils.lua
```diff
@@ -96,8 +96,32 @@ function utils.format_replicaset_error(replicaset_uuid, msg, ...)
...
+           replicaset.master ~= nil and
+           replicaset.master.conn.error == nil then
+            -- Break if there is master in replicaset and
+            -- connection to master is established.
+            break
+        end
+    until (
+        -- If at least one of the lower conditions in this block is not met
+        -- and the deadline has not yet been reached, the cycle will be repeated.
+        replicaset ~= nil and
```
DifferentialOrange, line 117: Well, I don't see a reason to do double-check here. For now there is no cases when the first condition (before `or`) may stop the cycle.

Since we want to try to extract replicasets at least once, `do ... while` seems more suitable here.

DifferentialOrange (commented):
I think you may continue with more complicated timeout implementation

crud/common/utils.lua
DifferentialOrange: Since replicasets used only to extract util space, I don't see any possible issues (but it would be nice to do a more thorough check, if you wish)

crud/common/utils.lua
GRISHNOV: Done, simplified 

AFAIS, the syntactic construction `do ... while (...)` dose not exist in Lua, and in similar cases we can use `repeat ... until NOT (...)`. 
Now I have rewritten the loop with `while (...) do ... end`, and as far as I understand, we can be sure that it will be run at least once (only if there is no anomaly with the `fiber.clock()` call inside while condition block)

crud/borders.lua
GRISHNOV: Done, implemented

CHANGELOG.md
GRISHNOV: Updated

CHANGELOG.md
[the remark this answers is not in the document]
GRISHNOV: Done, updated

crud/common/utils.lua
DifferentialOrange: > AFAIS, the syntactic construction `do ... while (...)` dose not exist in Lua, and in similar cases we can use `repeat ... until NOT (...)`.

Yeah, sorry, I was misled by some Lua tutorial website yesterday evening.

crud/common/utils.lua
```diff
@@ -96,8 +96,26 @@ function utils.format_replicaset_error(replicaset_uuid, msg, ...)
...
+        fiber.clock() < deadline
+    ) do
+        -- Try to get master with timeout.
+        fiber.yield()
+        replicasets = vshard_router:routeall()
+        replicaset = select(2, next(replicasets))
+        if replicaset ~= nil and
+           replicaset.master ~= nil and
+           replicaset.master.conn.error == nil then
+            -- Break if there is master in replicaset and
```
DifferentialOrange, line 114: I think this comment is excessive, but up to you

GRISHNOV: Done, deleted

README.md
```diff
@@ -1108,7 +1117,8 @@ where:
 
 * `space_name` (`string`) - name of the space
 * `opts`:
-  * `timeout` (`?number`) - `vshard.call` timeout (in seconds)
+  * `timeout` (`?number`) - `vshard.call` timeout and vshard master
+    discovery timeout (in seconds)
```
DifferentialOrange, line 1121: Since we've updated the descriptions, let's describe default everywhere

DifferentialOrange (approved):
Seems fine. Did you've discussed anything with @Totktonada about subtraction?

README.md
GRISHNOV: Done, added `, default value is 2`

GRISHNOV:
> Did you've discussed anything with @Totktonada about subtraction?

 Yes, I have discussed the current solution, it looks ok

PR #341, opened 2023-02-03
fix: pre-hotreload cartridge support

DifferentialOrange:
This PR fixes an unreasonable version support drop introduced in #244, as well as test pipeline update. See commit messages for more info.

I didn't forget about

- [x] Tests
- [x] Changelog
- Documentation (not needed)

Closes #244

### DIFF
--- a/test/integration/reload_test.lua
+++ b/test/integration/reload_test.lua
@@ -77,6 +77,9 @@ g.after_each(function()
 end)
 
 function g.test_router()
+    t.skip_if(not helpers.is_cartridge_hotreload_supported(),
+        "Cartridge roles reload is not supported")
+
     g.highload_fiber = fiber.new(highload_loop, 'A')
 
     g.cluster:retrying({}, function()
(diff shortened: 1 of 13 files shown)


test/integration/reload_test.lua
```diff
@@ -99,6 +102,9 @@ function g.test_router()
 end
 
 function g.test_storage()
+    t.skip_if(pcall(require, 'cartridge.hotreload') == false,
```
AnaNek, line 105: As I can see, there is a check (`pcall(require, 'cartridge.hotreload') == false`)  of `hotreload` support in several places in the code. Maybe it should be a separate function? BTW it is unlikely the realization of getting `hotreload` will change, but I would put 
```lua
function get_cartridge_hotreload()
    return pcall(require, 'cartridge.hotreload')
end
```
in a separate function.

AnaNek (approved):
Thank you for the patch! I left minor comments, but it's not critical.

test/integration/reload_test.lua
DifferentialOrange: That's fair, reworked

PR #369, opened 2023-06-29
adding new operators

AndreiSemenovMinsk:
**Added a way to search by multiple values:**
### Sample:
crud.select('users', {{'=', 'name', **{'Andrey', 'Sergey'}**}}, {fields = {'_id', 'name'}});

### Added new **operators** such as:
'<>' - not
'>><<' -  between
'<<>>' - out of
'<<!>>' - out of or null
'>)' - include
### Sample:
crud.select('users, {{ '**>><<**', 'activated_at', { dateFrom, dateTo } } }, { fields = { '_id' }})

### DIFF
--- a/crud/compare/conditions.lua
+++ b/crud/compare/conditions.lua
@@ -13,6 +13,11 @@ conditions.operators = {
     LE = '<=',
     GT = '>',
     GE = '>=',
+    NO = '@<>',
+    BTW = '@>><<',
+    OUT = '@<<>>',
+    OUTN = '@<<!>>',
+    INC = '@>)',
 }
 
 local tarantool_iter_by_cond_operators = {
@@ -54,6 +59,11 @@ local cond_operators_by_func_names = {
     le = conditions.operators.LE,
     gt = conditions.operators.GT,
     ge = conditions.operators.GE,
+    no = conditions.operators.NO,
+    between = conditions.operators.BTW,
+    outer = conditions.operators.OUT,
+    outer_or_null = conditions.operators.OUTN,
+    include = conditions.operators.INC,
 }
 
 for func_name, operator in pairs(cond_operators_by_func_names) do
@@ -75,6 +85,11 @@ local funcs_by_symbols = {
     ['<='] = conditions.funcs.le,
     ['>'] = conditions.funcs.gt,
     ['>='] = conditions.funcs.ge,
+    ['<>'] = conditions.funcs.no,
+    ['>><<'] = conditions.funcs.between,
+    ['<<>>'] = conditions.funcs.outer,
+    ['<<!>>'] = conditions.funcs.outer_or_null,
+    ['>)'] = conditions.funcs.include
 }
 
 function conditions.parse(user_conditions)
--- a/crud/compare/filters.lua
+++ b/crud/compare/filters.lua
@@ -204,16 +204,22 @@ end
 
 local function gen_tuple_fields_def_code(filter_conditions)
     -- get field names
-    local fields_added = {}
     local fields = {}
 
     for _, cond in ipairs(filter_conditions) do
-        for i = 1, #cond.values do
-            local field = cond.fields[i]
+        --no matter the number of fields coincide number of values, the fields array contains only uniq field names
+        if #cond.fields == 1 then
+            local field = cond.fields[1]
+            table.insert(fields, field)
+        else
+            local fields_added = {}
+            for i = 1, #cond.fields do
+                local field = cond.fields[i]
 
-            if not fields_added[field] then
-                table.insert(fields, field)
-                fields_added[field] = true
+                if not fields_added[field] then
+                    table.insert(fields, field)
+                    fields_added[field] = true
+                end
             end
         end
     end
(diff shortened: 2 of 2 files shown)


crud/compare/conditions.lua
DifferentialOrange, line 1: > Added a way to search by multiple values:
> Sample:
> 
> crud.select('users', {{'=', 'name', {'Andrey', 'Sergey'}})

Tarantool supports multi-part indexes. For example, let's consider index `full_name` with two string fields `name` and `surname`. In this case, select request will look like this.

```lua
crud.select('users', {{'=', 'full_name', {'Andrey', 'Andreev'}})
```

So, with this approach the same API is used for two different cases, which can worsen the debug and potentially break existing multipart key API.

DifferentialOrange: Tarantool also supports RTREE indexes, which can index arrays as keys. `crud` doesn't support RTREE as a primary condition, but it seems that it is possible to use it as a secondary condition.

Unfortunately, there are no tests for RTREE now, but it shouldn't be hard to cover this case too.

crud/compare/conditions.lua
```diff
@@ -13,6 +13,11 @@ conditions.operators = {
     LE = '<=',
     GT = '>',
     GE = '>=',
+    NO = '@<>',
```
DifferentialOrange, line 16: Is there a reason why operators here start with `@` and below they are not?

crud/compare/conditions.lua
```diff
@@ -13,6 +13,11 @@ conditions.operators = {
...
+    BTW = '@>><<',
+    OUT = '@<<>>',
+    OUTN = '@<<!>>',
+    INC = '@>)',
 }
 
 local tarantool_iter_by_cond_operators = {
```
DifferentialOrange, line 23: It seems rather weird that these table isn't updated after adding new operators.

crud/compare/conditions.lua
```diff
@@ -75,6 +85,11 @@ local funcs_by_symbols = {
     ['<='] = conditions.funcs.le,
     ['>'] = conditions.funcs.gt,
     ['>='] = conditions.funcs.ge,
+    ['<>'] = conditions.funcs.no,
```
DifferentialOrange, line 88: Does this syntax inspired by some existing database API or not? Not all of these are intuitive.

crud/compare/conditions.lua
DifferentialOrange, line 1: > crud.select('users, {{ '>><<', 'activated_at', { dateFrom, dateTo } } }, { fields = { '_id' }})

Again, here we have a clash with multipart key API.

crud/compare/filters.lua
```diff
@@ -469,6 +553,15 @@ local function gen_filter_code(filter_conditions)
     }
```
DifferentialOrange, line 553: Such new feature definitely requires

- tests,
- documentation update,
- changelog update.

crud/compare/filters.lua
```diff
@@ -264,25 +270,62 @@ local function add_collation_postfix(func_name, value_opts)
...
+    end
+
+    return format_comp_with_value(field, func_name, value)
+end
+
 local function format_eq(cond)
     local cond_strings = {}
     local values_opts = cond.values_opts or {}
 
+    if #cond.values == 1 then
```
DifferentialOrange, line 296: Multivalue equity select and range operators are different features, so it's better to implement them in separate commits (or maybe even separate PRs).

DifferentialOrange (changes requested):
Sorry for the long response. It is definitely a shame. To be honest, the main reason I've postponed my review is that features introduced is this PR are non-trivial. For a start, their API must be well-thought-out and compatible with existing API. API proposed in this PR, at least for multivalue select, is not satisfying enough. We also should consider other factors, like `crud` API in language connectors:
- https://github.com/tarantool/go-tarantool/blob/master/crud
- https://github.com/tarantool/tarantool-python/blob/66e53bca472fec0efacd690ebe6c54c33a7df3f9/tarantool/connection.py#L2132-L2971
- https://github.com/tarantool/cartridge-java/tree/167bd38c92ec3e581a9d087c6e29a0a2c66309de/src/main/java/io/tarantool/driver/core/proxy
New API should either be supported out-of-box or after some well-compatible updates of existing API.

The simplest solution to the question is to use some other operator for multivalue select and do not change the API for new operators.

The other approach to introduce conditions with more than three fields. But this one is badly compatible with tarantool-python and go-tarantool CRUD API (likely cartridge-java too, I'm not familiar with this one).

So you may either implement the "simplest" one or propose a different solution. I'm not sure I will be able to design a better API anytime soon.

We also need to justify new operators. If it is some existing API in other databases, we'll need to check whether the meaning is the same.

The other question is tests. This features definitely should be covered by new tests. You can use existing tests for a reference -- `crud` is one of the best modules here regarding the quantity of luatest tests.

Both "big" questions can be covered in any order (just keep in mind that you may need to rework tests a bit after possible API change). And both of them requires a lot of resources. It is fine if you won't be able to provide them. It is likely that one day someone from the Tarantool Ecosystem Team will finish the PR. But it is also likely that this one day may be really postponed until Product Team will give its approval.

(Just in case: I will be on vacation for the next two weeks, so I'm not sure I will be able to answer even minor questions for this time.)

DifferentialOrange:
I'll also need to fix the CI to run on external PRs.

PR #388, opened 2023-10-16
test: fix flaky cached schema case

DifferentialOrange:
We cannot guarantee that cached schema is always the old one since net.box may reload the schema, see [1] for example of fail.

1. https://github.com/tarantool/crud/actions/runs/6531939885/job/17734151700

- [x] Tests


oleg-jukovec (approved):
I think I need to add a comment to the `cached` option to clarify that the schema is not always a last/previous one.

PR #400, opened 2023-12-12
doc: fix example from lua fun

vakhov:
Fixing a bug in the lua fun example

I didn't forget about

- [x] Documentation


DifferentialOrange (approved):
Thank you for the contribution! These doc really haven't been updated for a while.

PR #404, opened 2023-12-22
Support vshard names as keys

### LINKED ISSUE #403: Support vshard's `identification_mode` = `name_as_key`
It is released in [vshard 0.1.25](https://github.com/tarantool/vshard/releases/tag/0.1.25). See https://github.com/tarantool/doc/issues/3942.

This mode has different `vshard.storage.info()` output: names are there instead of UUIDs.

At least the following crud's code is not ready for that.

https://github.com/tarantool/crud/blob/a897f4c652c6cb4313ed42abd4fc0248d3e45829/crud.lua#L182-L196

DifferentialOrange:
In vshard 0.1.25, new feature related to vshard configuration and storage info was introduced [1]. If the new mode is used, crud module fails to bootstrap and work in several places. This feature is enabled by Tarantool 3.0 if vshard cluster was configured with 3.0 config.

After this patch, it is possible to bootstrap a vshard cluster with new configuration mode. We run all existing vshard tests with new modes after this patch. Module code was updated to support both possible mods.

This PR also introduces several minor fixes inspired by flaky tests. (They mostly have failed locally on my PC with vinyl engine, but it is possible to get the same result on Github Actions.)

I didn't forget about

- [x] Tests
- [x] Changelog

Closes #403

### DIFF
--- a/.github/workflows/test_on_push.yaml
+++ b/.github/workflows/test_on_push.yaml
@@ -37,13 +37,13 @@ jobs:
             cartridge-version: "2.8.0"
           - tarantool-version: "2.11"
             metrics-version: "1.0.0"
-            vshard-version: "0.1.24"
+            vshard-version: "0.1.25"
           - tarantool-version: "2.11"
             external-merger-version: "0.0.5"
             external-keydef-version: "0.0.4"
           - tarantool-version: "master"
             metrics-version: "1.0.0"
-            vshard-version: "0.1.24"
+            vshard-version: "0.1.25"
       fail-fast: false
     # Can't install older versions on 22.04,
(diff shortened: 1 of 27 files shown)


better0fdead (commented):
Hi, thx for patch. Code looks amazing. I think it would be good to update README section about min and max operations. (add arguments description as it is done for other operations)
Also, pls fix grammar in commit messages:

- 1st commit last sentence: Module code was updated to support both possible MODES.
- 4th commit: Now it is possible to perform min/max on
write instances, and TESTS were updated to do so.

better0fdead (approved):

PR #410, opened 2023-12-29
readme: make consistent return variable name

ArtDu:
Min and max return only one tuple.
Name `objects` as a variable in return value may confuse users.

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Closes #???

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,7 +445,7 @@ errs[4].operation_data -- {92, 2040, "Artur", 29}
 ### Get
 
 ```lua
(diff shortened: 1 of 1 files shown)


ArtDu:
> As far as I can see, documentation is inconsistent: there are both `result` and `object/objects`. Since the first one is easier to keep consistent, let's try to stick to this one.

Then I'll change to result in other options

DifferentialOrange (approved):
Thank you for the PR!

PR #419, opened 2024-02-02
scan: fix filter erroneous early exit

### LINKED ISSUE #418: Проблема с crud.count
Делали запрос crud.count с такими-то параметрами (на видео, которое они скидывали, есть настоящий пример). При перестановке условий получается разный результат, хотя этого не ожидалось. При замене условий с >= на = результат второго запроса больше результата первого, хотя этого тоже не ожидалось.
В принципе, то, чего они не ожидали, действительно не должно быть
В условиях здорового кластера

DifferentialOrange:
The issue described below is related to the read operations which allows to scan: `crud.select`, `crud.pairs`, `crud.count`, `readview:select` and `readview:pairs`.

The erroneous behavior reported by [1] and #418 is as follows:
- result changes when reordering operation conditions;
- when `>=` condition operation is changed to `=`, there are more rows in the result.

The reason is as follows. Scanning read operates with two entities: an iterator and a filter. The iterator includes an index, a starting
value and iterator type (EQ, GT, etc.). The iterator is built from conditions, if possible, otherwise primary index is used. Remaining conditions form the filter, so the actual result satisfies all operation conditions.

The filter supports early exit. Let's consider the following example. For `crud.select(space, {{'>=', 'id', 1}, {'<=', 'id', 10}})`, where `id` is an index (or an indexed field), the iterator uses index `id`, starts from key = `1` and goes by GE rules, covering [1, +inf) ordered keys. On the other hand, when iterator reaches the tuple with `id` = `11`, all tuples after this one will never satisfy the second condition, because our iterator yields tuples sorted by `id` (due to underlying index). So filter tells than there is no reason to scan anymore, and we finish the scanning procedure.

Before this patch, the function behind early exit decision had worked as follows: "if the condition is an index, we go in forward (reverse) order and `<=` or `<` (`>=` or `>`) condition is violated, there is no reason to scan anymore". But the valid approach is "if the condition is SCANNING index...". Before this patch, filter had assumed that if the condition for index is specified, tuples are ordered, but it works only if iterator uses the same index as in the condition. This patch fixes the issue.

The erroneous behavior may happen in the following case:
- there are multiple conditions,
- there are at least two different index operands,
- non-scanning index condition uses `<=`, `<`, `>=` or `>` operation.

1. https://jira.vk.team/browse/TNT-941

Closes #418

### DIFF
--- a/crud/compare/filters.lua
+++ b/crud/compare/filters.lua
@@ -17,12 +17,8 @@ local filters = {}
     - opposite condition `'id' < 100` becomes false
     - in such case we can exit from iteration
 ]]
-local function is_early_exit_possible(index, tarantool_iter, condition)
-    if index == nil then
-        return false
-    end
-
-    if index.name ~= condition.operand then
+local function is_early_exit_possible(scan_index, tarantool_iter, condition)
+    if scan_index.name ~= condition.operand then
         return false
     end
 
@@ -77,8 +73,8 @@ local function get_values_opts(index)
     return index_field_opts
 end
 
-local function parse(space, conditions, opts)
-    dev_checks('table', '?table', {
+local function parse(space, scan_index, conditions, opts)
+    dev_checks('table', 'table', '?table', {
         scan_condition_num = '?number',
         tarantool_iter = 'number',
     })
@@ -137,7 +133,11 @@ local function parse(space, conditions, opts)
                 operator = condition.operator,
                 values = condition.values,
                 types = fields_types,
-                early_exit_is_possible = is_early_exit_possible(index, opts.tarantool_iter, condition),
+                early_exit_is_possible = is_early_exit_possible(
+                    scan_index,
+                    opts.tarantool_iter,
+                    condition
+                ),
                 values_opts = values_opts,
             })
         end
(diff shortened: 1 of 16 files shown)


DifferentialOrange:
CI is red due to network issues, it's not related to the actual fix.

crud/compare/filters.lua
```diff
@@ -17,12 +17,8 @@ local filters = {}
...
     - in such case we can exit from iteration
 ]]
-local function is_early_exit_possible(index, tarantool_iter, condition)
-    if index == nil then
-        return false
-    end
-
-    if index.name ~= condition.operand then
+local function is_early_exit_possible(index, scan_index, tarantool_iter, condition)
+    if index ~= scan_index then
```
oleg-jukovec, line 21: To avoid problems in the future (an index may be `deepcopy'ed` in some place).
```suggestion
    if index.name ~= scan_index.name then
```

oleg-jukovec (approved):
Thank you for the patch!

crud/compare/filters.lua
DifferentialOrange: Reworked

PR #425, opened 2024-03-15
scan: fix nil filters

### LINKED ISSUE #422: Handle 'wrong symbol )' exception in case of comparision field with nil or {} on nonindexed column
Please wrap a meaningless exception with something like this:

Please add index on column bla-bla like COMMAND

DifferentialOrange:
Before this patch, some nil conditions failed with internal filter library build error. This patch fixes this internal error, as well as normalize filters to behave similar to core indexes.

The logic in core Tarantool select is as follows. `nil` condition in index select is an absence of condition, thus all data is returned disregarding the condition (condition may affect the order). `box.NULL` condition is a condition for the null value -- in case of EQ, only records with null index value are returned, in case of GT, all non-null values are returned since nulls are in the beginning of an index and so on. `nil`s and `box.NULL`s in tuple are both satisfy `box.NULL` equity.

After this patch, `nil` filter condition is treated as no condition. This is a breaking change since conditions for `'>'` and `'<'` operator with `nil` operand had resulted with empty response before this patch. But since it was inconsistent with scanning index conditions and wasn't intentional, we change it here.

Closes #422

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   non-iterating indexes (#373).
 * Passing errors from storages for merger operations (`crud.select`,
   `crud.pairs`, `readview:select`, `readview:pairs`) (#423).
+* Working with `nil` operand conditions in case of non-indexed fields or
+  non-iterating indexes (#422).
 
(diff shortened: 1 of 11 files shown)


better0fdead (approved):
Thx for patch, code looks great, LGTM

PR #430, opened 2024-04-03
ci: install newer vshard for reusable test

DifferentialOrange:
Tarantool 3 config requires vshard 0.1.25 or newer, while tests install older versions by default (since crud is able to work with older vshards).

At the same time, newer vshard can work with old Tarantools.

I didn't forget about

- [x] Tests


0x501D (approved):
LGTM

PR #455, opened 2025-09-10
sharding: fail fast on invalid bucket_id in data or opts

vakhov:
- Added early validation of bucket_id in tuple/object and in opts inside tuple_set_and_return_bucket_id().
- Error messages now specify where the invalid value was provided:
  * "Invalid bucket_id in data: expected unsigned, got %s"
  * "Invalid bucket_id in opts: expected unsigned, got %s"
- Extended integration tests for insert/replace/upsert to cover these cases.

I didn't forget about

- [x] Tests
- [ ] Changelog
- [ ] Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ sdk: sdk-2 sdk-3
 	chmod 644 sdk-3/rocks/* && \
 	tt rocks make_manifest sdk-3/rocks
 
-lint: .rocks
+lint:
 	source sdk-2/env.sh && .rocks/bin/luacheck .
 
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -28,7 +28,7 @@ sdk: sdk-2 sdk-3
 	chmod 644 sdk-3/rocks/* && \
 	tt rocks make_manifest sdk-3/rocks
 
-lint: .rocks
```
ita-sammann, line 31: Why `.rocks` dependency is removed here?

PR #462, opened 2025-11-19
TNTP-2109: Switch to "safe" mode on vshard rebalance

ita-sammann:

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Closes #???

### DIFF
--- a/test/integration/double_buckets_test.lua
+++ b/test/integration/double_buckets_test.lua
@@ -1,0 +1,316 @@
+local t = require('luatest')
+local json = require('json')
+local fiber = require('fiber')
+
+local utils = require('crud.common.utils')
+
+local helpers = require('test.helper')
+
+local function wait_balance(g)
+    t.helpers.retrying({timeout=30}, function()
+        local buckets_count_s1 = g.cluster:server('s1-master').net_box:eval("return box.space._bucket:len()")
+        local buckets_count_s2 = g.cluster:server('s2-master').net_box:eval("return box.space._bucket:len()")
+        t.assert_equals(buckets_count_s1, 1500)
+        t.assert_equals(buckets_count_s2, 1500)
+    end)
+end
+
+local function balance_cluster(g)
+    if g.params.backend == "config" then
+        local cfg = g.cluster:cfg()
+        cfg.groups.storages.replicasets["s-1"].sharding = {
+            weight = 1,
+        }
+        cfg.groups.storages.replicasets["s-2"].sharding = {
+            weight = 1,
+        }
+        g.cluster:cfg(cfg)
+        wait_balance(g)
+    end
+end
+
+local pgroup_duplicates = t.group('double_buckets_duplicates', helpers.backend_matrix({
+    {engine = 'memtx', operation = 'replace'},
+    {engine = 'memtx', operation = 'insert'},
+    {engine = 'memtx', operation = 'upsert'},
+    {engine = 'memtx', operation = 'insert_many'},
+    {engine = 'memtx', operation = 'replace_many'},
+    {engine = 'memtx', operation = 'upsert_many'},
+}))
+
+pgroup_duplicates.before_all(function(g)
+    helpers.start_default_cluster(g, 'srv_simple_operations')
+end)
+
+pgroup_duplicates.after_all(function(g)
+    helpers.stop_cluster(g.cluster, g.params.backend)
+end)
+
+pgroup_duplicates.before_each(function(g)
+    helpers.truncate_space_on_cluster(g.cluster, 'customers')
+end)
+
+pgroup_duplicates.after_each(function(g)
+    balance_cluster(g)
+end)
+
+--- Rebalance stalls if we move all buckets at once; use a small subset.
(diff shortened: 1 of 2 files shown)


crud/common/map_call_cases/base_iter.lua
```diff
@@ -72,7 +72,7 @@ function BaseIterator:get()
     local replicaset = self.next_replicaset
     self.next_index, self.next_replicaset = next(self.replicasets, self.next_index)
 
-    return self.func_args, replicaset, replicaset_id
+    return self.func_args, replicaset, replicaset_id, {}
```
Satbek: please write 4th return name in doc for method

crud/common/rebalance.lua
```diff
@@ -0,0 +1,181 @@
+local fiber = require('fiber')
+local vshard_consts = require('vshard.consts')
+local utils = require('crud.common.utils')
+
+local MODULE_INTERNALS = '__module_crud_rebalance'
+local SETTINGS_SPACE_NAME = '_crud_settings'
+
+
+local M = rawget(_G, MODULE_INTERNALS)
```
Satbek: there is no rawset for MODULE_INTERNALS

crud/common/rebalance.lua
```diff
@@ -0,0 +1,181 @@
...
+    settings_space:create_index('primary', { parts = { 'key' }, if_not_exists = true })
+end
+
+local function safe_mode_trigger(_, new, space, op)
+    if space ~= '_bucket' then
+        return
+    end
+    if (op == 'INSERT' and new.status == vshard_consts.BUCKET.RECEIVING) or
+            (op == 'REPLACE' and new.status == vshard_consts.BUCKET.SENDING) then
+        box.broadcast('_crud.safe_mode_enable', true)
```
Satbek: may be add constant for `_crud.safe_mode_enable` too?

crud/common/rebalance.lua
```diff
@@ -0,0 +1,181 @@
...
+
+    for hook, _ in pairs(M.safe_mode_enable_hooks) do
+        hook()
+    end
+end
+
+local function safe_mode_disable()
+    if not box.info.ro then
+        box.space[SETTINGS_SPACE_NAME]:replace{ 'safe_mode', false }
+        box.space._bucket:on_replace(safe_mode_trigger)
```
Satbek: here you set trigger

crud/common/rebalance.lua
```diff
@@ -0,0 +1,181 @@
...
+    end
+    M.safe_mode = false
+
+    for hook, _ in pairs(M.safe_mode_disable_hooks) do
+        hook()
+    end
+end
+
+local function rebalance_init()
+    box.watch('box.status', function()
```
Satbek: box.watch was introduced https://github.com/tarantool/tarantool/releases/tag/2.10.0-beta2

need to write about it, that rebalance won't work on earlier versions

crud/common/call.lua
```diff
@@ -14,14 +16,79 @@ local CallError = errors.new_class('CallError')
...
+            bucket_unref_many(reffed, mode)
+            return nil, err
+        end
+        table.insert(reffed, bucket_id)
+    end
+    return true, nil
+end
+
+local function call_on_storage_safe(run_as_user, bucket_ids, mode, func_name, ...)
+    fiber.name(CRUD_CALL_FIBER_NAME .. func_name)
```
Satbek: should we mark safe functions?

Satbek (commented):
What if, instead of threading `bucket_id` through all the call chains just to support `safe_mode`, we add:

```lua
bucket_ref(tuple.bucket_id)
bucket_unref(tuple.bucket_id)
```

directly in all write methods?

* In “fast” mode `bucket_ref` / `bucket_unref` would be no-ops.
* In “slow/safe” mode they would delegate to `vshard.storage.bucket_ref` / `vshard.storage.bucket_unref`.

This way:

* the `call_on_storage` signature doesn’t have to change (no need to pass `bucket_id` around);
* `ref/unref` calls live exactly where they logically belong (next to writes), instead of being scattered across different places only for `safe_mode`.

To me this looks more readable and less error-prone than propagating `bucket_id` solely for safe mode handling.



Also we need to add document about rebalance in README or another document in doc dir

crud/common/call.lua
```diff
@@ -93,7 +162,18 @@ local function retry_call_with_master_discovery(replicaset, method, func_name, f
         return resp, err
     end
 
-    if err.name == 'MISSING_MASTER' and replicaset.locate_master ~= nil then
+    if err.name == 'WRONG_BUCKET' or
+       err.name == 'BUCKET_IS_LOCKED' or
+       err.name == 'TRANSFER_IS_IN_PROGRESS' then
```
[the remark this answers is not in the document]
ita-sammann: I just tried to beautify it)

crud/common/rebalance.lua
ita-sammann: Added a constant

Issue #448, opened 2025-03-20
Calls bypass vshard's checks 


Serpentian:
This is a critical issue and leads to duplication of data in cluster as described in https://jira.vk.team/browse/TNTP-2109. It may also break vshard. Crud makes call, using `replicaset` object, which is low-level API:

https://github.com/tarantool/crud/blob/419296e30db2144db7f6d30593124eb80ee807d1/crud/replace.lua#L102-L105

`CRUD_REPLACE_FUNC_NAME` is `crud.replace_on_storage`. It's called in `retry_call_with_master_discovery`:

https://github.com/tarantool/crud/blob/419296e30db2144db7f6d30593124eb80ee807d1/crud/common/call.lua#L90

So, at the end we have, that `replace_on_storage` is explicitly called. However, you CANNOT rely on `routeall` and `route` functions in vshard, they return cache value of the router, which may outdated. In order to figure out, whether the bucket is on the instance, explicit call must be done.

The example is about `replace` explicitly, however, In the scope of this issue, all requests must be checked. They should not use `replicaset:call` as it's now, `vshard.router.call` is more preferable. If `replicaset:call` should be preserved, then it must call `vshard.storage.call` with `{bucket_id, mode, func, args}` argument (see, how `vshard.router.call` is implemented: [link](https://github.com/tarantool/vshard/blob/76b3ad19b539150bf597a5ffec91b97758b69a00/vshard/router/init.lua#L648-L650)

PR #467, opened 2025-12-05
[TNTP-2109] Switch to safe mode on vshard rebalance

ita-sammann:
- Implement "safe" mode to prevent writing data to wrong replicaset when vshard rebalance is in progress.
 - Auto switch to safe mode when rebalance process starts.
 - Manual return to fast mode.

I didn't forget about

- [x] Tests
- [x] Changelog
- [ ] Documentation

### DIFF
--- a/crud/common/call.lua
+++ b/crud/common/call.lua
@@ -1,11 +1,15 @@
 local errors = require('errors')
+local log = require('log')
 
 local call_cache = require('crud.common.call_cache')
 local dev_checks = require('crud.common.dev_checks')
 local utils = require('crud.common.utils')
 local sharding_utils = require('crud.common.sharding.utils')
-local fiber_clock = require('fiber').clock
+local fiber = require('fiber')
+local fiber_clock = fiber.clock
 local const = require('crud.common.const')
+local rebalance = require('crud.common.rebalance')
+local bucket_ref_unref = require('crud.common.sharding.bucket_ref_unref')
 
 local BaseIterator = require('crud.common.map_call_cases.base_iter')
 local BasePostprocessor = require('crud.common.map_call_cases.base_postprocessor')
@@ -14,13 +18,37 @@ local CallError = errors.new_class('CallError')
 
 local CALL_FUNC_NAME = 'call_on_storage'
 local CRUD_CALL_FUNC_NAME = utils.get_storage_call(CALL_FUNC_NAME)
-
+local CRUD_CALL_FIBER_NAME = CRUD_CALL_FUNC_NAME .. '/fast'
 
 local call = {}
 
-local function call_on_storage(run_as_user, func_name, ...)
+local function call_on_storage_safe(run_as_user, func_name, ...)
     return box.session.su(run_as_user, call_cache.func_name_to_func(func_name), ...)
 end
+
+local function call_on_storage_fast(run_as_user, func_name, ...)
+    fiber.name(CRUD_CALL_FIBER_NAME)
+    return box.session.su(run_as_user, call_cache.func_name_to_func(func_name), ...)
+end
+
+local call_on_storage = rebalance.safe_mode and call_on_storage_safe or call_on_storage_fast
+
+rebalance.on_safe_mode_toggle(function(is_enabled)
+    if is_enabled then
+        call_on_storage = call_on_storage_safe
+
+        local fibers_killed = 0
+        for fb_id, fb in pairs(fiber.info()) do
+            if fb.name == CRUD_CALL_FIBER_NAME then
+                fiber.kill(fb_id)
+                fibers_killed = fibers_killed + 1
+            end
+        end
+        log.debug('Killed %d fibers with fast-mode crud requests.', fibers_killed)
+    else
+        call_on_storage = call_on_storage_fast
+    end
+end)
 
 call.storage_api = {[CALL_FUNC_NAME] = call_on_storage}
 
@@ -82,7 +110,10 @@ local function wrap_vshard_err(vshard_router, err, func_name, replicaset_id, buc
     ))
 end
 
-local function retry_call_with_master_discovery(replicaset, method, func_name, func_args, call_opts)
+--- Executes a vshard call and retries once after performing recovery actions
+--- like bucket cache reset, destination redirect (for single calls), or master discovery.
+local function call_with_retry_and_recovery(vshard_router,
+    replicaset, method, func_name, func_args, call_opts, is_single_call)
     local func_args_ext = utils.append_array({ box.session.effective_user(), func_name }, func_args)
 
     -- In case cluster was just bootstrapped with auto master discovery,
@@ -93,7 +124,27 @@ local function retry_call_with_master_discovery(replicaset, method, func_name, f
         return resp, err
     end
 
-    if err.name == 'MISSING_MASTER' and replicaset.locate_master ~= nil then
+    -- This is a partial copy of error handling from vshard.router.router_call_impl()
+    -- It is much simpler mostly because bucket_set() can't be accessed from outside vshard.
+    if err.class_name == bucket_ref_unref.BucketRefError.name then
+        if is_single_call and #err.bucket_ref_errs == 1 then
+            local single_err = err.bucket_ref_errs[1]
+            local destination = single_err.vshard_err.destination
+            if destination and vshard_router.replicasets[destination] then
+                replicaset = vshard_router.replicasets[destination]
+            end
+        end
+
+        for _, bucket_ref_err in pairs(err.bucket_ref_errs) do
+            local bucket_id = bucket_ref_err.bucket_id
+            local vshard_err = bucket_ref_err.vshard_err
+            if vshard_err.name == 'WRONG_BUCKET' or
+               vshard_err.name == 'BUCKET_IS_LOCKED' or
+               vshard_err.name == 'TRANSFER_IS_IN_PROGRESS' then
+                vshard_router:_bucket_reset(bucket_id)
+            end
+        end
+    elseif err.name == 'MISSING_MASTER' and replicaset.locate_master ~= nil then
         replicaset:locate_master()
     end
 
@@ -147,8 +198,8 @@ function call.map(vshard_router, func_name, func_args, opts)
     while iter:has_next() do
         local args, replicaset, replicaset_id = iter:get()
 
-        local future, err = retry_call_with_master_discovery(replicaset, vshard_call_name,
-            func_name, args, call_opts)
+        local future, err = call_with_retry_and_recovery(vshard_router, replicaset, vshard_call_name,
+            func_name, args, call_opts, false)
 
         if err ~= nil then
             local result_info = {
@@ -221,9 +272,8 @@ function call.single(vshard_router, bucket_id, func_name, func_args, opts)
     local timeout = opts.timeout or const.DEFAULT_VSHARD_CALL_TIMEOUT
     local request_timeout = opts.mode == 'read' and opts.request_timeout or nil
 
-    local res, err = retry_call_with_master_discovery(replicaset, vshard_call_name,
-        func_name, func_args, {timeout = timeout,
-                               request_timeout = request_timeout})
+    local res, err = call_with_retry_and_recovery(vshard_router, replicaset, vshard_call_name,
+        func_name, func_args, {timeout = timeout, request_timeout = request_timeout}, true)
     if err ~= nil then
         return nil, wrap_vshard_err(vshard_router, err, func_name, nil, bucket_id)
     end
@@ -248,8 +298,8 @@ function call.any(vshard_router, func_name, func_args, opts)
     end
     local replicaset_id, replicaset = next(replicasets)
 
-    local res, err = retry_call_with_master_discovery(replicaset, 'call',
-        func_name, func_args, {timeout = timeout})
+    local res, err = call_with_retry_and_recovery(vshard_router, replicaset, 'call',
+        func_name, func_args, {timeout = timeout}, false)
     if err ~= nil then
         return nil, wrap_vshard_err(vshard_router, err, func_name, replicaset_id)
     end
--- a/crud/common/sharding/bucket_ref_unref.lua
+++ b/crud/common/sharding/bucket_ref_unref.lua
@@ -1,0 +1,165 @@
+--- Module to call vshard.storage.bucket_ref / vshard.storage.bucket_unref
+--- on write requests
+--- there are two modes: safe and fast. on safe mode module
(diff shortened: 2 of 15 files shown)


crud/common/sharding/bucket_ref_unref.lua
```diff
@@ -0,0 +1,183 @@
...
+--- Slow bucket_refrw_many that calls bucket_refrw for every bucket and aggregates errors
+--- @param bucket_ids table<number, boolean>
+function M._bucket_refrw_many(bucket_ids)
+    local bucket_ref_errs = {}
+    local reffed_bucket_ids = {}
+    for bucket_id in pairs(bucket_ids) do
+        local ref_ok, bucket_refrw_err = safe_methods.bucket_refrw(bucket_id)
+        if not ref_ok then
+
+            table.insert(bucket_ref_errs, bucket_refrw_err.bucket_ref_errs[1])
```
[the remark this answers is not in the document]
ita-sammann: This error is returned by our own code and has bucket_ref_errs with at least one record in all cases

crud/common/sharding/bucket_ref_unref.lua
```diff
@@ -0,0 +1,183 @@
...
+local M = {
+    BucketRefError = errors.new_class('bucket_ref_error', {capture_stack = false})
+}
+
+local function make_bucket_ref_err(bucket_id, vshard_ref_err)
+    local err = M.BucketRefError:new(M.BucketRefError:new(
+        "failed bucket_ref: %s, bucket_id: %s",
+        vshard_ref_err.name,
+        bucket_id
+    ))
```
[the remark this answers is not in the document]
ita-sammann: Fixed

vakhov (approved):

crud.lua
```diff
@@ -23,6 +23,7 @@ local readview = require('crud.readview')
 local schema = require('crud.schema')
```
Serpentian: Guys, the commits are absolutely chaotic, this is not the way, we develop open-source modules. We should rebase that branch in master to make the commit history linear, and not just merge it. So, all commits like [Minor changes according to review comments](https://github.com/tarantool/crud/pull/467/commits/bb11b86b86c57abbc9cf3fbd93f9a7568c47091d) will be visible to our users. You can check out, how Georgy Moiseev did it: proper commits, proper commit messages, tests in every commit.

Now, I'm forced to review more than 1k lines in one step, which is very inconvenient and increases the chanse of missed bugs. And our users won't be able to check the individual commits, if they want to. Of course, it's up to you, since I'm not the maintainer of that module, but it's just not nice to develop and merge code like that.

IMHO, the features should be properly split between commits, every commit must include the associated tests, proper commit messages and mentioning of the https://github.com/tarantool/crud/issues/448 ticket. Of course, refactoring or code moving should be in the separate commits

crud/common/rebalance.lua
```diff
@@ -0,0 +1,211 @@
...
+            "when vshard router cache was cleared last time"
+    )
+
+    metrics.register_callback(function()
+        router_cache_length_gauge:set(router_cache_length())
+        router_cache_last_clear_ts_gauge:set(router_cache_last_clear_ts())
+    end)
+end
+
+M.init = rebalance_init
```
Serpentian: Nit: let's make the code consistent with all other modules, it seems we tend to make the variable with proper name and assign methods to it `.`:

https://github.com/tarantool/crud/blob/c52a6f588bdadc04d5b3e9f438d78851aa7e8055/crud/stats/local_registry.lua#L39-L39

crud/common/rebalance.lua
```diff
@@ -0,0 +1,211 @@
...
+M.safe_mode_status = safe_mode_status
+M.safe_mode_enable = safe_mode_enable
+M.safe_mode_disable = safe_mode_disable
+M.register_safe_mode_enable_hook = register_enable_hook
+M.remove_safe_mode_enable_hook = remove_enable_hook
+M.register_safe_mode_disable_hook = register_disable_hook
+M.remove_safe_mode_disable_hook = remove_disable_hook
+
+M.router = {
```
Serpentian: Nit: api naming is incosistent, `storage_api` vs `router`

crud/common/rebalance.lua
```diff
@@ -0,0 +1,211 @@
...
+local SAFE_MOD_ENABLE_EVENT = '_crud.safe_mode_enable'
+
+local M = {
+    safe_mode = false,
+    safe_mode_enable_hooks = {},
+    safe_mode_disable_hooks = {},
+    _router_cache_last_clear_ts = fiber.time()
+}
+
+local function create_space()
```
Serpentian: AFAICS, the space has general name `_crud_settings`, its schema is also general. So, I suppose, in the future, we may have new crud settings, which we want to persist. So, the space has nothing to do with `rebalance` module, which just uses the space. And it seems its creation should be done somewhere in more general place, e.g. `crud.schema` or smth like that.

crud/common/rebalance.lua
```diff
@@ -0,0 +1,211 @@
...
+        return
+    end
+    if (op == 'INSERT' and new.status == vshard_consts.BUCKET.RECEIVING) or
+            (op == 'REPLACE' and new.status == vshard_consts.BUCKET.SENDING) then
+        box.broadcast(SAFE_MOD_ENABLE_EVENT, true)
+    end
+end
+
+local function register_enable_hook(func)
+    M.safe_mode_enable_hooks[func] = true
```
Serpentian: Why do we need that variable and why cannot we just set `on_replace` trigger for `_bucket` on enable and remove it on disable?

crud/common/rebalance.lua
```diff
@@ -0,0 +1,211 @@
...
+
+    box.watch(SAFE_MOD_ENABLE_EVENT, function(_, do_enable)
+        if box.info.ro or not do_enable then
+            return
+        end
+        safe_mode_enable()
+    end)
+end
+
+local function router_cache_clear()
```
Serpentian: Aren't we supposed to check, that rebalancing is ended and we have no buckets in forbidden states?

crud/common/rebalance.lua
```diff
@@ -0,0 +1,211 @@
...
+local utils = require('crud.common.utils')
+
+local has_metrics_module, metrics = pcall(require, 'metrics')
+
+local SETTINGS_SPACE_NAME = '_crud_settings'
+local SAFE_MOD_ENABLE_EVENT = '_crud.safe_mode_enable'
+
+local M = {
+    safe_mode = false,
+    safe_mode_enable_hooks = {},
```
Serpentian: We can use `('internal.trigger')`for things like that

crud/common/sharding/bucket_ref_unref.lua
```diff
@@ -0,0 +1,183 @@
...
+--- @param bucket_ids table<number, boolean>
+function M._bucket_refrw_many(bucket_ids)
+    local bucket_ref_errs = {}
+    local reffed_bucket_ids = {}
+    for bucket_id in pairs(bucket_ids) do
+        local ref_ok, bucket_refrw_err = safe_methods.bucket_refrw(bucket_id)
+        if not ref_ok then
+
+            table.insert(bucket_ref_errs, bucket_refrw_err.bucket_ref_errs[1])
+            goto continue
```
Serpentian: Why do we continue on error and then just unref all of them? We should go to unref stright away

Serpentian (changes requested):

crud.lua
ita-sammann: > Guys, the commits are absolutely chaotic

Of course these commits will not go to master, I will re-split them before merging the PR.

> Now, I'm forced to review more than 1k lines in one step

My bad. Never thought of this PR as of multiple features that can be reviewed separately.

crud.lua
```diff
@@ -158,8 +159,23 @@ crud.readview = readview.new
...
 
+crud.rebalance = {}
+
+-- @refer rebalance.router_cache_clear
+-- @function router_cache_clear
+crud.rebalance.router_cache_clear = rebalance.router.cache_clear
+
+-- @refer rebalance.router_cache_length
+-- @function router_cache_length
+crud.rebalance.router_cache_length = rebalance.router.cache_length
```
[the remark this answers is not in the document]
ita-sammann: I think they will be useful in TCM/Cartridge to indicate that cache has actually been cleared and when did it happen.

test/entrypoint/srv_batch_operations/cartridge_init.lua
```diff
@@ -11,11 +11,11 @@ local cartridge = require('cartridge')
 if package.setsearchroot ~= nil then
     package.setsearchroot()
 else
-    package.path = package.path .. debug.sourcedir() .. "/?.lua;"
+    package.path = package.path .. debug.sourcedir() .. "/?.lua;" .. debug.sourcedir() .. "/?/init.lua;"
```
Serpentian: Nit: please, let's describe the motivation in the commit msg and the gh ticket number, if it's applicable

test/vshard_helpers/vtest.lua
```diff
@@ -501,15 +501,26 @@ local function cluster_new(g, cfg)
...
+    -- crud.init_storage() being finished on the corresponding master.
+    -- So masters go first in all_servers list.
+    for _, server in pairs(masters) do
+        table.insert(all_servers, server)
+    end
+    for _, server in pairs(replicas) do
+        table.insert(all_servers, server)
+    end
+
+
```
Serpentian: Nit: isn't two empty lines too much?

crud/common/rebalance.lua
```diff
@@ -0,0 +1,197 @@
...
+function rebalance.init()
+    local stored_safe_mode
+    if not box.info.ro then
+        if box.space[SAFE_MODE_SPACE] == nil then
+            create_space()
+            create_trigger()
+        end
+        stored_safe_mode = box.space[SAFE_MODE_SPACE]:get{ 'status' }
+    else
+        while box.space[SAFE_MODE_SPACE] == nil or box.space[SAFE_MODE_SPACE].index[0] == nil do
```
Serpentian: What if none of the instances are in `RW`, when `rebalance.init` is called. All of them will just indefinitely wait for replication of the space. We need to wait either for rw or for space here

crud/common/rebalance.lua
```diff
@@ -0,0 +1,197 @@
...
+            fiber.sleep(0.05)
+        end
+        create_trigger()
+        stored_safe_mode = box.space[SAFE_MODE_SPACE]:get{ 'status' }
+    end
+
+    if stored_safe_mode and stored_safe_mode.value then
+        _safe_mode_enable()
```
Serpentian: Why do we need these, when they're anyway will be run from the on commt trigger? Or are we afraid, that we won't manage to set the on_replace trigger in time?

crud/insert.lua
```diff
@@ -42,15 +43,33 @@ local function insert_on_storage(space_name, tuple, opts)
...
-    -- add_space_schema_hash is true only in case of insert_object
-    -- the only one case when reloading schema can avoid insert error
-    -- is flattening object on router
-    return schema.wrap_box_space_func_result(space, 'insert', {tuple}, {
-        add_space_schema_hash = opts.add_space_schema_hash,
-        field_names = opts.fields,
-        noreturn = opts.noreturn,
-        fetch_latest_metadata = opts.fetch_latest_metadata,
-    })
+    local function make_insert()
```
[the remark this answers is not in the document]
Serpentian: I've meant, that for example `insert` and `replace` has the same code:

```lua
    local bucket_id = tuple[utils.get_bucket_id_fieldno(space)]
    local ref_ok, bucket_ref_err, unref = bucket_ref_unref.bucket_refrw(bucket_id)
    if not ref_ok then
        return nil, bucket_ref_err
    end
    
 < ...>
 
     local unref_ok, err_unref = unref(bucket_id)
    if not unref_ok then
        return nil, err_unref
    end
```

Can't we make a wrappers for the functions, which will ref and unref the buckets?

Serpentian (commented):
Thank you for the fixes, way cleaner now! We're moving towards the end

test/entrypoint/srv_batch_operations/cartridge_init.lua
ita-sammann: Added description to the commit message, there are no related tickets though.

crud/common/rebalance.lua
```diff
@@ -0,0 +1,197 @@
+local fiber = require('fiber')
```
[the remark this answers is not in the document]
ita-sammann: Done

crud.lua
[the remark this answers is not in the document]
ita-sammann: Fair point. Removed these methods for now.

crud/common/rebalance.lua
ita-sammann: Okay. Renamed this space back and moved it's creation to `crud.schema` module.

crud/common/rebalance.lua
ita-sammann: Thanks! Added check that node became `RW`.

crud/insert_many.lua
```diff
@@ -98,7 +113,11 @@ local function insert_many_on_storage(space_name, tuples, opts)
                     return nil, errs, replica_schema_version
                 end
 
+                local unref_ok, bucket_unref_err = unref(bucket_ids)
                 box.commit()
+                if not unref_ok then
+                    return nil, bucket_unref_err
+                end
```
Serpentian: Nit: in 3 places, where unref stage is repeated several times. Can we do `goto error` here? Smth like:

```lua
   if smth then
       goto error
   end
   <...>
error:
    <unref logic>
    return inserted_tuples, errs, replica_schema_version
```

In order not to duplicate the code

Serpentian (approved):
I have only nit comments left. LGTM! Thank you for the fixes a lot.

crud/common/call.lua
```diff
@@ -1,11 +1,14 @@
 local errors = require('errors')
+local log = require('log')
 
 local call_cache = require('crud.common.call_cache')
 local dev_checks = require('crud.common.dev_checks')
 local utils = require('crud.common.utils')
 local sharding_utils = require('crud.common.sharding.utils')
-local fiber_clock = require('fiber').clock
+local fiber = require('fiber')
```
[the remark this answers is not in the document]
ita-sammann: We also use other methods from `fiber`, like `name`, `info` and `kill`. In `call_on_storage_fast` and `rebalance.on_safe_mode_toggle`.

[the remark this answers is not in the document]
ita-sammann: Ok, returned `fiber_clock`

PR #475, opened 2025-12-30
safe/fast mode: remove fiber kill

### LINKED ISSUE #473: Unsafe fiber termination by name causes killing unrelated fibers
After upgrading CRUD to 1.7.0, we observed unstable behavior on storage nodes when the expirationd role is enabled. During role application, the fiber executing apply() may be unexpectedly killed. This happens while CRUD is performing storage calls in “fast mode”. As a result, role application is interrupted and fail. The issue is caused by how CRUD currently identifies and terminates its internal fibers.

**Root cause**
CRUD relies on fiber.name() to identify internal “fast-mode” fibers and kills fibers by matching their name. This assumption is unsafe for several reasons:
 - changed fiber.name persists after crud request completed
 - As a result, a different, unrelated fiber may later be killed by crud

Satbek:
This pr improves performance of current crud operations and remove fiber killing in safe mode start. 

In this pr added:
- remove fiber killing on safe mode switch
- move safe mode switch from `on_commit` trigger to `on_replace` trigger
- crud methods run in safe mode for vinyl spaces

## remove fiber killing on safe mode switch 
In fast mode, every fiber was named **"fast"** so it could be killed when safe mode was enabled.

However, CRUD storage methods do not yield until a space operation is performed.  
The rebalancing trigger is executed on `insert`/`replace` into the `_bucket` space; this operation **does yield** and switches the implementation of the `bucket_ref` / `bucket_unref` functions.

Therefore, once a fiber starts in fast mode, it remains in fast mode until the first `box.space` operation:
- a **write** for memtx

As a result, there is no need to mark and kill iproto fibers that handle fast requests, because there is no situation in which a fiber remains in fast mode while rebalancing is in progress.

Yield checks were added to the tests (`yield_checks`) to ensure that no yields occur during these operations.

## move safe mode switch from `on_commit` to `on_replace` trigger

This change ensures that with turned on MVCC crud method will be executed on safe mode, when replace in `_bucket` space happened but not commited yet.

##  crud methods run in safe mode for vinyl spaces

vinyl can make more than one yield in box operation. So were made decision to run crud method for vinyl space always in 
safe mode.

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Closes #473

### DIFF
--- a/crud/delete.lua
+++ b/crud/delete.lua
@@ -44,7 +44,7 @@ local function delete_on_storage(space_name, key, field_names, opts)
         return nil, err
     end
 
-    local ref_ok, bucket_ref_err, unref = bucket_ref_unref.bucket_refrw(opts.bucket_id)
+    local ref_ok, bucket_ref_err, unref = bucket_ref_unref.bucket_refrw(opts.bucket_id, space.engine)
     if not ref_ok then
         return nil, bucket_ref_err
     end
@@ -58,7 +58,7 @@ local function delete_on_storage(space_name, key, field_names, opts)
         fetch_latest_metadata = opts.fetch_latest_metadata,
     }, space, key)
 
-    local unref_ok, err_unref = unref(opts.bucket_id)
+    local unref_ok, err_unref = unref(opts.bucket_id, space.engine)
     if not unref_ok then
         return nil, err_unref
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Stop marking/killing fast-mode iproto fibers on safe-mode switch.
   Storage ops are asserted by `yield_checks` in tests.
 * Move switch to safe mode from `on_commit` trigger to `on_replace` trigger.
+* vinyl spaces always work in safe mode.
 
 ## [1.7.0] - 26-12-25
 
--- a/crud/common/sharding/bucket_ref_unref.lua
+++ b/crud/common/sharding/bucket_ref_unref.lua
@@ -125,13 +125,23 @@ safe_methods = {
     bucket_unrefrw_many = bucket_ref_unref._bucket_unrefrw_many,
 }
 
+local function make_fast_method(method_name)
+    return function(arg, space_engine)
+        if space_engine == 'vinyl' then
+            -- vinyl always works in safe mode
+            return safe_methods[method_name](arg)
+        end
+        return bucket_ref_unref._fast()
+    end
+end
+
 fast_methods = {
-    bucket_refrw = bucket_ref_unref._fast,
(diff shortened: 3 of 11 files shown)


crud/delete.lua
```diff
@@ -58,6 +62,8 @@ local function delete_on_storage(space_name, key, field_names, opts)
         fetch_latest_metadata = opts.fetch_latest_metadata,
     }, space, key)
 
+    finish()
```
vakhov: if an unexpected Lua error is thrown between yield_checks.start() and finish() (for example, inside schema.wrap_func_result()), the fiber would remain registered in yield_checks.registry. could this potentially affect subsequent tests and lead to confusing errors?

crud/common/call.lua
```diff
@@ -18,38 +16,13 @@ local CallError = errors.new_class('CallError')
...
 local CALL_FUNC_NAME = 'call_on_storage'
 local CRUD_CALL_FUNC_NAME = utils.get_storage_call(CALL_FUNC_NAME)
-local CRUD_CALL_FIBER_NAME = CRUD_CALL_FUNC_NAME .. '/fast'
 
 local call = {}
 
-local function call_on_storage_safe(run_as_user, func_name, ...)
+local function call_on_storage(run_as_user, func_name, ...)
     return box.session.su(run_as_user, call_cache.func_name_to_func(func_name), ...)
 end
```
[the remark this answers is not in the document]
Satbek: These are internal implementation details, and users do not need to be aware of them.

The API with subscriptions to rebalancing events remains unchanged, so I do not see a reason to explicitly document that some internal code was removed.

crud/common/schema.lua
```diff
@@ -216,6 +217,8 @@ end
 -- `add_space_schema_hash` is true
 function schema.wrap_func_result(space, func, opts, ...)
     dev_checks('table', 'function', 'table')
+    yield_checks.check_no_yields()
```
a1div0: Why do we check for no context switches **before** executing a user method?

crud/common/yield_checks.lua
```diff
@@ -0,0 +1,51 @@
...
+	end
+	return res_err, ...
+end
+
+local function guard(f, ...)
+	register_fiber()
+	return finish(pcall(f, ...))
+end
+
+if os.getenv("TARANTOOL_CRUD_ENABLE_INTERNAL_CHECKS") == "ON" then
```
a1div0: Why do we enable functionality using an environment variable instead of a global setting?

ita-sammann: Why don't we add this var to the test environment by default?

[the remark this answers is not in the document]
Satbek: The variable is already set and value is "ON".

vakhov (approved):

PR #495, opened 2026-05-05
fix: handle nil bucket_id in storage get/update/delete operations

p0rtale:
Starting from crud 1.7.0, storage asserts if `opts.bucket_id` is nil during `get`, `update`, or `delete` operations. Since older routers (< 1.7.0) do not pass `bucket_id`, this causes protocol incompatibility.

This PR adds a nil check to skip bucket ref when `bucket_id` is missing, restoring backward compatibility with old routers.

It also introduces a rate-limited warning and the `tnt_crud_storage_nil_bucket_id_compat_total` metric to track
this state.

- [x] Tests
- [x] Changelog
- [ ] Documentation

Closes TNTP-7622

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+* Fixed storage compatibility with routers < 1.7.0 by handling nil bucket_id
+in get, update, and delete operations.
+
 ## [1.7.4] - 12-02-26
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 * Allow read-only operations (`get`, `select`, `pairs`, `count`, `min`, `max`)
   to execute successfully using healthy replicas when all cluster masters are down.
+* Add `tnt_crud_storage_nil_bucket_id_compat_total` metric to track operations
+  performed without bucket_ref (legacy router compatibility mode).
+* Fixed storage compatibility with routers < 1.7.0 by handling nil bucket_id
+  in get, update, and delete operations. Now storage skips bucket referencing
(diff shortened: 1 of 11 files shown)


vakhov (commented):
Thanks for the PR. I have suggestions here:
 - the workaround itself looks reasonable as a temporary solution for rolling cluster upgrades, but it should be made observable. Right now, when bucket_id == nil, storage silently skips bucket_ref, so during a partial upgrade some traffic goes through a path with weaker guarantees, and operations won’t see it.
 - maybe it’s worth adding a rate-limited warning through a shared helper, for example warn_nil_bucket_id_compat_once(operation, space_name, space.engine), to avoid duplicating the same logic across get/update/delete. The message could be something like: crud.%s_on_storage called without bucket_id; old router compatibility mode is used, bucket_ref is skipped, rebalance safety is reduced until routers are upgraded; space=%s engine=%s.
- it may also be worth adding a storage-side counter: tnt_crud_storage_nil_bucket_id_compat_total with labels like operation, engine, and possibly safe_mode.
- also, it may be worth documenting the upgrade workaround explicitly for clusters upgrading from < 1.7.0 to >= 1.7.0. The docs should say that during a rolling upgrade storages should be upgraded first, and routers afterwards. While old routers are still running, storages may enter the compatibility path where bucket_id is missing and bucket_ref is skipped, so should expect the warning/counter above until all routers are upgraded.

ita-sammann (commented):
It bothers me that where we have just added safe mode in 1.7.0 and now we are already punching holes in it. AFAIK 1.7.x router can work with 1.6.x storages. So user can just update routers first and update storages after that. And we don't need to add this check for nil bucket_id.

I suggest that we add instruction how to update cluster to the newer crud versions to README.md:
 - Update storages from pre-1.6.0 (role-based model) to 1.6.1
 - Update routers from pre-1.6.0 to 1.7.4 (safe mode)
 - Update storages from 1.6.1 to 1.7.4

vakhov (approved):

ita-sammann:
We have a convention that storages must be updated first and routesr should be updated after that. So we do need storage APIs to be compatible with older version of routers.

Satbek (approved):

PR #497, opened 2026-05-24
fix: allow read-only operations when all masters are down

p0rtale:
Read-only operations (get, select, pairs, count, min, max) used to fail with connection errors if all masters in the cluster were unavailable, even when healthy replicas were up and failover hadn't processed yet.

To resolve this, the following improvements were made:
- Introduced a `read_only` flag to `utils.get_space[s]` to fetch cluster schema from any healthy replica if masters are down.
- Updated `get`, `select`, `pairs`, `count`, `min`, `max` to use this new flag.
- Rewrote `call.any` to iterate through all replicasets and utilize vshard's `callro` instead of `call` to fetch metadata from replicas.

Closes TNTP-7102

### DIFF
--- a/crud/common/call.lua
+++ b/crud/common/call.lua
@@ -270,19 +270,39 @@ function call.any(vshard_router, func_name, func_args, opts)
     if replicasets == nil then
         return nil, CallError:new("Failed to get router replicasets: %s", err.err)
     end
-    local replicaset_id, replicaset = next(replicasets)
 
-    local res, err = call_with_retry_and_recovery(vshard_router, replicaset, 'call',
-        func_name, func_args, {timeout = timeout}, false)
-    if err ~= nil then
-        return nil, wrap_vshard_err(vshard_router, err, func_name, replicaset_id)
+    local last_replicaset_id = nil
+    local last_err = nil
+
+    local deadline = fiber_clock() + timeout
+
+    for replicaset_id, replicaset in pairs(replicasets) do
+        local wait_timeout = deadline - fiber_clock()
+
+        local is_timeout = wait_timeout < 0
+        if is_timeout then
+            wait_timeout = 0
+        end
+
+        local res, err = call_with_retry_and_recovery(vshard_router, replicaset, 'callro',
+            func_name, func_args, {timeout = wait_timeout}, false)
+
+        if err == nil then
+            if res == box.NULL then
+                return nil
+            end
+            return res
+        end
+
+        last_replicaset_id = replicaset_id
+        last_err = err
+
+        if is_timeout then
+            break
+        end
     end
 
-    if res == box.NULL then
-        return nil
-    end
-
-    return res
+    return nil, wrap_vshard_err(vshard_router, last_err, func_name, last_replicaset_id)
 end
 
 return call
--- a/crud/common/utils.lua
+++ b/crud/common/utils.lua
@@ -123,10 +123,44 @@ local function get_replicaset_by_replica_id(replicasets, id)
     return nil, nil
 end
 
-function utils.get_spaces(vshard_router, timeout, replica_id)
-    local replicasets, replicaset, replicaset_id, master
+local function find_any_healthy_replica_conn(replicasets)
+    for _, replicaset in pairs(replicasets) do
+        for _, replica in pairs(replicaset.replicas) do
+            if replica:is_connected() then
+                return replica.conn
+            end
+        end
+    end
+    return nil
+end
 
-    timeout = timeout or const.DEFAULT_VSHARD_CALL_TIMEOUT
+--- Returns all spaces existing in the schema.
+--
+-- @function get_spaces
+--
+-- @param table vshard_router
+--  A vshard router instance to route requests.
+--
+-- @tparam ?number opts.timeout
+--  Function call timeout. If not specified, `const.DEFAULT_VSHARD_CALL_TIMEOUT` is used.
+--
+-- @tparam ?boolean opts.read_only
+--  If true, the function will try to fetch spaces from any available healthy replica.
+--
+-- @tparam ?string opts.replica_id
+--  The exact replica ID to fetch the replicaset from.
+--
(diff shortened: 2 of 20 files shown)


ita-sammann (approved):

crud/common/utils.lua
```diff
@@ -123,8 +123,19 @@ local function get_replicaset_by_replica_id(replicasets, id)
     return nil, nil
 end
 
-function utils.get_spaces(vshard_router, timeout, replica_id)
-    local replicasets, replicaset, replicaset_id, master
+local function find_any_healthy_replica_conn(replicasets)
```
Satbek, line 126: why not https://www.tarantool.io/en/doc/latest/reference/reference_lua/net_box/#conn-is-connected ?

Satbek: https://github.com/tarantool/vshard/blob/0fd36939/vshard/replicaset.lua#L341-L362 vshard checks connections like this

crud/len.lua
```diff
@@ -62,7 +62,7 @@ function len.call(space_name, opts)
     end
 
     local results, err = call.map(vshard_router, CRUD_LEN_FUNC_NAME, {space_name}, {
-        mode = 'write',
```
Satbek, line 65: Why?

crud/common/utils.lua
```diff
@@ -184,8 +208,8 @@ function utils.get_spaces(vshard_router, timeout, replica_id)
     return master.conn.space, nil, master.conn.schema_version
 end
 
-function utils.get_space(space_name, vshard_router, timeout, replica_id)
-    local spaces, err, schema_version = utils.get_spaces(vshard_router, timeout, replica_id)
+function utils.get_space(space_name, vshard_router, timeout, replica_id, read_only)
```
Satbek: may be add  `timeout, replica_id, read_only` args in one opts table argument?

crud/len.lua
p0rtale: Otherwise `len()` will fail if masters are unavailable

Satbek: > Otherwise `len()` will fail if masters are unavailable

may be it's ok? We need to know exact number of tuples on master. No masters -> no result.

you may add ticket for some flag for len method to get len from replicas when master is absent

crud/common/utils.lua
p0rtale: Used vshard's `replica:is_connected`: https://github.com/tarantool/vshard/blob/5a03168d11e717e52a89aca9b28c453e3f737be4/vshard/replicaset.lua#L1441

crud/common/utils.lua
p0rtale: Done

crud/len.lua
p0rtale: Added the read only options (`mode`, `request_timeout`, `prefer_replica`, `balance`) to allow users to explicitly fetch the length from replicas if needed.
The default behavior is unchanged (mode = 'write'), so len will still fail without masters to guarantee exact data.

crud/common/call.lua
```diff
@@ -270,19 +270,26 @@ function call.any(vshard_router, func_name, func_args, opts)
...
-        func_name, func_args, {timeout = timeout}, false)
-    if err ~= nil then
-        return nil, wrap_vshard_err(vshard_router, err, func_name, replicaset_id)
-    end
+    local last_replicaset_id = nil
+    local last_err = nil
 
-    if res == box.NULL then
-        return nil
+    for replicaset_id, replicaset in pairs(replicasets) do
```
Satbek, line 279: If we send requests to all replicaset(s), then those requests need to share a common overall timeout. Otherwise, the effective timeout for a request right now is the number of replicasets multiplied by the per-request timeout.

In general, how necessary is it to try to go through all replicasets? What do you think - maybe just `callro` would be enough?

Satbek: also there is a 'request_timeout' option, may be it might be useful

crud/common/utils.lua
```diff
@@ -123,10 +123,23 @@ local function get_replicaset_by_replica_id(replicasets, id)
...
+        for _, replica in pairs(replicaset.replicas) do
+            if replica:is_connected() then
+                return replica.conn
+            end
+        end
+    end
+    return nil
+end
+
+function utils.get_spaces(vshard_router, opts)
```
Satbek, line 158: write comment about available options please

crud/len.lua
Satbek: Can you add this change as a separate commit? Because it changes the method options rather than fixing a bug.

Or would it be better to make a separate MR?

crud/common/utils.lua
p0rtale: done

crud/len.lua
p0rtale: Removed len-related changes from this commit. I'll create a separate PR for them

crud/common/call.lua
p0rtale: Fixed timeout

p0rtale: > In general, how necessary is it to try to go through all replicasets? What do you think - maybe just `callro` would be enough?

We iterate through all replicasets to provide a fallback mechanism in case the whole replicaset is down.

crud/common/call.lua
```diff
@@ -270,19 +270,33 @@ function call.any(vshard_router, func_name, func_args, opts)
...
+    local last_err = nil
 
-    if res == box.NULL then
-        return nil
+    local deadline = fiber_clock() + timeout
+
+    for replicaset_id, replicaset in pairs(replicasets) do
+        local wait_timeout = deadline - fiber_clock()
+        if wait_timeout < 0 then
+            wait_timeout = 0
```
Satbek, line 284: break here?

p0rtale: added

Satbek (approved):

PR #500, opened 2026-06-25
Yield every 1000 iterations when fetching sharding keys and funcs

ita-sammann:
Yield every 1000 iterations when fetching sharding keys and functions so the server stays responsive when there are millions of records.

### DIFF
--- a/crud/common/sharding/sharding_metadata.lua
+++ b/crud/common/sharding/sharding_metadata.lua
@@ -16,6 +16,7 @@ local FetchShardingMetadataError = errors.new_class('FetchShardingMetadataError'
 
 local FETCH_FUNC_NAME = 'fetch_on_storage'
 local CRUD_FETCH_FUNC_NAME = utils.get_storage_call(FETCH_FUNC_NAME)
+local FETCH_BATCH_SIZE = 1000
 
 local sharding_metadata_module = {}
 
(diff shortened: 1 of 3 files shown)


crud/common/sharding/sharding_metadata.lua
```diff
@@ -77,16 +79,26 @@ function sharding_metadata_module.fetch_on_storage()
                          'Ensure that you did a proper cleanup after DDL space drop.',
                          space_name)
             end
+
+            if i % 1000 == 0 then
```
vakhov: could we extract 1000 into a named constant to avoid duplicating the magic number?

crud/common/sharding/sharding_metadata.lua
```diff
@@ -77,16 +79,26 @@ function sharding_metadata_module.fetch_on_storage()
...
+                fiber.yield()
```
vakhov, line 85: on a cold cache, get_sharding_(func|key)_hash() calls init_cache(), which scans the whole DDL space without yielding. So the first fetch may still block. should we add yielding inside init_cache() instead?

vakhov (commented):
Thanks for the PR! I left a few comments below.

adding explicit yields changes the consistency semantics of this function: concurrent DDL updates may now interleave with the scan, so metadata_map can contain entries from different schema states, including a sharding key and function fetched at different moments. is this eventual consistency guaranteed to be handled by the hash/reload mechanism? could we add a regression test for the cold-cache case to verify that another fiber gets scheduled during a large metadata scan?

crud/common/sharding/sharding_metadata.lua
ita-sammann: Good idea, thanks! Extracted

crud/common/sharding/sharding_metadata.lua
ita-sammann: Added yield there, thanks for noticing!

ita-sammann:
> adding explicit yields changes the consistency semantics of this function

@vakhov, actually I don't like the idea of this yields at all. If there is a sane amount of sharding keys there are no problems.
Maybe we should just add migration helpers (tarantool/migrations#88) and update the documentation to highlight that user should drop the corresponding sharding key when dropping the space. And just close this PR.

PR #504, opened 2026-07-14
ci: shard coverage and upload report to S3

p0rtale:
Coverage reports are now uploaded to S3 on pushes to `master` and tag releases.

Since a single CE coverage job took too long, the CE coverage run is sharded across 6 parallel jobs: each job runs a subset of the test files with luacov and publishes its `luacov.stats*.out` as an artifact, then a dedicated job merges them with `test/luacov-merger.lua` and uploads the generated report.

Also server wait timeouts became configurable via `LUATEST_SERVER_WAIT_TIMEOUT` (set to 180 in coverage jobs) to accommodate the slower luacov-instrumented storage.

Additionally, `get_other_storage_bucket_id` now retries `vshard.storage.buckets_info` until it responds or the deadline expires, instead of calling `next()` on a `nil` result when the storage times out under coverage.

### DIFF
--- a/.github/workflows/test_on_push.yaml
+++ b/.github/workflows/test_on_push.yaml
@@ -46,6 +46,7 @@ jobs:
             vshard-version: "0.1.41"
           - tarantool-version: "master"
             vshard-version: "master"
+            s3-upload: true
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/test_on_push.yaml
```diff
@@ -130,13 +131,28 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Run tests
-        if: matrix.coveralls != true # See https://github.com/actions/runner/issues/1173
+        if: matrix.coveralls != true && matrix.s3-upload != true # See https://github.com/actions/runner/issues/1173
```
vakhov: > For regular branch pushes, the `s3-upload` configuration won’t run any tests. I suggest changing the condition to:
> ```yaml
> if: >-
>   matrix.coveralls != true &&
>   (matrix.s3-upload != true ||
>    (github.ref_type != 'tag' && github.ref_name != 'master'))
> ```

vakhov (approved):
lgtm

vakhov (approved):

vakhov (commented):
No questions about the PR itself. But I noticed one thing in `test/luacov-merger.lua` that is worth checking.

Currently, the counters are written like this:
```lua
file:write(table.concat(counts, " ") .. "\n")
```

When reading the stats, LuaCov expects a space after every number, including the last one. If it encounters a newline instead, it exits the loop without saving the last counter. For example, the merger writes `5 7 9`, but LuaCov only reads `5` and `7`. This can cause an executed line to be reported as uncovered.

I suggest adding a space before the newline:
```lua
file:write(table.concat(counts, " ") .. " \n")
```

Here is [the space check in LuaCov](https://github.com/lunarmodules/luacov/blob/v0.17.0/src/luacov/stats.lua#L37-L47). Version 0.13.0, which this workflow uses, has the same logic.

a1div0 (approved):

vakhov (approved):

PR #510, opened 2026-08-14
feat: add atomic_batch operation

p0rtale:
Add `crud.atomic_batch()` operation to execute heterogeneous CRUD operations (`get`, `insert`, `replace`, `update`, `upsert`, `delete`) atomically within a single replicaset transaction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ It also provides the `crud-storage` and `crud-router` roles for
   - [Replace many](#replace-many)
   - [Upsert](#upsert)
   - [Upsert many](#upsert-many)
+  - [Atomic batch](#atomic-batch)
   - [Select](#select)
     - [Select conditions](#select-conditions)
   - [Pairs](#pairs)
@@ -1116,6 +1117,67 @@ errs[3].operation_data -- {71, 1802, "Oksana", 29}
 errs[4].class_name     -- NotPerformedError
 errs[4].err            -- 'Operation with tuple was rollback'
 errs[4].operation_data -- {92, 2040, "Artur", 29}
+```
+
+### Atomic batch
+
+```lua
+local result, err = crud.atomic_batch(operations, opts)
+```
+
+Runs an ordered list of CRUD operations (`get`, `insert`, `replace`, `update`,
+`upsert`, `delete`) in a single transaction on one replicaset.
+
+where:
+
+* `operations` (`table`) - ordered array of operation descriptors
+* `opts`:
+  * `timeout` (`?number`) - `vshard.call` timeout (seconds), default `2`
+  * `noreturn` (`?boolean`) - suppress successful operation results
+  * `fields` (`?table`) - per-space output projection:
+    `{[space_name] = {field1, field2, ...}}`
+
+Returns `{metadata = {[space_name] = format}, data = {op_results...}}` or
+`nil, err`.
+
+**Example:**
+
+```lua
+local operations = {
+    {
+        type = 'insert',
+        space = 'orders',
+        object = {id = 101, customer_id = 10, status = 'new'}
+    },
+    {
+        type = 'update',
+        space = 'customers',
+        key = {101},
+        operations = {{'+', 'orders_count', 1}}
+    },
+    {
+        type = 'get',
+        space = 'orders',
+        key = {101}
+    },
+}
+
+local res, err = crud.atomic_batch(operations, {
+    timeout = 1.5,
+    fields = {
+        orders = {'id', 'status'},
+        customers = {'id', 'orders_count'},
+    },
+})
+
+if err ~= nil then
+    -- whole batch is rolled back on any failure
+    error(err)
+end
+
+-- res.data[1] -> inserted order tuple (projected by fields.orders)
+-- res.data[2] -> updated customer tuple (projected by fields.customers)
+-- res.data[3] -> get result for order
 ```
 
 ### Select
--- a/crud/atomic_batch.lua
+++ b/crud/atomic_batch.lua
@@ -1,0 +1,647 @@
+local checks = require('checks')
+local errors = require('errors')
+
+local call = require('crud.common.call')
+local const = require('crud.common.const')
+local utils = require('crud.common.utils')
+local sharding = require('crud.common.sharding')
+local sharding_key_module = require('crud.common.sharding.sharding_key')
+local sharding_metadata_module = require('crud.common.sharding.sharding_metadata')
+local dev_checks = require('crud.common.dev_checks')
+local schema = require('crud.common.schema')
+local bucket_ref_unref = require('crud.common.sharding.bucket_ref_unref')
+local clock = require('clock')
+local stats = require('crud.stats')
+
+local AtomicBatchError = errors.new_class('AtomicBatchError', {capture_stack = false})
+local AtomicBatchValidationError = errors.new_class('AtomicBatchValidationError', {capture_stack = false})
+
+local atomic_batch = {}
+
+local ATOMIC_BATCH_FUNC_NAME = 'atomic_batch_on_storage'
+local CRUD_ATOMIC_BATCH_FUNC_NAME = utils.get_storage_call(ATOMIC_BATCH_FUNC_NAME)
+
+-- Supported operation types.
+local SUPPORTED_OPERATIONS = {
+    get = true, insert = true, replace = true,
+    update = true, upsert = true, delete = true,
+}
+-- Operations that carry a tuple/object.
+local TUPLE_OPERATIONS = { insert = true, replace = true, upsert = true }
+-- Operations that use a primary-key lookup.
+local KEY_OPERATIONS = { get = true, update = true, delete = true }
+
+local STORAGE_OP_TO_STATS_OP = {
+    get = stats.op.GET,
+    insert = stats.op.INSERT,
+    replace = stats.op.REPLACE,
+    update = stats.op.UPDATE,
+    upsert = stats.op.UPSERT,
+    delete = stats.op.DELETE,
+}
+
+local function unref_buckets(unref_fn)
+    if unref_fn == nil then
+        return true
+    end
+
+    local ok, unref_ok, unref_err = pcall(unref_fn)
+    if not ok then
+        return nil, AtomicBatchError:new('Failed to unref buckets: %s', tostring(unref_ok))
+    end
+
+    if not unref_ok then
+        return nil, AtomicBatchError:new('Failed to unref buckets: %s', tostring(unref_err))
+    end
+
+    return true
+end
+
+local function observe_atomic_batch_stats(latency, operations, status)
+    if not stats.is_enabled() then
+        return
+    end
+
+    for _, op in ipairs(operations) do
+        if type(op) == 'table' and type(op.space) == 'string' then
+            local op_name = STORAGE_OP_TO_STATS_OP[op.type]
+            if op_name ~= nil then
+                stats.observe(latency, op.space, op_name, status)
+            end
+        end
+    end
+end
+
+local function collect_unique_spaces(operations)
+    local unique_spaces = {}
+    for _, op in ipairs(operations) do
+        if type(op) == 'table' and type(op.space) == 'string' then
+            unique_spaces[op.space] = true
+        end
+    end
+
+    return unique_spaces
+end
+
+local function observe_atomic_batch_stats_with_clock(started_at, operations, err)
+    local status = err == nil and 'ok' or 'error'
+    observe_atomic_batch_stats(clock.monotonic() - started_at, operations, status)
+end
+
+------------------------------------------------------------------------
+-- Storage-side implementation
+------------------------------------------------------------------------
+
+-- Execute a single CRUD operation inside an open box transaction.
+local function execute_single_op_on_storage(op, noreturn, space_fields)
+    local space = box.space[op.space]
+    if space == nil then
+        return nil, AtomicBatchError:new("Space %q doesn't exist", op.space)
+    end
+
+    local field_names = space_fields and space_fields[op.space] or nil
+    local wrap_opts = {
+        add_space_schema_hash = op._add_schema_hash or false,
+        field_names = field_names,
+        noreturn = noreturn,
+    }
+
+    if op.type == 'insert' then
+        return schema.wrap_func_result(space, space.insert, wrap_opts, space, op.tuple)
+    elseif op.type == 'replace' then
+        return schema.wrap_func_result(space, space.replace, wrap_opts, space, op.tuple)
+    elseif op.type == 'upsert' then
+        return schema.wrap_func_result(space, space.upsert, wrap_opts, space, op.tuple, op.operations)
+    elseif op.type == 'update' then
+        return schema.wrap_func_result(space, space.update, wrap_opts, space, op.key, op.operations)
+    elseif op.type == 'delete' then
+        return schema.wrap_func_result(space, space.delete, wrap_opts, space, op.key)
+    elseif op.type == 'get' then
+        return schema.wrap_func_result(space, space.get, wrap_opts, space, op.key)
+    end
+    return nil, AtomicBatchError:new("Unsupported operation type: %s", op.type)
+end
+
+-- Build bucket_id -> engine map; operations must already carry .bucket_id.
+local function get_bucket_ids_engine(operations)
+    local bucket_ids_engine = {}
+    for _, op in ipairs(operations) do
+        local space = box.space[op.space]
+        if space ~= nil and op.bucket_id ~= nil then
+            bucket_ids_engine[op.bucket_id] = space.engine
+        end
+    end
+    return bucket_ids_engine
+end
+
+-- Return error if mixed memtx/vinyl without MVCC, nil otherwise.
+local function check_mvcc_for_mixed_engines(operations)
+    local has_memtx = false
+    local has_vinyl = false
+    for _, op in ipairs(operations) do
+        local space = box.space[op.space]
+        if space ~= nil then
+            if space.engine == 'vinyl' then has_vinyl = true
+            else has_memtx = true end
+        end
+    end
+    if has_memtx and has_vinyl and not box.cfg.memtx_use_mvcc_engine then
+        return AtomicBatchError:new(
+            "atomic_batch over mixed memtx and vinyl spaces requires MVCC " ..
+            "(box.cfg.memtx_use_mvcc_engine = true)")
+    end
+    return nil
+end
+
+local function atomic_batch_on_storage(operations, opts)
+    dev_checks('table', {
+        noreturn = '?boolean',
+        fields = '?table',
+        sharding_key_hash = '?number',
+        sharding_func_hash = '?number',
+        skip_sharding_hash_check = '?boolean',
+    })
+
+    opts = opts or {}
+
+    -- Validate sharding hash for each unique space.
+    if opts.skip_sharding_hash_check ~= true then
+        local checked_spaces = {}
+        for _, op in ipairs(operations) do
+            if not checked_spaces[op.space] then
+                local _, err = sharding.check_sharding_hash(
+                    op.space, opts.sharding_func_hash, opts.sharding_key_hash,
+                    opts.skip_sharding_hash_check)
+                if err ~= nil then
+                    return nil, err
+                end
+                checked_spaces[op.space] = true
+            end
+        end
+    end
+
+    local err = check_mvcc_for_mixed_engines(operations)
+    if err ~= nil then
+        return { err = err }
+    end
+
+    local bucket_ids_engine = get_bucket_ids_engine(operations)
+    local ref_ok, ref_err, unref_fn = bucket_ref_unref.bucket_refrw_batch(bucket_ids_engine)
+    if not ref_ok then
+        return nil, ref_err
+    end
+
+    local results = {}
+    local execution_err, failed_index, failed_op
+
+    box.begin()
+
+    for i, op in ipairs(operations) do
+        local res, op_err = execute_single_op_on_storage(op, opts.noreturn, opts.fields)
+        if op_err ~= nil or (res ~= nil and res.err ~= nil) then
+            execution_err = op_err or AtomicBatchError:new('%s', res.err)
+            failed_index = i
+            failed_op = op
+            break
+        end
+        if opts.noreturn ~= true then
+            table.insert(results, res ~= nil and res.res or nil)
+        end
+    end
+
+    if execution_err ~= nil then
+        box.rollback()
+        local _, unref_err = unref_buckets(unref_fn)
+
+        local err = AtomicBatchError:new(
+            "Operation #%d (%s on %q) failed: %s",
+            failed_index, failed_op.type, failed_op.space, execution_err
+        )
+        err.operation_index = failed_index
+        err.operation_data = failed_op
+        err.unref_error = unref_err
+
+        return { err = err }
+    end
+
+    box.commit()
+
+    local unref_ok, unref_err = unref_buckets(unref_fn)
+    if not unref_ok then
+        return nil, unref_err
+    end
+
+    return { data = results }
+end
+
+atomic_batch.storage_api = {[ATOMIC_BATCH_FUNC_NAME] = atomic_batch_on_storage}
+
+------------------------------------------------------------------------
+-- Router-side helpers
+------------------------------------------------------------------------
+
+-- Validate a single operation descriptor, returns op or nil, err.
+local function validate_op(op, op_index)
+    if type(op) ~= 'table' then
+        return nil, AtomicBatchValidationError:new(
+            "Operation #%d must be a table, got %s", op_index, type(op))
+    end
+    if type(op.type) ~= 'string' or not SUPPORTED_OPERATIONS[op.type] then
+        return nil, AtomicBatchValidationError:new(
+            "Operation #%d has unsupported type %q " ..
+            "(allowed: get, insert, replace, update, upsert, delete)",
+            op_index, tostring(op.type))
+    end
+    if type(op.space) ~= 'string' then
+        return nil, AtomicBatchValidationError:new(
+            "Operation #%d: 'space' must be a string", op_index)
+    end
+    if TUPLE_OPERATIONS[op.type] then
+        if op.tuple == nil and op.object == nil then
+            return nil, AtomicBatchValidationError:new(
+                "Operation #%d (%s): 'tuple' or 'object' is required", op_index, op.type)
+        end
+        if op.type == 'upsert' and type(op.operations) ~= 'table' then
+            return nil, AtomicBatchValidationError:new(
+                "Operation #%d (upsert): 'operations' table is required", op_index)
+        end
+    end
+    if KEY_OPERATIONS[op.type] then
+        if op.key == nil then
+            return nil, AtomicBatchValidationError:new(
+                "Operation #%d (%s): 'key' is required", op_index, op.type)
+        end
+        if op.type == 'update' and type(op.operations) ~= 'table' then
+            return nil, AtomicBatchValidationError:new(
+                "Operation #%d (update): 'operations' table is required", op_index)
+        end
+    end
+    return op
+end
+
+-- Flatten an operation's .object field into a .tuple using the router schema.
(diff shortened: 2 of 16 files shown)


crud/atomic_batch.lua
```diff
@@ -0,0 +1,75 @@
...
+
+    opts = opts or {}
+
+    local vshard_router, err = utils.get_vshard_router_instance()
+    if err ~= nil then
+        return nil, AtomicBatchExecutionError:new(err)
+    end
+
+    local unique_spaces = router.collect_unique_spaces(operations)
+    local res, res_err = schema.wrap_func_reload(
```
a1div0, line 57: What's going on here?

crud/atomic_batch.lua
```diff
@@ -0,0 +1,75 @@
...
+        vshard_router,
+        sharding.wrap_method_for_spaces,
```
a1div0, line 59: What does this method do? It seems to manage updating the vshard cache? Can't vshard manage this itself?

a1div0 (changes requested):
Overall looks good. A couple of questions. And it needs a rebase.

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,420 @@
...
+        return nil
+    end
+
+    local metadata, metadata_err = build_result_metadata(
+        prepared.prepared_ops,
+        prepared.spaces_cache,
+        opts.fields
+    )
+    if metadata_err ~= nil then
+        return nil, AtomicBatchExecutionError:new("Failed to format result metadata: %s", metadata_err)
```
vakhov, line 408: I think we should validate `fields` before calling storage. Passing an unknown field, e.g. `fields = {customers = {'blablabla'}}`, returns an error, but the row has already been inserted and committed. The caller might assume nothing was written and retry the request. I suggest checking the fields before executing the batch.

crud/atomic_batch/storage.lua
```diff
@@ -0,0 +1,204 @@
...
+
+        return { err = err, op_latencies = op_latencies }
+    end
+
+    box.commit()
+
+    local unref_ok, unref_err = unref_buckets(unref_fn)
+    if not unref_ok then
+        return nil, unref_err
+    end
```
vakhov, line 197: `box.commit()` can throw, for [example if writing to WAL fails](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_txn_management/commit/). In that case we won't reach `unref_buckets()`, leaving the buckets referenced and preventing the rebalancer from moving them. I suggest ensuring the references are released on exceptions too.

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,420 @@
...
+    if sharding_data.sharding_func_hash ~= nil then
+        meta.sharding_func_hash = sharding_data.sharding_func_hash
+    end
+    if sharding_data.sharding_key_hash ~= nil then
+        meta.sharding_key_hash = sharding_data.sharding_key_hash
+    end
+    if sharding_data.skip_sharding_hash_check ~= nil then
+        meta.skip_sharding_hash_check = sharding_data.skip_sharding_hash_check
+    end
+end
```
vakhov, line 237: I'm not sure why we collect one pair of sharding hashes for the whole batch and then check it against every space. Different spaces can have different sharding keys and functions, right?
For example, using the test spaces, `customers` is sharded by `id`, while `customers_sharded_by_age` is sharded by `age`. Both sharding key values are `37`, so these operations target the same bucket
```lua
local res, err = crud.atomic_batch({
    {
        type = 'insert',
        space = 'customers',
        object = {id = 37, name = 'alice', age = 30},
    },
    {
        type = 'insert',
        space = 'customers_sharded_by_age',
        object = {id = 7101, name = 'bob', age = 37},
    },
})
```

I expected both rows to be inserted, but got `ShardingHashMismatchError` for `customer`s. It looks like the sharding key hash from `customers_sharded_by_age ends` up in the shared options and is then checked against customers too. Reloading the cache doesn't help because the keys really are different. I suggest keeping and passing these hashes separately for each space.

crud/atomic_batch/storage.lua
```diff
@@ -0,0 +1,204 @@
...
+
+        if op_err ~= nil or (res ~= nil and res.err ~= nil) then
+            execution_err = op_err or AtomicBatchExecutionError:new('%s', res.err)
+            failed_index = i
+            failed_op = op
+            break
+        end
+        if opts.noreturn ~= true then
+            table.insert(results, res ~= nil and res.res or nil)
+        end
```
vakhov, line 174: The result can be `nil` here, and `table.insert(results, nil)` doesn't preserve a slot for it. For example, if the first operation is a `get` for a missing row and the second is an `insert`, `data` contains just one element the inserted tuple at position `1`. The result indices no longer match the operation indices. The same happens with upsert. I suggest using `box.NULL` for operations without a result to preserve `their positions`.

vakhov (commented):
Thanks for the commit! I left a few comments.

crud/atomic_batch/router.lua
p0rtale: Same as other operations. They also raise the unknown-field error after the write via `utils.format_result`. Fixing only `atomic_batch` would be inconsistent, so I'd address it in a separate PR across all operations.

crud/atomic_batch/storage.lua
p0rtale: done

crud/atomic_batch/storage.lua
p0rtale: fixed

crud/atomic_batch/router.lua
vakhov: Sounds good. Could you please create an issue to track this for all operations, so it doesn't get lost?

crud/atomic_batch/router.lua
p0rtale: Fixed.

The sharding metadata is now kept per-space, so each operation is validated against its own space's sharding key/function hash.
The same bug also affected `skip_sharding_hash_check`. It was a single shared flag, and it's now per-space too.

crud/atomic_batch.lua
p0rtale: Here I'm wrapping `router.call_on_router` in two retry loops (the same pattern other operations use):
* `sharding.wrap_method_for_spaces` refreshes the sharding-metadata cache for the batch's spaces and retries when storage reports a sharding mismatch
* `schema.wrap_func_reload` reloads the schema on all replicasets and retries on a schema mismatch

crud/atomic_batch.lua
p0rtale: This isn't vshard's cache, it's crud's own.

vshard only maps `bucket_id -> replicaset` and has no idea about crud's sharding key/function. Those are declared in `_ddl_sharding_key`/`_ddl_sharding_func` and cached by crud on the router, so crud is the only one that can refresh them.

Same for the space schema: net.box caches it in `conn.space` and we have to reload schema explicitly after a ddl change.

vakhov (approved):

crud/atomic_batch/router.lua
p0rtale: https://github.com/tarantool/crud/issues/517

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,424 @@
...
+        fields = opts.fields,
+        sharding_meta = prepared.sharding_meta_by_space,
+    }
+
+    local storage_result, call_err = call.single(
+        vshard_router, prepared.bucket_id,
+        CRUD_ATOMIC_BATCH_FUNC_NAME,
+        {prepared.prepared_ops, on_storage_opts},
+        {mode = 'write', timeout = opts.timeout}
+    )
```
Satbek, line 381: A batch always goes to the master and takes rw bucket refs, even one made only of `get`s (`mode = 'write'` on the router, `bucket_refrw_batch` on the storage side). It shouldn't be rw-only when the ro path is available: CRUD has it for single ops (`bucket_refro`, and `call.single` supports `mode = 'read'`), and `get` is the only read-only operation, so an all-`get` batch could take it. That means new options on the method (as `get` has) with validation, plus ro ref sets in `bucket_ref_unref.lua` — only single `bucket_refro`/`bucket_unrefro` exist there now.

crud/atomic_batch/storage.lua
```diff
@@ -0,0 +1,211 @@
...
+    return nil, AtomicBatchExecutionError:new("Unsupported operation type: %s", op.type)
+end
+
+-- Build bucket_id -> engine map; operations must already carry .bucket_id.
+local function get_bucket_ids_engine(operations)
+    local bucket_ids_engine = {}
+    for _, op in ipairs(operations) do
+        local space = box.space[op.space]
+        if space ~= nil and op.bucket_id ~= nil then
+            bucket_ids_engine[op.bucket_id] = space.engine
```
Satbek, line 72: `get_bucket_ids_engine` maps `bucket_id → engine` with the last operation winning, and `_fast_bucket_refrw_batch` takes refs only for `vinyl` buckets. So in a batch with a vinyl op and a memtx op on the same `bucket_id`, no ref is taken if the memtx op comes later — while for single operations vinyl always goes the safe way (`make_fast_method`). Vinyl can yield inside a transaction, so a rebalance may start mid-transaction and duplicate data — exactly what the safe mode guards against. Minimal fix: let vinyl win in the map regardless of the operation order.

crud/atomic_batch.lua
```diff
@@ -0,0 +1,75 @@
...
+function atomic_batch.call(operations, opts)
+    checks('table', {
+        timeout = '?number',
+        noreturn = '?boolean',
+        fields = '?table',
+    })
+
+    opts = opts or {}
+
+    local vshard_router, err = utils.get_vshard_router_instance()
```
Satbek, line 51: `atomic_batch` always works with the default vshard router: it has no `opts.vshard_router`, unlike every other CRUD method except `readview`. So with a custom router or a vshard group (Cartridge) the method is unusable. Either add support for the option or state the limitation in README.

crud/atomic_batch/common.lua
```diff
@@ -0,0 +1,38 @@
+---- Shared constants and error classes for `crud.atomic_batch`.
+-- @module crud.atomic_batch.common
+--
+
+local errors = require('errors')
+
+local utils = require('crud.common.utils')
+
+local AtomicBatchExecutionError = errors.new_class('AtomicBatchExecutionError', {capture_stack = false})
+local AtomicBatchValidationError = errors.new_class('AtomicBatchValidationError', {capture_stack = false})
```
Satbek, line 10: Error classes are part of the public contract, yet the tests only check message text: no test asserts
  `err.class_name` for `AtomicBatchValidationError` or `AtomicBatchExecutionError`, while other CRUD tests pin
  the class (e.g. `custom_bucket_id_test.lua` asserts `err.class_name == 'InsertError'`). Asserting the class
  in at least one case per type would catch a rename or a merge of the classes.

crud/atomic_batch/common.lua
```diff
@@ -0,0 +1,38 @@
...
+local SUPPORTED_OPERATIONS = {
+    get = true, insert = true, replace = true,
+    update = true, upsert = true, delete = true,
+}
+-- Operations that carry a tuple/object.
+local TUPLE_OPERATIONS = { insert = true, replace = true, upsert = true }
+-- Operations that use a primary-key lookup.
+local KEY_OPERATIONS = { get = true, update = true, delete = true }
+
+return {
```
Satbek, line 29: Six of the eight exports in `common.lua` have exactly one consumer: `ATOMIC_BATCH_FUNC_NAME` and `CROSS_ENGINE_TXNS_SUPPORTED` are used only by `storage.lua`, `AtomicBatchValidationError`, `SUPPORTED_OPERATIONS`, `TUPLE_OPERATIONS` and `KEY_OPERATIONS` only by `router.lua`.

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,424 @@
...
+local function observe_atomic_batch_sub_op_stats(op_latencies, operations, status)
+    if not stats.is_enabled() then
+        return
+    end
+
+    for i, op in ipairs(operations) do
+        if type(op) == 'table' and type(op.space) == 'string' then
+            local op_name = STORAGE_OP_TO_STATS_OP[op.type]
+            if op_name ~= nil then
+                local latency = op_latencies ~= nil and op_latencies[i] or 0
```
Satbek, line 60: Sub-operations that never ran (the batch broke off on an earlier one) are still observed — with latency `0` — in the same `error` series as the failing one, so `latency_average` mixes "not executed" with "executed in 0". They shouldn't be counted at all.

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,424 @@
...
+        end
+    end
+    return op
+end
+
+-- Flatten an operation's .object field into a .tuple using the router schema.
+-- Returns a (possibly new) op table, or nil, err.
+-- The returned op has _add_schema_hash=true if flattening was done,
+-- which tells the storage side to include schema hash in the result
+-- so the router can detect a schema mismatch and retry.
```
Satbek, line 110: The comment is too big and partly states the obvious; other operations keep a one-liner above the flag  _add_schema_hash

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,424 @@
...
+    if known_replicasets == nil then
+        return nil, nil,
+            AtomicBatchExecutionError:new("Failed to get router replicasets: %s", tostring(routeall_err)),
+            const.NEED_SHARDING_RELOAD
+    end
+
+    local replicaset_id_by_obj = {}
+    for replicaset_id, replicaset in pairs(known_replicasets) do
+        replicaset_id_by_obj[replicaset] = replicaset_id
+    end
```
Satbek, line 198: The `replicaset_id_by_obj` map built from `routeall()` is redundant: `route(bucket_id)` returns the same replicaset object, and it already has `id`. Dropping the map greatly simplifies both functions: `resolve_replicaset_id` loses the map lookup and the `nil` branch, and `load_prepare_context` keeps only `get_spaces`.

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,424 @@
...
+        opts.fields
+    )
+    if metadata_err ~= nil then
+        return nil, AtomicBatchExecutionError:new("Failed to format result metadata: %s", metadata_err)
+    end
+
+    return {
+        metadata = metadata,
+        data = storage_result ~= nil and storage_result.data or nil,
+    }
```
Satbek, line 418: `data` items are positional and carry neither the operation type nor the space, so a client can't tell what `data[2]` is — a `replace` result or a `get` result — nor how to parse it, without going back to the original request (on errors this binding exists: `err.operation_index`, `err.operation_data`). Adding `ops[i] = {type = ..., space = ...}` (the same index as in the request) makes the response self-sufficient: the format comes from `metadata[ops[i].space]`, the type from `ops[i].type`.

The current shape also forces a client to keep the request and the response together — neither logically nor in code can the response be handled on its own. That doesn't fit go-tarantool, where every operation is a self-contained request type (`GetRequest`, `InsertRequest`, `ReplaceRequest`, ...) that carries its own result decoding, and a batch would have to be an exception:

```go
type AtomicBatchResult struct {
    Ops      []Operation              // type + space per item
    Metadata map[string][]FieldFormat
    Data     []interface{}
}
```

With `ops` a connector decodes each item on its own; without it, it has to keep the original request around, pair items by index and still can't decode an item into a typed row itself.

README.md
```diff
@@ -1118,6 +1119,67 @@ errs[4].err            -- 'Operation with tuple was rollback'
...
+```lua
+local result, err = crud.atomic_batch(operations, opts)
+```
+
+Runs an ordered list of CRUD operations (`get`, `insert`, `replace`, `update`,
+`upsert`, `delete`) in a single transaction on one replicaset.
+
+where:
+
+* `operations` (`table`) - ordered array of operation descriptors
```
Satbek, line 1133: `operations` is described as "ordered array of operation descriptors", but the descriptor options are never listed. Most importantly, `tuple` is not mentioned anywhere: the example uses `object` only, though `tuple` is supported for `insert`/`replace`/`upsert` and is used in tests more often than `object`. Add a table of the descriptor options per operation — with references to the options of the matching method (`crud.insert`, `crud.update`, .)

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,424 @@
...
+        -- When sharding index is the primary index, bucket_id may be part of the key.
+        sharding.fill_bucket_id_pk(space, op.key, bucket_id_data.bucket_id)
+
+        return bucket_id_data.bucket_id, {
+            bucket_id = bucket_id_data.bucket_id,
+            sharding_func_hash = bucket_id_data.sharding_func_hash,
+            sharding_key_hash = sharding_key_hash,
+            skip_sharding_hash_check = skip_sharding_hash_check,
+        }, nil
+    end
```
Satbek, line 177: The bucket_id derivation (`extract_from_pk` → `key_get_bucket_id` → `fill_bucket_id_pk`) is copied four times: `crud.get`, `crud.update`, `crud.delete` and the batch. Makes sense to add a helper for the key path, like `sharding.tuple_set_and_return_bucket_id` for tuples — e.g. `sharding.key_set_and_return_bucket_id`.

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,424 @@
...
+        local space = spaces_cache[op.space]
+        if space == nil then
+            return nil,
+                AtomicBatchExecutionError:new("Space %q doesn't exist", op.space),
+                const.NEED_SCHEMA_RELOAD
+        end
+
+        local flat_op, err = flatten_op_object(vshard_router, op)
+        if err ~= nil then
+            return nil, err, const.NEED_SCHEMA_RELOAD
```
Satbek, line 297: `NEED_SCHEMA_RELOAD` is redundant here: `flatten_op_object` calls `flatten_obj_reload`, which is `wrap_func_reload` around `flatten_obj` — the schema was already reloaded and the flatten retried before this error comes back. The flag only makes the outer wrapper run the whole batch a second time, while other methods return the flatten error without it (`insert.lua`, `replace.lua`, `upsert.lua`).

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,424 @@
...
+        return nil, nil,
+            AtomicBatchExecutionError:new("Failed to load spaces metadata: %s", spaces_err),
+            const.NEED_SCHEMA_RELOAD
+    end
+
+    local known_replicasets, routeall_err = vshard_router:routeall()
+    if known_replicasets == nil then
+        return nil, nil,
+            AtomicBatchExecutionError:new("Failed to get router replicasets: %s", tostring(routeall_err)),
+            const.NEED_SHARDING_RELOAD
```
Satbek, line 192: `NEED_SHARDING_RELOAD` is set in five places, but only the call-error branch checks the class (`sharding.result_needs_sharding_reload`), the way every other method does. Reloading the sharding cache fixes none of the other four (`routeall()` returned nothing, routing errors, `op_get_bucket_id`): `route()`/`routeall()` use vshard's own bucket→replicaset map, and `op_get_bucket_id` fails on the router, where the sharding hash check never runs. The flag only costs an extra full batch run and sharding-metadata requests with the same result.

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,424 @@
...
+            return nil, err
+        end
+
+        if single_bucket_id == nil then
+            single_bucket_id = bucket_id
+        end
+
+        merge_sharding_meta(sharding_meta_by_space, op.space, sharding_data)
+
+        local prepared_op = table.copy(flat_op)
```
Satbek, line 335: `table.copy(flat_op)` can be dropped: nothing reads the `bucket_id` written into the copy, and it doesn't protect the input anyway — the copy is shallow, while `sharding.tuple_set_and_return_bucket_id` writes bucket_id into the tuple itself.

crud/atomic_batch/common.lua
Satbek, line 10: Suggest a single `AtomicBatchError` instead of the `AtomicBatchValidationError` / `AtomicBatchExecutionError` pair, as every other method has one class named after the operation (`InsertError`, `ReplaceManyError`, ...). Nothing distinguishes the two from outside anyway — validation errors use the operation class elsewhere too, and the tests don't check the class at all.

crud/atomic_batch/storage.lua
```diff
@@ -0,0 +1,211 @@
...
+    if not ok then
+        return nil, AtomicBatchExecutionError:new('Failed to unref buckets: %s', tostring(unref_ok))
+    end
+
+    if not unref_ok then
+        return nil, AtomicBatchExecutionError:new('Failed to unref buckets: %s', tostring(unref_err))
+    end
+
+    return true
+end
```
Satbek, line 34: `pcall` and the `nil` check are unnecessary: `unref_fn` is always set, and `bucket_unrefrw` returns an error instead of throwing.

crud/atomic_batch/storage.lua
```diff
@@ -0,0 +1,211 @@
...
+
+        local err = AtomicBatchExecutionError:new(
+            "Operation #%d (%s on %q) failed: %s",
+            failed_index, failed_op.type, failed_op.space, execution_err
+        )
+        err.operation_index = failed_index
+        err.operation_data = failed_op
+        err.unref_error = unref_err
+
+        return { err = err, op_latencies = op_latencies }
```
Satbek, line 188: The `space_schema_hash` computed on storage for `object` operations never reaches the router — the result carries only `err` and `op_latencies`, so `schema.result_needs_reload` is never called in the batch. With an outdated router schema a batch returns an error to the client, while single `insert`/`replace`/`upsert` reload the schema and retry. Passing the hash back inside the error and checking it on the router, as the single operations do, would close the gap

crud/common/sharding/sharding_func.lua
```diff
@@ -119,7 +119,7 @@ function sharding_func_module.construct_as_callable_obj_cache(vshard_router, met
             local sharding_func, err = as_callable_object(metadata.sharding_func_def,
                                                           space_name)
             if err ~= nil then
-                if specified_space_name == space_name then
```
Satbek, line 122: Errors for different spaces are written into the same variable, so all but the last are lost (which one
  survives is decided by `pairs` order). It couldn't happen twice before this commit — a reload carried one
  space — but the batch passes a whole set now. Collecting errors per space instead would keep them all: both
  builders use the same class, and the message already contains the space name.

crud/common/sharding/sharding_key.lua
```diff
@@ -117,7 +117,7 @@ function sharding_key_module.construct_as_index_obj_cache(vshard_router, metadat
                                                                    metadata.space_format,
                                                                    metadata.sharding_key_def)
             if err ~= nil then
-                if specified_space_name == space_name then
+                if space_names[space_name] == true then
```
Satbek, line 120: Errors for different spaces are written into the same variable, so all but the last are lost (which one
  survives is decided by `pairs` order). It couldn't happen twice before this commit — a reload carried one
  space — but the batch passes a whole set now. Collecting errors per space instead would keep them all: both
  builders use the same class, and the message already contains the space name.

crud/common/sharding/sharding_metadata.lua
```diff
@@ -100,8 +100,8 @@ end
 -- cache.fetch_lock become unlocked during timeout passed to
 -- _fetch_on_router().
 -- metadata_map_name == nil means forced reload.
-local _fetch_on_router = locked(function(vshard_router, space_name, metadata_map_name, timeout)
-    dev_checks('table', 'string', '?string', 'number')
+local _fetch_on_router = locked(function(vshard_router, space_names, metadata_map_name, timeout)
```
Satbek, line 103: `space_names` rides through four functions (`wrap_method_for_spaces` → `reload_sharding_cache_for_spaces` → `_fetch_on_router` → both builders) only to pick a log level at the end, and where the error is returned it always holds one space. Suggest letting each space keep its own error:

```lua
-- sharding_func.lua / sharding_key.lua: a map instead of a single error
function sharding_func_module.construct_as_callable_obj_cache(vshard_router, metadata_map)
    local errs_by_space = {}
    ...
        if err ~= nil then
            errs_by_space[space_name] = err
        end
    ...
    return errs_by_space
end

-- sharding_metadata.lua: no space_names argument, errors are logged and cached
local _fetch_on_router = locked(function(vshard_router, metadata_map_name, timeout)
    local key_errs = sharding_key.construct_as_index_obj_cache(vshard_router, metadata_map)
    local func_errs = sharding_func.construct_as_callable_obj_cache(vshard_router, metadata_map)
    ...
end)

-- fetch_on_router(space_name) returns the error for its own space,
-- so whoever asks decides what the failure means
```

Then reload_sharding_cache(vshard_router) loses the argument, and the parameter leaves all four functions. This also fixes the overwritten errors in `construct_as_callable_obj_cache` and `construct_as_index_obj_cache`.

crud/common/sharding/bucket_ref_unref.lua
```diff
@@ -67,11 +67,33 @@ function bucket_ref_unref._bucket_unrefro(bucket_id)
     return vshard.storage.bucket_unrefro(bucket_id)
 end
 
---- Safe bucket_refrw_many that calls bucket_refrw for every bucket and aggregates errors
+--- Safe bucket_unrefrw_many that calls vshard.storage.bucket_unrefrw for every bucket.
```
Satbek, line 70: as I see bucket_ref_unref._bucket_unrefrw_many didn't change, can you revert this changes from diff?

Satbek (changes requested):
### One bucket, not one replicaset

The requirement is enforced as "same replicaset" (a `replicaset_id` comparison), yet a client can only rely on "same bucket": with different buckets, landing in one replicaset depends on the current topology and may break after a rebalance, and there is no way to see or influence it. Maybe the rule (and the error text) should say "one bucket", and the router should compare `bucket_id`.

### Commit is too big

2387 changed lines in 23 files, and only about a third is the operation itself. The sharding-metadata refactor (`specified_space_name` → `space_names`, `reload_sharding_cache_for_spaces`) changes signatures in shared code used by other methods, and the `BucketRefError` fix and the `vtest.lua` change are unrelated to the batch. Separate commits for the refactor and the fixes would be easier to review and to revert.

crud/atomic_batch/storage.lua
Satbek: > `box.commit()` can throw, for [example if writing to WAL fails](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_txn_management/commit/). In that case we won't reach `unref_buckets()`, leaving the buckets referenced and preventing the rebalancer from moving them. I suggest ensuring the references are released on exceptions too.

we should add this checks in other methods too

crud/atomic_batch/router.lua
Satbek: I think it is not quite critical and may be not implemented just yet. But in readme should be written that atomic_batch works only on masters.

crud/atomic_batch/router.lua
```diff
@@ -0,0 +1,374 @@
...
+    return flat_op
+end
+
+-- Compute bucket_id for a single router-side operation (after flattening).
+-- Returns { bucket_id = ..., sharding_data = ... } or nil, err, need_reload.
+local function op_get_bucket_id(vshard_router, op, space)
+    if TUPLE_OPERATIONS[op.type] then
+        -- insert / replace / upsert: bucket_id is embedded in the tuple
+        local sharding_data, err = sharding.tuple_set_and_return_bucket_id(
+            vshard_router, op.tuple, space, nil)
```
vakhov: Here we write `bucket_id` into the input tuple. If we then get `ShardingHashMismatchError`, we refresh the cache and retry, but the tuple already has a bucket filled in. We treat it as explicitly provided, so we neither recalculate it nor check the sharding hash.

I checked this by warming up the cache with sharding by `age`, then changing the key to `id` on storage. Inserting `{7101, box.NULL, 'test', 37}` succeeded, but the row ended up in bucket `254` instead of `1838`.

I suggest copying the original tuple on each attempt, like [crud.insert](https://github.com/tarantool/crud/blob/bafd117/crud/insert.lua#L96-L98) does: `local tuple = table.deepcopy(original_tuple)`. That way, the bucket gets recalculated using the updated cache.

PR #511, opened 2026-08-18
[TNTP-8909] fix: stop following the transfer destination on bucket ref error retries

### LINKED ISSUE #509: Violation of request consistency during rebalancing
Issue: https://jira.vk.team/browse/TNTP-8909

## What happens?

During bucket transfer, the source replicaset publishes the destination before the destination replicaset creates the corresponding `_bucket` record in the `RECEIVING` state. This makes the following situation possible:

  1. Bucket B and tuple T are located on storage S1. The router cache still maps B to S1.
  2. S1 changes the bucket state to `SENDING` with destination = S2.
  4. S2 has not yet processed the first bucket_recv() call. It has neither _bucket[B] nor tuple T.
  5. A crud request is sent to S1 using the cached route. 
  6. The S1 is already in safe mode. S1 rejects the request and reports S2 as the transfer destination. Crud retries the request directly on S2.
  7. Safe mode is still disabled on S2, because the replicaset has not yet received the bucket. In fast mode, crud does not validate bucket ownership. The operation is executed against an empty space on S2.
  8. Afterwards, bucket_recv() creates the bucket and copies tuple T from S1.

For delete, this produces a successful response with `rows=[]`. Once the transfer completes, the tuple still exists on S2, even though the delete was acknowledged without an error. The same race can make the final result of replace, update, or upsert differ from the result acknowledged to the client.

This is not caused by vshard 0.1.41 or by the introduction of the `READONLY` state. Version 0.1.41 only made the race easier to observe through changed bucket-reference handling. Older versions created a bucket ref during sending and commonly returned `BUCKET_IS_LOCKED` without a destination, now it's possible, that `TRANSFER_IS_IN_PROGRESS` with a destination is returned, allowing CRUD to retry immediately on the master without a bucket. Before 0.1.41 the same problem may happen, if storage e.g. restarts during rebalancing or failover (very common) happens during rebalancing, when replicaset has `SENDING` buckets, then crud makes request to these buckets and storage redirects it. 

## Reproducer

<details><summary>Luatest reproducer</summary>
<p>

```diff
diff --git a/test/integration/double_buckets_test.lua b/test/integration/double_buckets_test.lua
index b400b7e..488ee85 100644
--- a/test/integration/double_buckets_test.lua
+++ b/test/integration/double_buckets_test.lua
@@ -322,3 +322,94 @@ pgroup_not_applied.test_not_applied = function(g)
         not_applied_operations[g.params.operation].check_not_applied(not_applied_ids)
     end
 end
+
+local pgroup_transfer_gap = t.group('double_buckets_transfer_gap', {
+    {
+        backend = helpers.backend.CONFIG,
+        engine = 'memtx',
+        safe_mode = false,
+    },
+})
+
+pgroup_transfer_gap.before_all(function(g)
+    helpers.start_default_cluster(g, 'srv_simple_operations')
+    wait_balance(g, 1500, 1500)
+end)
+
+pgroup_transfer_gap.after_all(function(g)
+    helpers.stop_cluster(g.cluster, g.params.backend)
+end)
+
+--
+-- Once the source becomes SENDING, bucket_refrw() reports its destination and
+-- CRUD retries there directly. The destination must not accept that retry
+-- before bucket_recv() creates the bucket as RECEIVING.
+--
+pgroup_transfer_gap.test_delete_before_bucket_receive = function(g)
+    local source = g.cluster:server('s1-master')
+    local destination = g.cluster:server('s2-master')
+
+    -- Create the tuple.
+    local tuple_id = 8909
+    local bucket_id = source:exec(function(id)
+        local active = require('vshard.consts').BUCKET.ACTIVE
+        local bucket = box.space._bucket.index.status:min({active})
+        assert(bucket ~= nil)
+        box.space.customers:replace({id, bucket.id, 'A', 42})
+        return bucket.id
+    end, {tuple_id})
+
+    -- Populate the router cache with the source replicaset.
+    local get_result, get_err = g.router:call('crud.get', {
+        'customers', {tuple_id}, {bucket_id = bucket_id, mode = 'write'}})
+    t.assert_equals(get_err, nil)
+    t.assert_equals(#get_result.rows, 1)
+
+    -- Start sending the bucket.
+    local destination_id = destination:exec(function()
+        require('vshard').storage.internal.errinj.ERRINJ_RECEIVE_DELAY = true
+        return box.info.replicaset.name
+    end)
+    source:exec(function(bucket_id, destination_id)
+        local fiber = require('fiber')
+        local vshard = require('vshard')
+        rawset(_G, 'send_fiber', fiber.new(function()
+            return vshard.storage.bucket_send(bucket_id, destination_id)
+        end))
+        _G.send_fiber:set_joinable(true)
+        local sending = require('vshard.consts').BUCKET.SENDING
+        t.helpers.retrying({timeout = 60}, function()
+            t.assert_equals(box.space._bucket:get({bucket_id}).status, sending)
+        end)
+        -- Safe mode is enabled.
+        t.assert(_crud.rebalance_safe_mode_status())
+    end, {bucket_id, destination_id})
+    -- But destination doesn't have it yet.
+    destination:exec(function(bucket_id)
+        t.assert_not(box.space._bucket:get({bucket_id}))
+    end, {bucket_id})
+
+    --
+    -- And this one is the problem! The sender is in safe mode, it redirects
+    -- the router to the node, which is not in safe mode yet: it hasn't
+    -- received the bucket yet, but the request will be done anyway, even
+    -- without bucket.
+    --
+    local delete_result, delete_err = g.router:call('crud.delete', {
+        'customers', {tuple_id}, {bucket_id = bucket_id, timeout = 1},
+    })
+
+    destination:exec(function()
+        require('vshard').storage.internal.errinj.ERRINJ_RECEIVE_DELAY = false
+    end)
+    source:exec(function()
+        t.assert(_G.send_fiber:join())
+    end)
+    -- The tuple exists, even though, the delete succeeded.
+    destination:exec(function(tuple_id)
+        t.assert(box.space.customers:get({tuple_id}))
+    end, {tuple_id})
+
+    t.assert(delete_err ~= nil, ('delete succeeded with result %s, but ' ..
+        'tuple remained on destination'):format(json.encode(delete_result)))
+end
```

</p>
</details>

## What to do?

The only correct solution is using `router.call`, but here're the alternatives:

* redirected request always use the real bucket_refrw() check on the destination, even when global CRUD safe mode is disabled (bad, because the mode should be the property of a storage);
* router should ignore the redirects completely (do not make second request, just `bucket_reset`), redirects happen only during rebalancing (this one is probably better, more failing requests during rebalancing, but at least, we don't break data)

----

Another issue is that now vshard has `READONLY` bucket state, which is completely ignored by the Crud. Though, I don't see, how it violates the consistency right now

Satbek:
During rebalancing the source storage rejects a request with a bucket
ref error (e.g. TRANSFER_IS_IN_PROGRESS) and reports the transfer
destination. The router followed this redirect and retried the request
directly on the destination.

The destination, however, may not have received the bucket yet: it has
neither the _bucket record nor the data, and it is still in fast mode,
so no bucket ownership check is performed. The request was applied
against an empty space and acknowledged to the client. Once
bucket_recv() delivered the data, the cluster state diverged from the
acknowledged result: a deleted tuple reappeared, and the final result
of replace/update/upsert could differ from what was acknowledged.

The router no longer follows the destination reported by the error:
the bucket route cache is reset and the retry target is re-discovered
via vshard discovery, so a request is applied only on the replicaset
that actually holds the bucket.

A failed request is retried at most once and only after a recovery
action: the master is located on MISSING_MASTER, the cached master is
updated on NON_MASTER, the bucket route is reset on bucket ref errors
and the retry is re-routed by the bucket id. Other errors are returned
to the caller without a retry: the old code retried them blindly,
which silently doubled the effective request timeout. The retry is
given only the time left before the request deadline. Map requests are
not recovered at all: they are async, so storage errors arrive later,
in the future payload (see #513).

The rebalance safe mode also missed the start of a transfer on vshard
0.1.41+, where the bucket status is UPDATEd to READONLY instead of
being REPLACEd with SENDING, so safe mode was not enabled during
rebalancing and writes were not protected with bucket refs. Now the
READONLY update enables the safe mode as well.

Closes #509

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+* Fix consistency violation during rebalancing: a request could be retried on a
+  storage that did not own the bucket yet, so a read returned no data and a
+  write went to the wrong replicaset (#509).
+* Stop retrying requests blindly: a request is retried only if the router has
+  recovered from the error, and a retry never runs past the original timeout —
+  before, a request could take up to twice as long as requested.
+
 ## [1.7.5] - 15-07-26
 
 ### Added
--- a/crud/common/call.lua
+++ b/crud/common/call.lua
@@ -73,47 +73,145 @@ local function wrap_vshard_err(err, func_name, replicaset_id)
     ))
 end
 
---- Executes a vshard call and retries once after performing recovery actions
---- like bucket cache reset, destination redirect (for single calls), or master discovery.
-local function call_with_retry_and_recovery(vshard_router,
-    replicaset, method, func_name, func_args, call_opts, is_single_call)
+--- Executes a CRUD function on a vshard replicaset.
+local function call_on_replicaset(replicaset, method, func_name, func_args, call_opts)
     local func_args_ext = utils.append_array({ box.session.effective_user(), func_name }, func_args)
-
-    -- In case cluster was just bootstrapped with auto master discovery,
-    -- replicaset may miss master.
-    local resp, err = replicaset[method](replicaset, CRUD_CALL_FUNC_NAME, func_args_ext, call_opts)
-
-    if err == nil then
-        return resp, err
-    end
-
-    -- This is a partial copy of error handling from vshard.router.router_call_impl()
-    -- It is much simpler mostly because bucket_set() can't be accessed from outside vshard.
-    if err.class_name == bucket_ref_unref.BucketRefError.name then
-        if is_single_call and #err.bucket_ref_errs == 1 then
-            local single_err = err.bucket_ref_errs[1]
-            local destination = single_err.vshard_err.destination
-            if destination and vshard_router.replicasets[destination] then
-                replicaset = vshard_router.replicasets[destination]
-            end
-        end
-
-        for _, bucket_ref_err in pairs(err.bucket_ref_errs) do
-            local bucket_id = bucket_ref_err.bucket_id
-            local vshard_err = bucket_ref_err.vshard_err
-            if vshard_err.name == 'WRONG_BUCKET' or
-               vshard_err.name == 'BUCKET_IS_LOCKED' or
-               vshard_err.name == 'TRANSFER_IS_IN_PROGRESS' then
-                vshard_router:_bucket_reset(bucket_id)
-            end
-        end
-    elseif err.name == 'MISSING_MASTER' and replicaset.locate_master ~= nil then
-        replicaset:locate_master()
-    end
-
-    -- Retry only once: should be enough for initial discovery,
-    -- otherwise force user fix up cluster bootstrap.
     return replicaset[method](replicaset, CRUD_CALL_FUNC_NAME, func_args_ext, call_opts)
+end
+
+--- The bucket is not on the replicaset it is routed to anymore, so its
+--- route is stale and must be dropped from the router cache.
+local BUCKET_MOVED_ERRS = {
+    WRONG_BUCKET = true,
+    BUCKET_IS_LOCKED = true,
+    TRANSFER_IS_IN_PROGRESS = true,
+}
+
+--- Performs a recovery action for a call error:
+--- * MISSING_MASTER - the replicaset master is not discovered yet, so it
+---   is discovered explicitly;
+--- * NON_MASTER - the cached master is stale, but the bucket did not move,
+---   so the master is updated and the same replicaset can be retried;
+--- * WRONG_BUCKET, BUCKET_IS_LOCKED, TRANSFER_IS_IN_PROGRESS - the bucket
+---   route is stale, so the route cache is reset and the bucket is routed
+---   anew.
+---
+--- The routes of all the moved buckets are reset even when the request itself
+--- cannot be retried: the cache is stale regardless of the retry decision.
+---
+--- Returns:
+--- * replicaset - the replicaset to retry the request on;
+--- * nil - the request cannot be retried;
+--- * nil, err - the recovery itself failed, so the request cannot be retried.
+local function recover_from_err(vshard_router, replicaset, err)
+    local vshard_err = err
+
+    if err.class_name == bucket_ref_unref.BucketRefError.name then
+        for _, bucket_ref_err in ipairs(err.bucket_ref_errs) do
+            if BUCKET_MOVED_ERRS[bucket_ref_err.vshard_err.name] then
+                vshard_router:_bucket_reset(bucket_ref_err.bucket_id)
+            end
+        end
+
+        -- A request that failed on several buckets has no single replicaset
+        -- to be retried on, so only a single-bucket one is recovered further.
+        if #err.bucket_ref_errs ~= 1 then
+            return nil
+        end
+
+        vshard_err = err.bucket_ref_errs[1].vshard_err
+
+        if BUCKET_MOVED_ERRS[vshard_err.name] then
+            -- The stale route has been reset above, so the bucket is routed
+            -- anew.
+            local new_replicaset, route_err = vshard_router:route(err.bucket_ref_errs[1].bucket_id)
+            if route_err ~= nil then
+                return nil, CallError:new(
+                    "Failed to get router replicaset: %s, after an error: %s",
+                    tostring(route_err),
+                    tostring(err)
+                )
+            end
+            return new_replicaset
+        end
+    end
+
+    if vshard_err.name == 'MISSING_MASTER' then
+        replicaset:locate_master()
+        -- The master is not found, so the retry would get the same error.
+        if replicaset.master == nil then
+            return nil
+        end
+        return replicaset
+    elseif vshard_err.name == 'NON_MASTER' then
+        -- If the master update failed, the retry would get the same error.
+        if not replicaset:update_master(vshard_err.replica, vshard_err.master) then
+            return nil
+        end
+        return replicaset
+    end
+
+    return nil
+end
+
+local function call_single_with_recovery(vshard_router,
+    replicaset, method, func_name, func_args, call_opts)
+    local deadline = fiber_clock() + call_opts.timeout
+
+    local resp, err = call_on_replicaset(replicaset, method, func_name, func_args, call_opts)
+    if err == nil then
+        return resp, err, replicaset.id
+    end
+
+    local retry_replicaset, recover_err = recover_from_err(vshard_router, replicaset, err)
+    if retry_replicaset == nil then
+        return resp, recover_err or err, replicaset.id
+    end
+
+    local timeout = deadline - fiber_clock()
+    if timeout <= 0 then
+        return resp, err, replicaset.id
+    end
+
+    replicaset = retry_replicaset
+
+    call_opts.timeout = timeout
+    if call_opts.request_timeout ~= nil and call_opts.request_timeout > timeout then
+        call_opts.request_timeout = timeout
+    end
+
+    resp, err = call_on_replicaset(replicaset, method, func_name, func_args, call_opts)
+    return resp, err, replicaset.id
+end
+
+local function call_map_with_recovery(replicaset, method, func_name, func_args,
+    call_opts, deadline)
+    local future, err = call_on_replicaset(replicaset, method, func_name, func_args, call_opts)
+    if err == nil or err.name ~= 'MISSING_MASTER' then
+        return future, err
+    end
+
+    if fiber_clock() >= deadline then
+        return future, err
+    end
+
+    replicaset:locate_master()
+
+    -- The master is not found, so the retry would get the same error.
+    if replicaset.master == nil then
+        return future, err
+    end
+
+    local timeout = deadline - fiber_clock()
+    if timeout <= 0 then
+        return future, err
+    end
+
+    if call_opts.request_timeout ~= nil and call_opts.request_timeout > timeout then
+        call_opts.request_timeout = timeout
+    end
+
+    return call_on_replicaset(replicaset, method, func_name, func_args, call_opts)
 end
 
 function call.map(vshard_router, func_name, func_args, opts)
@@ -158,11 +256,13 @@ function call.map(vshard_router, func_name, func_args, opts)
         is_async = true,
         request_timeout = opts.mode == 'read' and opts.request_timeout or nil,
     }
+
+    local deadline = fiber_clock() + timeout
     while iter:has_next() do
         local args, replicaset, replicaset_id = iter:get()
 
-        local future, err = call_with_retry_and_recovery(vshard_router, replicaset, vshard_call_name,
-            func_name, args, call_opts, false)
+        local future, err = call_map_with_recovery(replicaset, vshard_call_name,
+            func_name, args, call_opts, deadline)
 
         if err ~= nil then
             local result_info = {
@@ -184,7 +284,6 @@ function call.map(vshard_router, func_name, func_args, opts)
         futures_by_replicasets[replicaset_id] = future
     end
 
-    local deadline = fiber_clock() + timeout
     for replicaset_id, future in pairs(futures_by_replicasets) do
         local wait_timeout = deadline - fiber_clock()
         if wait_timeout < 0 then
@@ -235,10 +334,11 @@ function call.single(vshard_router, bucket_id, func_name, func_args, opts)
     local timeout = opts.timeout or const.DEFAULT_VSHARD_CALL_TIMEOUT
     local request_timeout = opts.mode == 'read' and opts.request_timeout or nil
 
-    local res, err = call_with_retry_and_recovery(vshard_router, replicaset, vshard_call_name,
-        func_name, func_args, {timeout = timeout, request_timeout = request_timeout}, true)
+    local res, err, replicaset_id = call_single_with_recovery(vshard_router, replicaset, vshard_call_name,
+        func_name, func_args, {timeout = timeout, request_timeout = request_timeout})
+
     if err ~= nil then
-        return nil, wrap_vshard_err(err, func_name, replicaset.id)
+        return nil, wrap_vshard_err(err, func_name, replicaset_id)
     end
 
     if res == box.NULL then
@@ -265,7 +365,7 @@ function call.any(vshard_router, func_name, func_args, opts)
 
     local deadline = fiber_clock() + timeout
 
-    for replicaset_id, replicaset in pairs(replicasets) do
+    for _, replicaset in pairs(replicasets) do
         local wait_timeout = deadline - fiber_clock()
 
         local is_timeout = wait_timeout < 0
@@ -273,8 +373,8 @@ function call.any(vshard_router, func_name, func_args, opts)
             wait_timeout = 0
         end
 
-        local res, err = call_with_retry_and_recovery(vshard_router, replicaset, 'callro',
-            func_name, func_args, {timeout = wait_timeout}, false)
+        local res, err, replicaset_id = call_single_with_recovery(vshard_router, replicaset, 'callro',
+            func_name, func_args, {timeout = wait_timeout})
 
         if err == nil then
             if res == box.NULL then
--- a/crud/common/call.lua
+++ b/crud/common/call.lua
@@ -53,27 +53,16 @@ function call.get_vshard_call_name(mode, prefer_replica, balance)
     return 'callbre'
 end
 
-local function wrap_vshard_err(vshard_router, err, func_name, replicaset_id, bucket_id)
+local function wrap_vshard_err(err, func_name, replicaset_id)
     -- Do not rewrite ShardingHashMismatchError class.
     if err.class_name == sharding_utils.ShardingHashMismatchError.name then
         return errors.wrap(err)
     end
 
     if replicaset_id == nil then
-        local replicaset, _ = vshard_router:route(bucket_id)
-        if replicaset == nil then
-            return CallError:new(
-                "Function returned an error, but we couldn't figure out the replicaset: %s", err
-            )
-        end
-
-        replicaset_id = utils.get_replicaset_id(vshard_router, replicaset)
-
-        if replicaset_id == nil then
-            return CallError:new(
-                "Function returned an error, but we couldn't figure out the replicaset id: %s", err
-            )
-        end
+        return CallError:new(
+            "Function returned an error, but we couldn't figure out the replicaset id: %s", err
+        )
     end
 
     err = utils.update_storage_call_error_description(err, func_name, replicaset_id)
@@ -249,7 +238,7 @@ function call.single(vshard_router, bucket_id, func_name, func_args, opts)
     local res, err = call_with_retry_and_recovery(vshard_router, replicaset, vshard_call_name,
         func_name, func_args, {timeout = timeout, request_timeout = request_timeout}, true)
     if err ~= nil then
-        return nil, wrap_vshard_err(vshard_router, err, func_name, nil, bucket_id)
+        return nil, wrap_vshard_err(err, func_name, replicaset.id)
     end
 
     if res == box.NULL then
@@ -302,7 +291,7 @@ function call.any(vshard_router, func_name, func_args, opts)
         end
     end
 
-    return nil, wrap_vshard_err(vshard_router, last_err, func_name, last_replicaset_id)
+    return nil, wrap_vshard_err(last_err, func_name, last_replicaset_id)
 end
 
 return call
--- a/crud/common/map_call_cases/batch_postprocessor.lua
+++ b/crud/common/map_call_cases/batch_postprocessor.lua
@@ -55,7 +55,7 @@ function BatchPostprocessor:collect(result_info, err_info)
                 err_to_wrap = err.err
             end
 
-            local err_obj = err_info.err_wrapper(self.vshard_router, err_to_wrap, unpack(err_info.wrapper_args))
+            local err_obj = err_info.err_wrapper(err_to_wrap, unpack(err_info.wrapper_args))
             err_obj.operation_data = err.operation_data
             err_obj.space_schema_hash = err.space_schema_hash
 
--- a/crud/common/rebalance.lua
+++ b/crud/common/rebalance.lua
@@ -25,10 +25,13 @@ local function safe_mode_bucket_trigger(_, new, space, op)
     if space ~= '_bucket' then
         return
     end
-    -- We are interested only in two operations that indicate the beginning of bucket migration:
-    --  * We are receiving a bucket (new bucket with status RECEIVING)
-    --  * We are sending a bucket to another node (existing bucket status changes to SENDING)
+    -- Rebalancing start markers:
+    --  * INSERT with RECEIVING - the bucket is being received.
+    --  * UPDATE to READONLY - since vshard 0.1.41 a transfer starts in
+    --    READONLY.
+    --  * REPLACE to SENDING - compatibility with older vshard versions.
     if (op == 'INSERT' and new.status == vshard_consts.BUCKET.RECEIVING) or
+       (op == 'UPDATE' and new.status == vshard_consts.BUCKET.READONLY) or
        (op == 'REPLACE' and new.status == vshard_consts.BUCKET.SENDING) then
         local stored_safe_mode = schema.settings_space:get{ SAFE_MODE_STATUS }
         if not stored_safe_mode or not stored_safe_mode.value then
--- a/crud/common/sharding/init.lua
+++ b/crud/common/sharding/init.lua
@@ -18,7 +18,8 @@ function sharding.get_replicasets_by_bucket_id(vshard_router, bucket_id)
         return nil, GetReplicasetsError:new("Failed to get replicaset for bucket_id %s: %s", bucket_id, err.err)
     end
 
-    local replicaset_id = utils.get_replicaset_id(vshard_router, replicaset)
+    local replicaset_id = replicaset.id
+
     if replicaset_id == nil then
         return nil, GetReplicasetsError:new("Failed to get replicaset id for bucket_id %s replicaset", bucket_id)
     end
@@ -314,7 +315,7 @@ function sharding.split_tuples_by_replicaset(vshard_router, tuples, space, opts)
                     sharding_data.bucket_id, err.err)
         end
 
-        local replicaset_id = utils.get_replicaset_id(vshard_router, replicaset)
+        local replicaset_id = replicaset.id
         if replicaset_id == nil then
             return nil, GetReplicasetsError:new(
                     "Failed to get replicaset id for bucket_id %s replicaset",
--- a/test/entrypoint/srv_say_hi/all.lua
+++ b/test/entrypoint/srv_say_hi/all.lua
@@ -39,18 +39,63 @@ return {
             local _, replicaset = next(replicasets)
             local replicaset_mt = getmetatable(replicaset)
 
-            for _, vshard_call_name in ipairs(vshard_call_names) do
-                local old_func = replicaset_mt.__index[vshard_call_name]
-                assert(old_func ~= nil, vshard_call_name)
+            if rawget(_G, 'vshard_call_originals') == nil then
+                rawset(_G, 'vshard_call_originals', {})
+            end
+            local originals = rawget(_G, 'vshard_call_originals')
 
-                replicaset_mt.__index[vshard_call_name] = function(...)
-                    local func_name = select(2, ...)
-                    if not string.startswith(func_name, 'vshard.') or func_name == 'vshard.storage.call' then
-                        table.insert(_G.vshard_calls, vshard_call_name)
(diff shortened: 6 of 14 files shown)


a1div0 (approved):

vakhov (approved):

crud/common/call.lua
```diff
@@ -85,9 +85,9 @@ local function wrap_vshard_err(vshard_router, err, func_name, replicaset_id, buc
 end
```
Serpentian, line 74: Since vshard 0.1.41 a transfer no longer starts at `SENDING`, it's `READONLY` now, which crud completely ignores. I'd fix it, while we're here, the safe mode should be enabled also on `UPDATE` to `READONLY`.

We should not drop the `SENDING` for compatibility with old vshard versions, IMHO.

----

P.S. VShard itself uses that status to wait for all rw refs to end and in order to check, that on replicas rw refs are also ended. In the future it's possible, that vshard will use that status in order to make arbitrary bucket `READONLY`, and crud should respect that.

test/integration/double_buckets_test.lua
```diff
@@ -420,3 +420,69 @@ pgroup_transfer_gap.test_delete_before_bucket_receive = function(g)
     t.assert_str_contains(delete_err.err, "TRANSFER_IS_IN_PROGRESS")
 end
 
+--
+-- Safe mode is enabled by an on_replace trigger on _bucket. The
+-- SENDING/RECEIVING rows replicate and fire the trigger on replicas too,
+-- so safe mode must survive a failover: a promoted replica stays in safe
+-- mode.
+--
+local pgroup_safe_mode_failover = t.group('safe_mode_failover', {
```
Serpentian: Why the whole new cluster for the single case? Can't it reuse the `pgroup_transfer_gap` (maybe just rename it to smth more general)

crud/common/call.lua
```diff
@@ -85,9 +85,9 @@ local function wrap_vshard_err(vshard_router, err, func_name, replicaset_id, buc
...
 
 --- Executes a vshard call and retries once after performing recovery actions
---- like bucket cache reset, destination redirect (for single calls), or master discovery.
+--- like bucket cache reset or master discovery.
 local function call_with_retry_and_recovery(vshard_router,
```
Serpentian: Why do we need the `retry` now, when no "recovery actions" (as in the comment above, which is btw not correct) are performed. The search for the leader is anyway done inside the `rs:call*` before the call itself, so one more search for it won't help anyway.

So, maybe we can just completely get rid of retry?

crud/common/call.lua
```diff
@@ -118,7 +110,15 @@ local function call_with_retry_and_recovery(vshard_router,
                 vshard_router:_bucket_reset(bucket_id)
             end
         end
-    elseif err.name == 'MISSING_MASTER' and replicaset.locate_master ~= nil then
+
+        -- No retry here: a redirect may point to a replicaset that hasn't
+        -- received the bucket yet and would apply the request against an
+        -- empty space in fast mode (gh-509). The next request re-discovers
+        -- the bucket location via the router.
+        return resp, err
```
Serpentian: And if we do want to retry in that function, we can do that, but it should be done with `bucket_reset` + `new_rs = router:route(bid)` + `new_rs:call`.

And why do we start exiting on all of the ref errors: e.g. there `NON_MASTER` error, we can safely retry it even without bucket reset. If there's destination in the , then for retry `reset + route`.

So, it seems to me, we should either retry or not in that function, now it looks messy to me: the retry is present, but it's meaningless, or am I missing smth?

crud/common/call.lua
```diff
@@ -247,7 +247,7 @@ function call.single(vshard_router, bucket_id, func_name, func_args, opts)
     local request_timeout = opts.mode == 'read' and opts.request_timeout or nil
 
     local res, err = call_with_retry_and_recovery(vshard_router, replicaset, vshard_call_name,
-        func_name, func_args, {timeout = timeout, request_timeout = request_timeout}, true)
+        func_name, func_args, {timeout = timeout, request_timeout = request_timeout})
     if err ~= nil then
         return nil, wrap_vshard_err(vshard_router, err, func_name, nil, bucket_id)
```
Serpentian: This `wrap_vshard_err` function internally does `router:route()` in order to create error object, but the `call_with_retry_and_discovery` uses the `bucket_reset`. So the `wrap_vshard_err` will mention `replicaset_id`of another replicaset and not the one, we did request to

crud/common/call.lua
```diff
@@ -101,14 +101,6 @@ local function call_with_retry_and_recovery(vshard_router,
...
     if err.class_name == bucket_ref_unref.BucketRefError.name then
-        if is_single_call and #err.bucket_ref_errs == 1 then
-            local single_err = err.bucket_ref_errs[1]
-            local destination = single_err.vshard_err.destination
-            if destination and vshard_router.replicasets[destination] then
-                replicaset = vshard_router.replicasets[destination]
-            end
-        end
-
         for _, bucket_ref_err in pairs(err.bucket_ref_errs) do
```
Serpentian: The `call.map` also calls the `call_with_retry_and_recovery`, but it's done with `is_async` flag, which means, the futures are returned. This loop doesn't work, the bucket is never reset for the calls done via `call.map`. Are we ok with that?

test/integration/double_buckets_test.lua
```diff
@@ -322,3 +322,101 @@ pgroup_not_applied.test_not_applied = function(g)
...
+    wait_balance(g, 1500, 1500)
+end)
+
+pgroup_transfer_gap.after_all(function(g)
+    helpers.stop_cluster(g.cluster, g.params.backend)
+end)
+
+--
+-- Once the source becomes SENDING, bucket_refrw() reports its destination and
+-- CRUD retries there directly. The destination must not accept that retry
```
Serpentian: Nit: we should not describe the bug in the present tense

test/integration/double_buckets_test.lua
```diff
@@ -322,3 +322,101 @@ pgroup_not_applied.test_not_applied = function(g)
         not_applied_operations[g.params.operation].check_not_applied(not_applied_ids)
     end
 end
+
+local pgroup_transfer_gap = t.group('double_buckets_transfer_gap', {
+    {
+        backend = helpers.backend.CONFIG,
```
Serpentian: It seams, we cannot hardcode the `config` backend in the test group, the same comment for other commit. The other groups in this file get their backends from `helpers.backend_matrix()`, which adds `helpers.backend.CONFIG` only `if helpers.is_tarantool3_crud_roles_supported()`

test/integration/double_buckets_test.lua
```diff
@@ -322,3 +322,101 @@ pgroup_not_applied.test_not_applied = function(g)
...
+    local delete_result, delete_err = g.router:call('crud.delete', {
+        'customers', {tuple_id}, {bucket_id = bucket_id, timeout = 1},
+    })
+
+    destination:exec(function()
+        require('vshard').storage.internal.errinj.ERRINJ_RECEIVE_DELAY = false
+    end)
+    source:exec(function()
+        local t = require('luatest')
+        t.assert(_G.send_fiber:join())
```
Serpentian: Nit: this checks, that the fiber was joined successfully, not that the send succeeded

test/integration/double_buckets_test.lua
```diff
@@ -420,3 +420,69 @@ pgroup_transfer_gap.test_delete_before_bucket_receive = function(g)
...
+
+pgroup_safe_mode_failover.before_all(function(g)
+    helpers.start_default_cluster(g, 'srv_simple_operations')
+end)
+
+pgroup_safe_mode_failover.after_all(function(g)
+    helpers.stop_cluster(g.cluster, g.params.backend)
+end)
+
+pgroup_safe_mode_failover.test_safe_mode_survives_failover = function(g)
```
Serpentian: Why the test (and the group) are named as `failover`, when master is not changed in the test

test/integration/double_buckets_test.lua
```diff
@@ -420,3 +420,69 @@ pgroup_transfer_gap.test_delete_before_bucket_receive = function(g)
...
+    local source = g.cluster:server('s1-master')
+    local source_replica = g.cluster:server('s1-replica')
+    local destination = g.cluster:server('s2-master')
+
+    -- -- bucket_send requires both masters to be synced with their replicas
```
Serpentian: Double comment marker, and below too

test/integration/double_buckets_test.lua
```diff
@@ -322,3 +322,101 @@ pgroup_not_applied.test_not_applied = function(g)
...
+pgroup_transfer_gap.after_all(function(g)
+    helpers.stop_cluster(g.cluster, g.params.backend)
+end)
+
+--
+-- Once the source becomes SENDING, bucket_refrw() reports its destination and
+-- CRUD retries there directly. The destination must not accept that retry
+-- before bucket_recv() creates the bucket as RECEIVING.
+--
+pgroup_transfer_gap.test_delete_before_bucket_receive = function(g)
```
Serpentian: If it's not difficult, can we also reproduce the issue with faliover and instance restart to be sure, that the issue is resolved. And add these cases to tests. Up to u

Serpentian (commented):
Great work!

crud/common/call.lua
Satbek: thanks for information about READONLY status

added additional condition (update + READONLY) in trigger

crud/common/call.lua
Satbek: yes, it's ok

crud/common/call.lua
Satbek: yes, I agree. I got rid of retry, only bucket reset happens

test/integration/double_buckets_test.lua
Satbek: added matrix

crud/common/call.lua
Satbek: thanks for noticing. In fact there were no need to provide vshard_router and bucket_id in wrap_vshard_err.

I simplified wrap_vshard_err signature

crud/common/call.lua
Satbek: I think it's pretty fixable but in another pr

test/integration/double_buckets_test.lua
Satbek: added checks that safe mode survives restart and enabled on replicas

crud/common/call.lua
Serpentian: I've actually missed, that this is not the case for the async requests. Vshard doesn't search or wait for masters, when request is async. And crud uses async requests in the `call,map` (`insert/replace/upsert_many`).

So, it'll be the regression: before the commit map requests were able to tolerate the missing master, now they don't.

crud/common/call.lua
Serpentian: https://github.com/tarantool/crud/pull/511#discussion_r3854287384

crud/common/call.lua
Serpentian: This should be in the separate commti

crud/common/call.lua
```diff
@@ -243,13 +214,14 @@ function call.single(vshard_router, bucket_id, func_name, func_args, opts)
         return nil, CallError:new("Failed to get router replicaset: %s", tostring(err))
     end
 
+    local replicaset_id = utils.get_replicaset_id(vshard_router, replicaset)
```
Serpentian: `utils.get_replicaset_id()` walks every known replicaset comparing it by identity. It is needed only inside the `if err ~= nil` branch two lines below, but it is computed unconditionally, so every `get`, `insert`, `update`, `delete` - the whole hot path of single calls - now pays for a scan over all replicasets.

I'd consider moving it to thhe error branch below. And btw, why you don't use the replicaset.id?)

crud/common/map_call_cases/batch_postprocessor.lua
```diff
@@ -55,7 +55,7 @@ function BatchPostprocessor:collect(result_info, err_info)
                 err_to_wrap = err.err
             end
 
-            local err_obj = err_info.err_wrapper(self.vshard_router, err_to_wrap, unpack(err_info.wrapper_args))
```
Serpentian, line 58: `self.vshard_router` is not used anymore

crud/common/call.lua
```diff
@@ -53,29 +53,12 @@ function call.get_vshard_call_name(mode, prefer_replica, balance)
...
-        local replicaset, _ = vshard_router:route(bucket_id)
-        if replicaset == nil then
-            return CallError:new(
-                "Function returned an error, but we couldn't figure out the replicaset: %s", err
-            )
-        end
-
-        replicaset_id = utils.get_replicaset_id(vshard_router, replicaset)
-
-        if replicaset_id == nil then
```
Serpentian, line 72: `replicaset_id` can still be `nil`, and the `format_replicaset_error` will raise luajit in such case, regression

crud/common/call.lua
Satbek: ok, I will fix commits after review

crud/common/call.lua
Satbek: ok, I will return missingMaster retry

crud/common/call.lua
Satbek: > And btw, why you don't use the replicaset.id?)

legacy reasons from 2.x -> 3.x migration.

I suppose replicaset.id is name, when identification_mode == 'name_as_key' and uuid when identification_mode == 'uuid_as_key'?

Serpentian: > I suppose replicaset.id is name, when identification_mode == 'name_as_key' and uuid when identification_mode == 'uuid_as_key'?

Yep, exactly. And `replicaset.id` is the same value you get in the `get_replicaset_id` (but it might be missing on old vshard versions, which seems is not the case for crud: it works with specific vshard)

crud/common/call.lua
Serpentian: Sure, it was there even before this patchset. Created the issue: https://github.com/tarantool/crud/issues/513

crud/common/call.lua
Serpentian: Let's then retry on the ref error too, as described in the https://github.com/tarantool/crud/pull/511#discussion_r3820714882. The thing, which breaks request consistency for us: using `destination` from vshard as

test/integration/double_buckets_test.lua
```diff
@@ -322,3 +334,196 @@ pgroup_not_applied.test_not_applied = function(g)
...
+        rawset(_G, 'send_fiber', fiber.new(function()
+            return vshard.storage.bucket_send(bucket_id, destination_id)
+        end))
+        _G.send_fiber:set_joinable(true)
+        t.helpers.retrying({timeout = 15}, function()
+            t.assert_equals(box.space._bucket:get({bucket_id}).status,
+                expected_status)
+        end)
+        -- Safe mode is enabled.
+        t.assert(rawget(_G, '_crud').rebalance_safe_mode_status())
```
Serpentian, line 428: Let's return the cluster to the same it was before the test start: let's reset the mode back to fast after every test

Serpentian (commented):
Almost there! We're moving fast

Please, let's properly split the changes between the commits. Otherwise, I cannot judge, whether the code is properly splitted and I am forced to look at all the changes in their entirety, this way it's way easier to skip the bug.

And if you don't mind, can you in the future please write your responses to each comment before re-requesting their review? This helps a lot to see if you have addressed the comments. Even a simple "Fixed" is better than nothing. Thanks in advance!

crud/common/map_call_cases/batch_postprocessor.lua
Satbek: fixed

crud/common/call.lua
Satbek: fixed, got rid of `get_repliacset_id` helper

crud/common/call.lua
Satbek: fixed, add check for   `replicaset_id == nil` in wrap_vshard_err

test/integration/double_buckets_test.lua
Satbek: fixed

crud/common/call.lua
Satbek: done, added also tests for rety, recovery logic

crud/common/call.lua
Satbek: fixed

crud/common/sharding/init.lua
```diff
@@ -314,14 +313,13 @@ function sharding.split_tuples_by_replicaset(vshard_router, tuples, space, opts)
                     sharding_data.bucket_id, err.err)
         end
 
-        local replicaset_id = utils.get_replicaset_id(vshard_router, replicaset)
```
Serpentian, line 317: Nit: let's do the

```
local replicaset_id = replicaset.id
```

To minimize the diff and in order to avoid table access on every line

crud/common/call.lua
```diff
@@ -235,8 +281,8 @@ function call.single(vshard_router, bucket_id, func_name, func_args, opts)
     local timeout = opts.timeout or const.DEFAULT_VSHARD_CALL_TIMEOUT
     local request_timeout = opts.mode == 'read' and opts.request_timeout or nil
 
-    local res, err = call_with_retry_and_recovery(vshard_router, replicaset, vshard_call_name,
-        func_name, func_args, {timeout = timeout, request_timeout = request_timeout}, true)
+    local res, err = call_single_with_recovery(vshard_router, replicaset, vshard_call_name,
+        func_name, func_args, {timeout = timeout, request_timeout = request_timeout})
     if err ~= nil then
         return nil, wrap_vshard_err(err, func_name, replicaset.id)
```
Serpentian: When `recover_from_err()` returns a bucket to re-route, `call_single_with_recovery()` reassigns its local `replicaset` and retries there. If that retry fails, `call.single()` wraps the error with `replicaset.id` of the *original* replicaset - the one the request did not go to on the failing attempt.

This is in the third commit

test/entrypoint/srv_say_hi/all.lua
```diff
@@ -39,20 +39,64 @@ return {
...
+
+        rawset(_G, 'bucket_ref_rw', function(bucket_id)
+            local bucket_ref_unref = require('crud.common.sharding.bucket_ref_unref')
+
+            local ref_ok, err, unref = bucket_ref_unref.bucket_refrw(bucket_id)
+            if not ref_ok then
+                return nil, err
+            end
+
+            return unref(bucket_id)
```
Serpentian: Nit: with the fast mode this will return the function over IPROTO, which is prohibited (and it will fail with hard to parse error). it seems we should return only the first two return values from here

CHANGELOG.md
```diff
@@ -12,6 +12,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+  route cache is reset and the retry target is re-discovered via vshard
+  discovery (#509).
+* Retry requests only after a recovery action: master discovery on
+  MISSING_MASTER, master update on NON_MASTER and route re-discovery on
+  other bucket ref errors. Other errors are returned to the caller without
+  a blind retry, which could silently double the effective request timeout.
+  The retries do not extend the timeout: a retry is given only the time
+  left before the deadline. Map requests are not recovered at all:
+  they are async, so storage errors arrive in the future payload
+  (#511, #513).
```
Serpentian: 511 is the PR itself

CHANGELOG.md
```diff
@@ -12,6 +12,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   with vshard 0.1.41+ a bucket transfer starts in READONLY instead of
   SENDING, so the safe mode was not enabled during rebalancing on such
   versions (#509).
+* Stop following the destination from bucket ref errors on retry. During
```
Serpentian: I as developer cannot parse these changelogs, they're to technical and says nothing to a user. Let's say, what was fixed from the user point of view: that we've fixed the request consistency violation

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
```
Serpentian, line 10: I see no reason for that changelog here, nothing was broken because of that

test/integration/double_buckets_test.lua
```diff
@@ -322,3 +334,198 @@ pgroup_not_applied.test_not_applied = function(g)
...
+    end)
+    local destination_id = get_replicaset_id(destination)
+    -- Since vshard 0.1.41 a transfer starts in READONLY instead of SENDING,
+    -- but the SENDING is kept for compatibility with older versions.
+    local consts = require('vshard.consts')
+    local expected_status
+    if helpers.is_bucket_transfer_starts_readonly() then
+        expected_status = consts.BUCKET.READONLY
+    else
+        expected_status = consts.BUCKET.SENDING
```
Serpentian: `ERRINJ_RECEIVE_DELAY` blocks inside `bucket_recv()`, which the source calls only *after* the REPLACE to SENDING - so the errinj does not pin the bucket in READONLY. The READONLY window is very small and we should alway use the `SENDING` state in order to avoid flakiness.

P.S. And it'll also allow us to drop the helper from the second commit

crud/common/call.lua
```diff
@@ -73,47 +73,90 @@ local function wrap_vshard_err(err, func_name, replicaset_id)
...
+--- * MISSING_MASTER - the replicaset master is not discovered yet, so it
+---   is discovered explicitly;
+--- * NON_MASTER - the cached master is stale, but the bucket did not move,
+---   so the master is updated and the same replicaset can be retried;
+--- * WRONG_BUCKET, BUCKET_IS_LOCKED, TRANSFER_IS_IN_PROGRESS - the bucket
+---   route is stale, so the route cache is reset.
+---
+--- Returns:
+--- * true - the request can be retried on the same replicaset;
+--- * true, bucket_id - the request can be retried on the replicaset the
```
Serpentian: Will returning the `replicaset` or `nil, err` be easier as API here? I find it very strange, that if bucket_id is returned, we should find the replicaset.

Up to u, feel free to resolve without fix

crud/common/call.lua
```diff
@@ -73,47 +73,90 @@ local function wrap_vshard_err(err, func_name, replicaset_id)
...
+    if vshard_err.name == 'MISSING_MASTER' then
+        replicaset:locate_master()
+        return true
+    elseif vshard_err.name == 'NON_MASTER' then
+        -- If the master update failed, the retry would get the same error.
+        return replicaset:update_master(vshard_err.replica, vshard_err.master)
+    elseif vshard_err.name == 'WRONG_BUCKET' or
+           vshard_err.name == 'BUCKET_IS_LOCKED' or
+           vshard_err.name == 'TRANSFER_IS_IN_PROGRESS' then
+        for _, bucket_ref_err in ipairs(err.bucket_ref_errs) do
```
Serpentian: The WRONG_BUCKET/BUCKET_IS_LOCKED/TRANSFER_IS_IN_PROGRESS branch iterates `err.bucket_ref_errs`, but the branch is entered by `vshard_err.name` alone, which is the first error in that list.

So, the question is, shouldn't we just iterate over the list of errors and apply all these checks to all of them? Can they differ?

P.S. But it seems to me that in the current code this list always has the length of 1, the `_bucket_refrw_many` stops on first error, and `call.map` are not recovered at all, as we remember

Serpentian (commented):
Looks like the last iteration to me. Thank you for putting up with my whining :)

test/entrypoint/srv_say_hi/all.lua
Satbek: fixed

crud/common/call.lua
Satbek: fixed

crud/common/sharding/init.lua
Satbek: fixed

test/integration/double_buckets_test.lua
Satbek: fixed

crud/common/call.lua
Satbek: Yes, they have always the length of 1. Created an issue to get rid of `bucket_ref_errs` table https://github.com/tarantool/crud/issues/514

CHANGELOG.md
Satbek: FIxed

CHANGELOG.md
Satbek: fixed

CHANGELOG.md
Satbek: fixed

crud/common/call.lua
Satbek: fixed

crud/common/call.lua
Satbek: Anyway I added cycle over table to handle them properly.

Satbek:
I also returned back recovery for call.map because we actually can handle MISSING_MASTER error in it.

crud/common/rebalance.lua
```diff
@@ -25,10 +25,13 @@ local function safe_mode_bucket_trigger(_, new, space, op)
     if space ~= '_bucket' then
         return
     end
-    -- We are interested only in two operations that indicate the beginning of bucket migration:
-    --  * We are receiving a bucket (new bucket with status RECEIVING)
-    --  * We are sending a bucket to another node (existing bucket status changes to SENDING)
+    -- Rebalancing start markers:
```
Serpentian, line 28: Nit: the commit msg in the second one ends with `and` and that's it

test/helper.lua
```diff
@@ -922,6 +922,32 @@ function helpers.start_tarantool3_cluster(g, cfg)
     g.cluster:start()
 end
 
+function helpers.set_safe_mode(cluster, is_enabled)
```
Serpentian, line 925: Nit: It seems, this change belongs to the third commit, it has nothing to do with the `READONLY` state commit

crud/common/call.lua
```diff
@@ -235,10 +332,16 @@ function call.single(vshard_router, bucket_id, func_name, func_args, opts)
...
     local request_timeout = opts.mode == 'read' and opts.request_timeout or nil
 
-    local res, err = call_with_retry_and_recovery(vshard_router, replicaset, vshard_call_name,
-        func_name, func_args, {timeout = timeout, request_timeout = request_timeout}, true)
+    local res, err = call_single_with_recovery(vshard_router, replicaset, vshard_call_name,
+        func_name, func_args, {timeout = timeout, request_timeout = request_timeout})
     if err ~= nil then
-        return nil, wrap_vshard_err(err, func_name, replicaset.id)
+        local replicaset_id, route_err
+        replicaset, route_err = vshard_router:route(bucket_id)
```
Serpentian: this still may produce different replicaset id from the one, we errored on. Every call to `route` can give different replicasets due to internal `discovery` process in vshard. It's also possible, that this `route` is very costly and we may execute it, when timeout has already passed.

Can't we wrap the error inside the `call_single`. But I don't insist, I'm ok with how it's now

crud/common/call.lua
```diff
@@ -73,47 +73,143 @@ local function wrap_vshard_err(err, func_name, replicaset_id)
...
+
+    call_opts.timeout = timeout
+    if call_opts.request_timeout ~= nil and call_opts.request_timeout > timeout then
+        call_opts.request_timeout = timeout
+    end
+
+    return call_on_replicaset(replicaset, method, func_name, func_args, call_opts)
+end
+
+local function call_map_with_recovery(replicaset, method, func_name, func_args,
```
Serpentian, line 187: The commit msg of the third one says: `Map calls do not go through the recovery at all:`. This is not true anymore

crud/common/call.lua
Satbek: fixed

crud/common/call.lua
Satbek: fixed

test/helper.lua
Satbek: moved to 3rd

crud/common/rebalance.lua
Satbek: fixed

crud/common/call.lua
Serpentian: Nah, it's not, now we take the replicaset id before request, but inside the `call_signle_with_recovery` it may change (`recover_from_err` can return another replicaset).

We either should call for `wrap_vshard_err` inside `call_single_with_recovery` or return the replicaset id from the `call_single_with_recovery` itself, I see no other way

Satbek: ok, I see.  Now call_single_with_recovery returns replicaset_id

Serpentian (approved):
Thank you for the fixes, the current version seems correct. Speaking of existing code, it requires substantial refactoring, it's very hard to make changes to it now: there're too many rough edges and small bugs in it. But I really appreciate, that you've fixed the things we've found.

I really hope, that someday we'll drop all that code from crud and allow vshard to do its work in order to avoid such nasty bugs. But for that to happen we must find a way to bump the perf of all requests

PR #512, opened 2026-08-20
feat: add single and batch storage calls (TNTP-9261)

vakhov:
Adds crud.storage_call() and crud.storage_call_many() to invoke named stored functions on storage masters through regular net.box.call(), without exposing vshard to clients.

Routing supports an explicit bucket_id or a space primary key, including custom sharding keys and functions. Batch calls are dispatched concurrently to affected replicasets, and results retain input order.

Target functions must have a persistent body and setuid = false. Execution and result serialization preserve the original caller’s privileges.

Stored functions are responsible for closing their local transactions on both success and error paths. CRUD does not check for open transactions or roll them back.

Results are serialized when the storage sends its response, without preliminary validation or copying of individual return values. A serialization or infrastructure error returns a top-level error without partial results; committed changes remain committed.

Individual function errors do not stop other batch items. CRUD does not automatically retry target functions. Ambiguous completion is marked with may_have_side_effects = true.

TNTP-9261.


a1div0 (approved):

README.md
```diff
@@ -1513,6 +1514,113 @@ crud.storage_info()
 ...
 ```
 
+### Storage call
```
p0rtale, line 1518: nit: make this section consistent with the other methods

p0rtale (approved):
LGTM, thanks for the PR!

Left one non-blocking nit on the README

README.md
vakhov: Done.

